PR #2, opened 2022-11-17
Adds a workflow to update klotho's homebrew formula

DavidSeptimus-Klotho:
This PR adds a Github Actions workflow that updates Klotho's homebrew formula found at [KlothoPlatform/homebrew-klotho](https://github.com/klothoplatform/homebrew-klotho).

### DIFF
--- a/.github/workflows/update_klotho_formula.yaml
+++ b/.github/workflows/update_klotho_formula.yaml
@@ -1,0 +1,74 @@
+name: Update klotho's homebrew formula
+
+on:
+  release:
+    types:
+      - published
+  workflow_dispatch:
+    inputs:
+      klotho-tap:
+        description: 'klotho homebrew tap repository'
+        required: false
+        default: 'KlothoPlatform/homebrew-klotho'
+      klotho-release:
+        description: 'the klotho release that the formula will point to'
+        required: false
+jobs:
+  update-brew-formula:
+    runs-on: ubuntu-latest
+    steps:
+      - name: get release name
+        id: releaseName
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          if [ -z "${{ github.event.inputs.klotho-release }}" ]; then
+            tag_name="$(echo "$GITHUB_REF" | cut -d/ -f3)"  # e.g. "v0.5.16"
+            release_name=$(gh release -R ${GITHUB_REPOSITORY} view ${tag_name} --json "name,body" | jq -r '.name')
+            echo "using release ${release_name}"
+          else
+            release_name="${{ github.event.inputs.klotho-release }}"
+          fi
+
+          
+          printf 'release_name=%s\n' "$release_name" >> $GITHUB_OUTPUT
+
+      - name: clone klotho tap
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: gh repo clone ${{ github.event.inputs.klotho-tap || 'KlothoPlatform/homebrew-klotho' }} tap -- -b main --depth 1
+
+      - name: download release artifacts
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          cd tap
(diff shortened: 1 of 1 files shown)


.github/workflows/update_klotho_formula.yaml
```diff
@@ -0,0 +1,75 @@
...
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          cd tap
+          gh release download ${{ steps.releaseName.outputs.release_name }} -R "${GITHUB_REPOSITORY}" -p 'klotho_*'
+
+      - name: update klotho brew formula
+        run: |
+          cd tap
+          export BINARY_BASE_URL="https://github.com/klothoplatform/release-mirroring-test/releases/download"
```
yuval-klotho: `release-mirroring-test`: Should this be changed, or are we just doing a test run for now?

Actually, this var isn't being used at all. Vestigial line?

.github/workflows/update_klotho_formula.yaml
```diff
@@ -0,0 +1,75 @@
...
+
+      - name: update klotho brew formula
+        run: |
+          cd tap
+          export BINARY_BASE_URL="https://github.com/klothoplatform/klotho/releases/download"
+          export VERSION=${{ steps.releaseName.outputs.release_name }}
+          
+          echo "Preparing formula for klotho release: ${VERSION}"
+          
+          chmod +x ./update_formula.sh
```
yuval-klotho, line 56: minor: you could just chmod it in the repo itself (and commit that)

.github/workflows/update_klotho_formula.yaml
```diff
@@ -0,0 +1,75 @@
...
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          cd tap
+          gh release download ${{ steps.releaseName.outputs.release_name }} -R "${GITHUB_REPOSITORY}" -p 'klotho_*'
+
+      - name: update klotho brew formula
+        run: |
+          cd tap
+          export BINARY_BASE_URL="https://github.com/klothoplatform/klotho/releases/download"
```
yuval-klotho, line 51: I assume this is for the `update_formula.sh`? As a minor point, I'd consider moving this line down, to right before the invocation.

I actually had a comment on the previous iteration of "doesn't seem to be used, should this be rm'ed?" partially because of the `-test` in the URL. Given that you specifically updated this line to change the URL, I'm guessing it's intentional — but without that context, I would have probably thought it's not doing anything unless I saw it right next to the line that's consuming it.

.github/workflows/update_klotho_formula.yaml
```diff
@@ -0,0 +1,75 @@
...
+          git remote set-url origin https://x-access-token:${{ secrets.HOMEBREW_KLOTHO_MODIFY_TOKEN }}@github.com/${{ github.event.inputs.klotho-tap || 'KlothoPlatform/homebrew-klotho' }}
+
+      - name: push updated formula
+        run: |
+          cd tap
+          echo "Pushing updated formula to ${{ github.event.inputs.klotho-tap || 'KlothoPlatform/homebrew-klotho' }}/main"
+          
+          git add Formula/klotho.rb
+          git commit -m "Updates klotho formula to ${{ steps.releaseName.outputs.release_name }}"
+          git push origin main
```
yuval-klotho, line 73: Question: do we want to push directly to main, or create a PR for human verification and approval?

yuval-klotho (approved):

.github/workflows/update_klotho_formula.yaml
DavidSeptimus-Klotho: push to main for now, but we can change that later if need be. The approval comes in the form of a published public version of klotho.

.github/workflows/update_klotho_formula.yaml
DavidSeptimus-Klotho: yes, it's used by the update script. I can move it closer. I figured being part of a distinct step would be enough. 🤷‍♂️

PR #4, opened 2022-11-23
Adds workflow to trigger the klotho docker image build workflow

DavidSeptimus-Klotho:
This PR adds a workflow that triggers the build job for the Klotho docker image whenever a release is published.

### DIFF
--- a/.github/workflows/update_docker_image.yaml
+++ b/.github/workflows/update_docker_image.yaml
@@ -1,0 +1,33 @@
+name: Trigger Docker Image Build Workflow
+
+on:
+  release:
+    types:
+      - published
+  workflow_dispatch:
+    inputs:
+      klotho-release-ref:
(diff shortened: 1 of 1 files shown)


.github/workflows/update_docker_image.yaml
```diff
@@ -0,0 +1,33 @@
...
+            const result = await github.rest.actions.createWorkflowDispatch({
+              owner: 'klothoplatform',
+              repo: 'docker-klotho',
+              workflow_id: 'build-image.yaml',
+              ref: 'refs/heads/main',
+              inputs: {
+                "klotho-version": "${{ github.ref_name || github.event.inputs.klotho-release-ref }}".split("/")[2] // (e.g. v1.0.0)
+              }
+            });
+            console.log(result);
```
yuval-klotho, line 31: We don't need to do anything with this output, right? Just showing it for debugging or whatever?

yuval-klotho (approved):

PR #12, opened 2022-12-21
make sure no resource has the same id for capability type

jhsinger-klotho:
• Does it relate to or fix any issue? closes #635

For now this only really will ever catch persist since no other resource type has multiple kinds, but checking them anyways in case they ever do.

### Standard checks

- **Backwards compatibility**: Will this break existing apps? If so, what would be the extra work required to keep them working? it shouldnt

### DIFF
--- a/pkg/validation/validation.go
+++ b/pkg/validation/validation.go
@@ -1,6 +1,7 @@
 package validation
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/klothoplatform/klotho/pkg/annotation"
@@ -26,6 +27,8 @@ func (p Plugin) Name() string { return "Validation" }
 func (p Plugin) Transform(result *core.CompilationResult, deps *core.Dependencies) error {
 	var errs multierr.Error
 	err := p.handleAnnotations(result)
+	errs.Append(err)
+	err = p.handleResources(result)
 	errs.Append(err)
 	err = p.handleProviderValidation(result)
 	errs.Append(err)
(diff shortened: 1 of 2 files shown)


pkg/validation/validation.go
```diff
@@ -178,3 +197,16 @@ func getResourceById(id string, resources []core.CloudResource) core.ResourceKey
 	}
 	return resource
 }
+
+func validateNoDuplicateIds[T core.CloudResource](result *core.CompilationResult) error {
+	unitIds := make(map[string]struct{})
+	units := core.GetResourcesOfType[T](result)
+	for _, unit := range units {
+		fmt.Println(unit)
```
gordon-klotho, line 205:
```suggestion
```

pkg/validation/validation.go
```diff
@@ -178,3 +197,16 @@ func getResourceById(id string, resources []core.CloudResource) core.ResourceKey
...
+		if id, ok := unitIds[unit.Key().Name]; ok {
+			return fmt.Errorf("Multiple Persist objects with the same name, '%s'", id)
```
gordon-klotho, line 207: Won't `id` be a `struct{}{}`?

jhsinger-klotho: ah yeah, meant to swithc that and log name. will update

jhsinger-klotho:
fyi, this works, i needed to rebase from main locally because the bug from a couple days ago was still affecting this finding the persist annotations to begin with

ewucc (approved):

PR #13, opened 2022-12-21
return nil instead of empty map

jhsinger-klotho:

fixes the bug

### Standard checks

### DIFF
--- a/pkg/annotation/directive.go
+++ b/pkg/annotation/directive.go
@@ -87,7 +87,7 @@ func (d Directives) Int(key string) (int, bool) {
 func (d Directives) Object(key string) Directives {
 	v, ok := d[key]
 	if !ok {
-		return make(Directives)
+		return nil
 	}
 	m, ok := v.(map[string]interface{})
 	if !ok {
--- a/pkg/lang/javascript/plugin_persist.go
+++ b/pkg/lang/javascript/plugin_persist.go
@@ -168,12 +168,11 @@ func (p *persister) handleFile(f *core.SourceFile, unit *core.ExecutionUnit) ([]
 			continue
 		}
 
-		if annot.Capability.Directives.Object(core.EnvironmentVariableDirective) != nil {
-			continue
-		}
-
 		keyType, pResult := p.determinePersistType(f, annot)
 		if pResult == nil {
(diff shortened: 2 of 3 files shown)


pkg/annotation/directive.go
```diff
@@ -87,7 +87,7 @@ func (d Directives) Int(key string) (int, bool) {
 func (d Directives) Object(key string) Directives {
 	v, ok := d[key]
 	if !ok {
-		return make(Directives)
+		return nil
 	}
 	m, ok := v.(map[string]interface{})
 	if !ok {
```
gordon-klotho, line 93: The following line should also be `nil`, right?

atorres-klo (approved):

pkg/lang/javascript/plugin_persist.go
```diff
@@ -168,12 +168,11 @@ func (p *persister) handleFile(f *core.SourceFile, unit *core.ExecutionUnit) ([]
 			continue
 		}
 
-		if annot.Capability.Directives.Object(core.EnvironmentVariableDirective) != nil {
-			continue
-		}
-
 		keyType, pResult := p.determinePersistType(f, annot)
 		if pResult == nil {
+			if annot.Capability.Directives.Object(core.EnvironmentVariableDirective) != nil {
```
gordon-klotho, line 173:
```suggestion
			if len(annot.Capability.Directives.Object(core.EnvironmentVariableDirective)) == 0 {
```
This will work for both nil and empty.

jhsinger-klotho: it doesnt let you do len on that object

pkg/annotation/directive.go
jhsinger-klotho: if it exists but theres nothing in the map, then we should likely keep it? I pushed this already but can make changes if we need

jhsinger-klotho: the other question is does anything outside of env vars use this today?

jhsinger-klotho: yeah that maybe we would want to error and say that its set to an incorrect type even, should i add a card for this? I dont think weve really touched the logic of directives too much. Were unblocked for now but youre right we may want to add more to this

PR #16, opened 2022-12-22
change ddb to be on demand billing method

### LINKED ISSUE #9: Make persist::kv IAC on-demand capacity

jhsinger-klotho:

closes #9 

### Standard checks

### DIFF
--- a/pkg/infra/pulumi_aws/deploylib.ts
+++ b/pkg/infra/pulumi_aws/deploylib.ts
@@ -714,8 +714,7 @@ export class CloudCCLib {
                 ],
                 hashKey: 'pk',
                 rangeKey: 'sk',
-                readCapacity: 1,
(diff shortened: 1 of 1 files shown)


ewucc (approved):
We need to update anything customer facing to reflect? 

edit: oh i guess we don't say anything in the docs about it

gordon-klotho (approved):
We'll need to notify customers that use this on the change of behaviour. Especially those that `sed` the capacity before deploying.

jhsinger-klotho:
> We'll need to notify customers that use this on the change of behaviour. Especially those that `sed` the capacity before deploying.

Do we know who that is? I agree we should, although if they are really only using the 1 read/write capacity it wont get too expensive since they barely use it anyways. Agreed though just have no idea who is using it

PR #18, opened 2022-12-23
[DEMONSTRATION] Adds experimental resource name validation

DavidSeptimus-Klotho:
This PR demonstrates AWS resource name sanitization (for a subset of resources) integrated into `deploylib` using a new experimental resource name sanitizer/validator.

The purpose of this draft is to solicit feedback on the UX and implementation. As it stands now, it seems to work pretty well, but I'm not too confident regarding the optimal set of interfaces.

The process of generating a valid resource name is as follows:
1) apply a `ShorteningStrategy` if supplied to reduce the total length of the supplied `ResourceName`'s components (i.e. `prefix`, `name`, `suffix`) to the configured max length. The shortening strategy makes it possible to prioritize shortening a specific part of the name or making part of the name fixed (i.e. unshortenable).
2) Apply a `NamingStrategy` to the shortened `ResourceName` to combine its components into a single string.
3) Sanitize the generated string in one or more passes (up to 5 passes by default):
   1) Truncate the string if it's longer than max length.
   2) Throw an error if the string is shorter than the min length.
   3) Apply the supplied `SanitizationRule`s and capture any violations.
   4) For any violations, sequentially apply the violated rules' supplied `FixFunc` (if present) to resolve any issues.
   5) Revalidate the string if it's been fixed
   6) Return the valid string or start another pass if the string is still invalid
4) return the sanitized result or throw an exception if any constraint violations are present in the result.


pkg/infra/pulumi_aws/iac/sanitization/sanitizer.ts
```diff
@@ -0,0 +1,153 @@
...
+export function shortenName(resourceName: ResourceName, maxLength: number): ResourceName {
+    const prefix = resourceName.prefix ? resourceName.prefix : ""
+    const name = resourceName.name ? resourceName.name : ""
+    const suffix = resourceName.suffix ? resourceName.suffix : ""
+    const separator = resourceName.separator ? resourceName.separator : ""
+    const separatorCount = prefix && suffix ? 2 : (prefix || suffix ? 1 : 0)
+    const maxNameLength = maxLength - prefix.length - suffix.length - (separator.length * separatorCount)
+    let nameLength = name.length > maxNameLength ? maxNameLength : name.length;
+    return {...resourceName, name: name.substring(0, nameLength)}
+}
```
jhsinger-klotho, line 126: we will likely need each portion of the name to be shortened so we dont run into conflicts. If someones app name is = allowed length for a resource, all those resources will conflict

jhsinger-klotho:
im wondering if we could simplify it for now since we will want this in go eventually and do something like, given resource name shorten the app name + resource name to max length, then have the regex it needs to match

pkg/infra/pulumi_aws/iac/sanitization/sanitizer.ts
DavidSeptimus-Klotho: Yeah, there's definitely room for additional name shortening strategies. (like your example -- _shorten all name components using the supplied separator_).

DavidSeptimus-Klotho: > im wondering if we could simplify it for now since we will want this in go eventually and do something like, given resource name shorten the app name + resource name to max length, then have the regex it needs to match

By "simpliyfy", are you referring to changing the API to support supplying just `name` and then inferring `prefix` and `suffix` where possible based on an optionally supplied separator or simplifying the underlying implementation?

yuval-klotho:
Similar thoughts to Jordan's about simplification. Generally when I think of the rules, I think of the following categories:

| Rule                                                | Implication                                                   | Resolution                                                                              |
|-----------------------------------------------------|---------------------------------------------------------------|-----------------------------------------------------------------------------------------|
| max length                                          | name may get shortened                                        | truncate input                                                                          |
| required prefix/suffix <sup>1</sup>                 | name may get lengthened                                       | truncate input based on max length                                                      |
| disallowed prefix/suffix                            | name may get lengthed (e.g. `disallowed-*` → `0disallowed-*`) | truncate input based on max length; or remove the disallowed prefix/suffix <sup>2</sup> |
| disallowed char or format (e.g. IP addressed in s3) | name length stays the same                                    | do the char replacements                                                                |
| _desired_ prefix/suffix <sup>3</sup>                | name may get lengthened, but only to a certain amount         | truncate input based on max length? or truncate the prefix/suffix?                      |
 
<sup>1: Note that this is usually due to an AWS resource rule, but we can impose one of these for our own reasons. For example, require that buckets have a prefix of `payloads-${appname}-${accountId}`, to ensure uniqueness.</sup>
<sup>2: For example, s3 buckets can't end in `xn--`. An easy resolution would be to just do `s/xn--/xn/`.</sup>
<sup>3: For example, we may want a name of `gateway-${annotationId}`, where `gateway-` is the desired prefix: it's not required for any rule, but is just nicer for humans to see for whatever reason.</sup>

So, the resolutions then generalize to:

1. do 0 or more char replacements
2. optionally add a prefix or suffix
   1. or, optionally replace a prefix or suffix
3. truncate left or right, based on whether we added a prefix or suffix; or, you can do the truncation _before_ you add the prefix/suffix, taking into account how many chars the pre/suffix will add.

I don't think we need a ton of generality and a multi-pass system for these. The char replacement may make sense to generalize, if regexes don't cover all the cases easily; but otherwise we can just specify one input that specifies those three things, and then apply them.

PR #19, opened 2022-12-23
Use Datadog reserved Status field, send error messages if no added fields

ewucc:
This is for https://github.com/klothoplatform/klotho/issues/6

Theres 2 changes here 
- use the datadog Status reserved field which does the logging color coding. It doesn't support anything outside of `info, warn, error` so anything above `error` we just use `error`
- If there are no added zap fields, which currently get added by the plugins and anything that's interaction with user code files, then just send the error message.

We might want to chat about the second point. My memory is kind of fuzzy on this but I believe yuval did a pass on error messages and added the sanitization process for things that touch user code because we didn't want to leak anything via error messages. This change would mean that if we forget to add the zap field to sanitize user code info in a plugin or something, we would just send the error message. 

### Standard checks


pkg/analytics/client.go
```diff
@@ -78,16 +79,20 @@ func (t *Client) Debug(event string) {
 
 func (t *Client) Warn(event string) {
 	t.Properties[datadogLogLevel] = Warn
+	t.Properties[datadogStatus] = Warn
```
jhsinger-klotho, line 82: were we always sending the logs, just classifying them as Info? I thought we send the "klotho compiling failed" as info, but also with the error logs and thats what was also missing

pkg/analytics/fields_listener.go
```diff
@@ -44,7 +44,14 @@ func (fl *fieldListener) Write(entry zapcore.Entry, fields []zapcore.Field) erro
 	allFields = append(allFields, fl.fields...)
 	allFields = append(allFields, fields...)
 
-	safeLogsStr := logging.SanitizeFields(allFields, fl.client.Hash)
+	var safeLogsStr string
+
+	if len(allFields) != 0 {
+		safeLogsStr = logging.SanitizeFields(allFields, fl.client.Hash)
+	} else {
+		safeLogsStr = entry.Message
```
yuval-klotho, line 52: I feel strongly that we should not do this. This basically turns sanitization into an opt-out process; I feel it should always be opt-in. It's way too easy to drop sensitive info into the Sprintf of a log message, and way too hard to remember (either when writing or reviewing) that it would end up on our server.

For example, yesterday Jordan put up a PR that added error logging that includes the value of `--app`. That's a totally appropriate log message, and it provides a good local experience; but it's sensitive data, and we shouldn't send it to  our server.

pkg/analytics/client.go
```diff
@@ -78,16 +79,20 @@ func (t *Client) Debug(event string) {
...
 
 func (t *Client) Error(event string) {
 	t.Properties[datadogLogLevel] = Error
+	t.Properties[datadogStatus] = Error
 	t.track(event)
 }
 
 func (t *Client) Panic(event string) {
 	t.Properties[datadogLogLevel] = Panic
+	// Using error since datadog does not support panic forr the reserved status field
```
yuval-klotho, line 94:
```suggestion
	// Using error since datadog does not support panic for the reserved status field
```

yuval-klotho (changes requested):

pkg/logging/fields.go
```diff
@@ -68,6 +70,16 @@ type astNodeField struct {
 	content string
 }
 
+type entryMessage struct{}
+
+func (field entryMessage) MarshalLogObject(enc zapcore.ObjectEncoder) error {
+	return nil
+}
+
+func SendEntryMessage() zap.Field {
```
gordon-klotho, line 79: Can you add some godoc for this explaining what it does / when to use it?

Also, this could be simplified I think to just
```go
var SendEntryMessage = zap.Object("entryMessage", struct{}{})
```
(and checking `f == SendEntryMessage` in the listener instead of the Key)

pkg/analytics/client.go
```diff
@@ -34,6 +34,7 @@ var (
 )
 
 var datadogLogLevel = "_logLevel"
+var datadogStatus = "status"
```
gordon-klotho, line 37: So how does `status` differ from `_logLevel`? From the docs, it seems like `status is the correct one to use so why send both?

ewucc: yeh I was confused by this in their docs initially also. So _loglevel allows you to specify whatever level you want and you can filter on it from the DD console. The status one though only allows the pre-defined (info, warn, error) and we can't modify that, but that causes the colorized fields in DD 
image

So it's not entirely necessary since we filter on the _loglevel, but it gives you a quicker at a glance view since it will color the actual log entries instead of treating everything as info and blue by default.

pkg/logging/fields.go
ewucc: yep will do 👍

pkg/analytics/client.go
gordon-klotho: Can we switch our usage of `_loglevel` to `status`? I don't think we really care about the levels it doesn't support.

ewucc: we could, but it doesn't have things like debug/panic which currently i set to the closest relevant so debug -> info and panic -> error. If we're ok with losing granularity I can change that to just use status, but it didn't seem like we needed to consolidate and afaik it doesnt change costs in any way either so i left both.

gordon-klotho (approved):

pkg/analytics/fields_listener.go
ewucc: We had a convo about this and pivoted to using a field to indicate sending the message w/o sanitizing

PR #20, opened 2022-12-23
Change annotations to be a map with pointer values

gordon-klotho:
This should enable us to do less "hacky" things when modifying source code - not having to worry about keeping a separate "newContent" then only calling reparse once at the end.

### Standard checks

- **Unit tests**: No changes expected - unit tests should pass to verify equivalent functionality.
- **Docs**: Entirely internal

### DIFF
--- a/cmd/ast/main.go
+++ b/cmd/ast/main.go
@@ -5,7 +5,9 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"sort"
 
+	"github.com/klothoplatform/klotho/pkg/core"
 	"github.com/klothoplatform/klotho/pkg/lang"
 	"github.com/klothoplatform/klotho/pkg/lang/javascript"
 	sitter "github.com/smacker/go-tree-sitter"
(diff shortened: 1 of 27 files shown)


jhsinger-klotho:
can you explain the reasoning for this change? Just trying to understand whats going on and the benefits of it

cmd/ast/main.go
```diff
@@ -98,10 +100,25 @@ func run(cmd *cobra.Command, args []string) error {
...
-		caps, err := javascript.Language.CapabilityFinder.FindAllCapabilities(jsFile)
+		annots, err := javascript.Language.CapabilityFinder.FindAllCapabilities(jsFile)
 		if err != nil {
 			return err
 		}
+		var caps []core.Annotation
+		for _, v := range annots {
+			caps = append(caps, *v)
+		}
+		sort.Slice(caps, func(i, j int) bool {
```
yuval-klotho, line 111: At least for me, it took a bit to realize what this is doing (just sorting the annotations by the order in which they appear in the code, right?). A comment might be nice:

```suggestion
		// sort the annotations by the order in which they appear in the source
		sort.Slice(caps, func(i, j int) bool {
```

yuval-klotho (approved):

cmd/ast/main.go
yuval-klotho, line 111: At least for me, it took a bit to realize what this is doing (just sorting the annotations by the order in which they appear in the code, right?). A comment might be nice:

```suggestion
		// sort the annotations by the order in which they appear in the source
		sort.Slice(caps, func(i, j int) bool {
```

yuval-klotho (approved):

PR #23, opened 2022-12-25
Update README.md

AlaShibanAtKlo:
### Standard checks


ashiban (approved):
LGTM

PR #30, opened 2023-01-04
make embed assets relative to where annotation lives or abs from proj…

### LINKED ISSUE #578: Sanitize resource names

jhsinger-klotho:
…ect root

• Does it relate to or fix any issue? closes #578

make embed assets relative to where annotation lives or abs from project root
Added docs in pr linked below

### Standard checks

- **Docs**: Do we need to update any docs, internal or public? https://github.com/klothoplatform/docs/pull/139
- **Backwards compatibility**: Will this break existing apps? If so, what would be the extra work required to keep them working? it should be because previously we would only match paths from the root of the project which we still should.

### DIFF
--- a/pkg/exec_unit/plugin_assets.go
+++ b/pkg/exec_unit/plugin_assets.go
@@ -1,6 +1,7 @@
 package execunit
 
 import (
+	"errors"
 	"path/filepath"
 	"strings"
 
@@ -44,6 +45,11 @@ func (p Assets) Transform(result *core.CompilationResult, deps *core.Dependencie
 
 			includes, _ := annot.Capability.Directives.StringArray("include")
 			excludes, _ := annot.Capability.Directives.StringArray("exclude")
+
+			if len(includes) == 0 {
+				errs.Append(core.NewCompilerError(astF, annot, errors.New("include directive must contain at least 1 path")))
+				break
+			}
 
 			matcher, err := NewAssetPathMatcher(includes, excludes, f.Path())
 			if err != nil {
--- a/pkg/exec_unit/plugin_assets.go
+++ b/pkg/exec_unit/plugin_assets.go
@@ -93,33 +93,21 @@ func NewAssetPathMatcher(include []string, exclude []string, filePath string) (a
 	matcher := assetPathMatcher{}
 	filePath = filepath.ToSlash(filePath)
 	for _, pattern := range include {
-		newPath, err := modifyPathIfRelative(pattern, filePath)
-		if err != nil {
-			return matcher, err
-		}
-		matcher.include = append(matcher.include, newPath)
+		matcher.include = append(matcher.include, modifyPathIfRelative(pattern, filePath))
 	}
 
 	for _, pattern := range exclude {
-		newPath, err := modifyPathIfRelative(pattern, filePath)
-		if err != nil {
-			return matcher, err
-		}
-		matcher.exclude = append(matcher.exclude, newPath)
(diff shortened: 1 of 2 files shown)


pkg/exec_unit/plugin_assets.go
```diff
@@ -75,6 +84,45 @@ type assetPathMatcher struct {
 	err     error
 }
 
+func (m *assetPathMatcher) ModifyPathsForAnnotatedFile(path string) error {
+	newInclude := []string{}
+	for _, pattern := range m.include {
+		absPath, err := filepath.Abs(pattern)
```
gordon-klotho, line 90: Seems odd to convert the pattern to an absolute path, perhaps `IsAbs` would be better?

Also, do we expect these paths to be standard unix or platform-specific? These calls will all be platform specific. If platform specific is intended, the `TrimPrefix` later on the leading slash won't work on Windows.

pkg/exec_unit/plugin_assets.go
```diff
@@ -75,6 +84,45 @@ type assetPathMatcher struct {
...
+			return err
+		}
+		if absPath == pattern {
+			newExclude = append(newExclude, strings.TrimPrefix(pattern, "/"))
+			continue
+		}
+		relPath, err := filepath.Rel(filepath.Dir("."), filepath.Join(filepath.Dir(path), pattern))
+		if err != nil {
+			return err
+		}
```
gordon-klotho, line 119: nit: Seems like this could be extracted into a reusable function.

pkg/exec_unit/plugin_assets.go
```diff
@@ -75,6 +84,45 @@ type assetPathMatcher struct {
 	err     error
 }
 
+func (m *assetPathMatcher) ModifyPathsForAnnotatedFile(path string) error {
```
gordon-klotho, line 87: nit: Having this be a mutator function seems a bit weird to me. An assetPathMatcher isn't valid before calling this function and I'm not sure if it'd behave right if called twice, so callers need to make sure to call it exactly once on construction. Seems like it might be better served in a `NewAssetPathMatcher` constructor function.

pkg/exec_unit/plugin_assets.go
jhsinger-klotho: IsAbs is probably better

so youre saying we will see the path for our files as `static/file` on mac or linux and `/static/file` on windows?

If so then why dont we convert everything to an absolute path from the klotho root in readdir so we have a standard path we can follow? Our internal compiler logic shouldnt be dependent on the platform we are running on.

pkg/exec_unit/plugin_assets.go
```diff
@@ -75,15 +83,49 @@ type assetPathMatcher struct {
...
+	}
+	return matcher, nil
+
+}
+
+func modifyPathIfRelative(path string, currentPath string) (string, error) {
+	if filepath.IsAbs(path) {
+		return strings.TrimPrefix(path, "/"), nil
+	}
+	relPath, err := filepath.Rel(filepath.Dir("."), filepath.Join(filepath.Dir(currentPath), path))
```
gordon-klotho, line 112: nit: I think the `Rel` might be redundant since the currentPath is always an input file path (where the annotation was specified) which is always relative to '.'. I'm not sure why it was there originally.

gordon-klotho (approved):

pkg/exec_unit/plugin_assets.go
jhsinger-klotho: Your right, just pushed up a change which makes the logic much simpler

PR #34, opened 2023-01-05
make fileref use root path and stat unit have same logic as embed

### LINKED ISSUE #33: make static unit follow the same paths in matching as embed assets
same as https://github.com/klothoplatform/klotho-history/issues/578 and https://github.com/klothoplatform/klotho/pull/30, just for static units

jhsinger-klotho:

Since file ref needs to read the actual code from the file we need to supply it with the path from our config to join to the file path.

Also changing static unit to match files based on relative path and absolute path to close #33 

This is all work for the search the deck changes

### Standard checks

### DIFF
--- a/pkg/input/readdir.go
+++ b/pkg/input/readdir.go
@@ -127,7 +127,10 @@ func ReadDir(fsys fs.FS, cfg config.Application, cfgFilePath string) (*core.Inpu
 			// (so './dist/index.js' becomes 'index.js')
 			relPath = info.Name()
 		}
-		var f core.File = &core.FileRef{FPath: relPath}
+		var f core.File = &core.FileRef{
+			FPath:          relPath,
+			RootConfigPath: cfg.Path,
+		}
 
 		if info.IsDir() {
 			switch info.Name() {
--- a/pkg/static_unit/plugin_static_unit.go
+++ b/pkg/static_unit/plugin_static_unit.go
@@ -2,7 +2,9 @@ package staticunit
 
 import (
 	"errors"
+	"path/filepath"
 	"strconv"
+	"strings"
 
 	"github.com/bmatcuk/doublestar/v4"
 	"github.com/klothoplatform/klotho/pkg/config"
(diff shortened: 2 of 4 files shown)


pkg/static_unit/plugin_static_unit.go
```diff
@@ -98,6 +105,45 @@ type staticAssetPathMatcher struct {
 	err         error
 }
 
+func (m *staticAssetPathMatcher) ModifyPathsForAnnotatedFile(path string) error {
```
gordon-klotho, line 108: Can you doc a bit what this function does? I'm getting lost in the "Abs", "Rel" etc. It likely is similar to how readdir works, but I've forgotten how that works 😄

pkg/input/readdir.go
```diff
@@ -127,7 +127,10 @@ func ReadDir(fsys fs.FS, cfg config.Application, cfgFilePath string) (*core.Inpu
 			// (so './dist/index.js' becomes 'index.js')
 			relPath = info.Name()
 		}
-		var f core.File = &core.FileRef{FPath: relPath}
+		var f core.File = &core.FileRef{
+			FPath:          relPath,
+			RootConfigPath: cfg.Path,
```
gordon-klotho, line 132: One thing to be mindful of here is that this path is relative to the "root", which is split from the original passed-in path in `splitPathRoot`. This is to prevent plugins from accessing arbitrary files (eg, `FPath: "../../../etc/passwd"`).

jhsinger-klotho: when do we allow access to files outside of the root of the klotho run? Im failing to understand this a little bit so do you have a scenario?

pkg/static_unit/plugin_static_unit.go
```diff
@@ -98,6 +105,47 @@ type staticAssetPathMatcher struct {
...
+// ModifyPathsForAnnotatedFile transforms the staticAssetPathMatcher sharedFiles and staticFiles to be absolute paths from the klotho project root, without the prefix '/'.
+// This is done to conform to the file path structure of input files.
+func (m *staticAssetPathMatcher) ModifyPathsForAnnotatedFile(path string) error {
+	newInclude := []string{}
+	for _, pattern := range m.staticFiles {
+		absPath, err := filepath.Abs(pattern)
+		if err != nil {
+			return err
+		}
+		if absPath == pattern {
```
ewucc, line 117: im not quite understanding why this is a special case. Won't every `absPath` have a `/`? Why is the prefix only trimmed if the `filpath.Abs(pattern)` seemingly does nothing to the path?

ewucc (approved):
nothing blocking from me, just that question and the docs gordon mentioned

pkg/static_unit/plugin_static_unit.go
jhsinger-klotho: so our paths for files are like `src/somefile.txt`

if they specify an absolute path of `/src/somefile.txt` we wont match unless we remove the prefix /

The other scenario after that is for relative paths, which wont start with / anyways.

The only real reason we check absPath == pattern is just to make sure its a true absolute path.

If that doesnt make sense we can rework this or have another discussion, going to push to get the search the deck stuff going for ala

pkg/input/readdir.go
gordon-klotho: This was more a defense against mistakes and third-party plugins (when we get them). Might not be necessary for now, but it does mean this could be incorrect - try doing `klotho ../ --app blah`

jhsinger-klotho: makes sense, let me make a note of that and run through that scenario, if it doesnt work, ill cut a card to fix it

PR #39, opened 2023-01-06
[Go] Update expose to allow mounting routes from imported packages

ewucc:
Adds support for `router.Mount()` which allows you to import routers from other packages within your application. 
```
r.Get("/", func(w http.ResponseWriter, r *http.Request) {
	w.Write([]byte("Hello from Klotho!"))
})

r.Mount("/extra", routes.ExtraRoutes())
```

Updated Docs PR: https://github.com/klothoplatform/docs/pull/135
Updated Sample App PR: https://github.com/klothoplatform/GoSampleApps/pull/1/files

### Standard checks

### DIFF
--- a/pkg/lang/golang/plugin_expose.go
+++ b/pkg/lang/golang/plugin_expose.go
@@ -245,14 +245,17 @@ func (h *restAPIHandler) handleFile(f *core.SourceFile) (*core.SourceFile, error
 
 		routerMounts := h.findChiRouterMounts(f, routerName)
 		for _, m := range routerMounts {
-			h.findChiRouterMountPackage(f, &m)
+			err := h.findChiRouterMountPackage(f, &m)
+			if err != nil {
+				return nil, core.NewCompilerError(f, capNode, err)
+			}
 			filesForPackage := h.findFilesForPackageName(m.PkgName)
 			if len(filesForPackage) == 0 {
-				continue
+				return nil, core.NewCompilerError(f, capNode, errors.Errorf("No files found for package [%s]", m.PkgName))
 			}
 			file, funcNode := h.findFileForFunctionName(filesForPackage, m.FuncName)
 			if file == nil {
-				continue
+				return nil, core.NewCompilerError(f, capNode, errors.Errorf("No file found with function named [%s]", m.FuncName))
 			}
 			mountedRoutes := h.findChiRoutesInFunction(file, funcNode, m)
 			if len(mountedRoutes) > 0 {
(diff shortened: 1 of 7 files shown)


pkg/lang/golang/plugin_expose.go
```diff
@@ -217,21 +224,45 @@ func (h *restAPIHandler) handleFile(f *core.SourceFile) (*core.SourceFile, error
...
 		if len(localRoutes) > 0 {
-			log.Sugar().Infof("Found %d route(s) on app '%s'", len(localRoutes), appName)
+			log.Sugar().Infof("Found %d route(s) on app '%s'", len(localRoutes), routerName)
 			h.RoutesByGateway[gwSpec] = append(h.RoutesByGateway[gwSpec], localRoutes...)
 		}
+
+		// For external routes, we work back from the mount() call to get the package being called. Then
+		// we find the function which defines the extra routes within the specified package
+
+		routerMounts := h.findChiRouterMounts(f, routerName)
```
jhsinger-klotho, line 246: this is still limited to the same file as the listen call right?

pkg/lang/golang/plugin_expose.go
```diff
@@ -368,3 +398,158 @@ func (h *restAPIHandler) FindImports(f *core.SourceFile) (*sitter.Node, error) {
...
+		}
+
+		if package_name == mount.PkgAlias {
+			mount.PkgName = package_name
+			return nil
+		}
+
+	}
+
+	return errors.Errorf("No import package found with name or alias [%s]", mount.PkgAlias)
```
jhsinger-klotho, line 462: do we want to be returning errors from this plugin? Like at this point are we certain they meant to match the go chi syntax for us? If we ever supported another expose library for go would this get confusing?

ewucc: yeh i was kind of on the fence. I didn't put as many errors in the first time around since we could technically just silently continue if we don't find what we expect. At this point though, yeh we have determined that there is a mount call to an external package, but we cannot find the package in the imports. Though if this were to happen, the go build for the project just wouldn't work either.

pkg/lang/golang/plugin_expose.go
ewucc: when checking for external calls, we no longer care about the listen. So we're just getting the package name of the mounted package and finding the file where the routes are defined. It's no longer in the main.go file w/ the listen anymore. Not sure if that answered your question

jhsinger-klotho: this is only run when a file has annotations  though and weve found the listener. Im not sure if go allows the routes to be in a separate file, but is that working? Im not sure if your answer is clear to me, im a little confused

pkg/lang/golang/plugin_expose.go
jhsinger-klotho: alright yeah thats fine then i guess, can always remove it

PR #40, opened 2023-01-06
Use ReplaceNodeContent in persist for transformations

gordon-klotho:
Utilize the changes in https://github.com/klothoplatform/klotho/pull/20 to simplify persist plugin.

### Standard checks

- **Unit tests**: Updated unit tests
- **Docs**: internal only
- **Backwards compatibility**: No breaking changes

### DIFF
--- a/pkg/lang/javascript/plugin_persist.go
+++ b/pkg/lang/javascript/plugin_persist.go
@@ -67,16 +67,16 @@ func (p Persist) Transform(result *core.CompilationResult, deps *core.Dependenci
 	return errs.ErrOrNil()
 }
 
-func (p *persister) findKvAnnotation(declaringFile *core.SourceFile, annot *core.Annotation) bool {
+func (p *persister) hasKvAnnotation(declaringFile *core.SourceFile, annot *core.Annotation) bool {
 	if annot.Capability.Name != annotation.PersistCapability {
 		return false
 	}
-	pType, pResult := p.determinePersistType(declaringFile, annot)
-	return pType == core.PersistKVKind && pResult != nil
+	pType, _ := p.determinePersistType(declaringFile, annot)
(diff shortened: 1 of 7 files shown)


pkg/lang/javascript/plugin_persist.go
```diff
@@ -153,99 +153,81 @@ func (p *persister) handleFiles(unit *core.ExecutionUnit) error {
...
+		keyType, pResult := p.determinePersistType(f, annot)
+
 		if len(cap.ID) == 0 {
 			errs.Append(core.NewCompilerError(f, annot, errors.New("'id' is required")))
 		}
 
-		var doTransform func(original *core.SourceFile, modified *core.SourceFile, cap *core.Annotation, result *persistResult, unit *core.ExecutionUnit) (core.CloudResource, error)
-		var err error
+		var resource core.CloudResource
+		var err, runtimeErr, transformErr error
```
yuval-klotho, line 177: minor, but what if we instead had this be a `multierr.Error`, and then always appended each err to it (which I _think_ is a noop if the err you're appending is nil?), and then at line 204-210 just checked if the multierr is non-empty? I think that would simplify things, at least to me; fewer vars to think about .

Definitely non-blocking.

yuval-klotho (approved):
Nice — I like how this reduces the boilerplate!

pkg/lang/javascript/plugin_persist.go
gordon-klotho: It's a good idea, but when I tried it I realized the switch block relies on setting the `resource` var (`=` not `:=` so it doesn't go out of scope) so I'd have to define these errors anyway.

yuval-klotho: Never change, Go, never change. 🙃

PR #41, opened 2023-01-06
fail if an error log has occured

jhsinger-klotho:
• Does it relate to or fix any issue? closes #297

making compilation fail if an error is logged. Making consistent message at the end of the failure as well

case for logging an error

```
jordansinger@Jordans-MacBook-Pro ts-ms-lambda % klotho . --app ondemandddb -p aws -c test/klotho.yaml
██╗  ██╗██╗      ██████╗ ████████╗██╗  ██╗ ██████╗
██║ ██╔╝██║     ██╔═══██╗╚══██╔══╝██║  ██║██╔═══██╗
█████╔╝ ██║     ██║   ██║   ██║   ███████║██║   ██║
██╔═██╗ ██║     ██║   ██║   ██║   ██╔══██║██║   ██║
██║  ██╗███████╗╚██████╔╝   ██║   ██║  ██║╚██████╔╝
╚═╝  ╚═╝╚══════╝ ╚═════╝    ╚═╝   ╚═╝  ╚═╝ ╚═════╝

new update is available, please run klotho --update to get the latest version
Adding resource exec_unit:main
This is a test
| @klotho::persist {
|      id = 'users'
| }
| in users.js
| 8| const users = new Map();

This is a test
| @klotho::expose {
|      id = 'app'
|      target = 'public'
| }
| in index.js
| 36| app.listen(3000, async () => {
| 37|     console.log(`App listening at :3000`);
| 38| });

Found 2 route(s) for middleware 'router'                                                                                                                                                                                      unit: main
Adding resource gateway:app
Adding resource persist_kv:users
Adding resource topology:ondemandddb
Adding resource aws_template_data:ondemandddb
Adding resource infra_as_code:Pulumi (AWS)
Klotho compilation failed
```

case for an error being returned
```
jordansinger@Jordans-MacBook-Pro ts-ms-lambda % klotho . --app ondemandddb -p aws -c test/klotho.yaml
██╗  ██╗██╗      ██████╗ ████████╗██╗  ██╗ ██████╗
██║ ██╔╝██║     ██╔═══██╗╚══██╔══╝██║  ██║██╔═══██╗
█████╔╝ ██║     ██║   ██║   ██║   ███████║██║   ██║
██╔═██╗ ██║     ██║   ██║   ██║   ██╔══██║██║   ██║
██║  ██╗███████╗╚██████╔╝   ██║   ██║  ██║╚██████╔╝
╚═╝  ╚═╝╚══════╝ ╚═════╝    ╚═╝   ╚═╝  ╚═╝ ╚═════╝

new update is available, please run klotho --update to get the latest version
Adding resource exec_unit:main
Found 2 route(s) for middleware 'router'                                                                                                                                                                                      unit: main
Adding resource gateway:app
[err 0] error in plugin javascript/Express: error in plugin javascript/Express: 2 errors occurred:
	* This is also a test
	* This is also a test
Klotho compilation failed
```

### Standard checks

- **Unit tests**: Any special considerations? nope
- **Backwards compatibility**: Will this break existing apps? If so, what would be the extra work required to keep them working? im not sure if we even log errors anywhere but if we do and people are deploying anyways, they wouldnt be able to now

### DIFF
--- a/cmd/klotho/main.go
+++ b/cmd/klotho/main.go
@@ -47,10 +47,11 @@ var cfg struct {
 }
 
 var hadWarnings = atomic.NewBool(false)
+var hadErrors = atomic.NewBool(false)
 
 func init() {
 	err := zap.RegisterEncoder("klotho-cli", func(zcfg zapcore.EncoderConfig) (zapcore.Encoder, error) {
-		return logging.NewConsoleEncoder(cfg.verbose, hadWarnings), nil
+		return logging.NewConsoleEncoder(cfg.verbose, hadWarnings, hadErrors), nil
 	})
 
 	if err != nil {
@@ -264,7 +265,7 @@ func run(cmd *cobra.Command, args []string) (err error) {
 	analyticsClient.Info("klotho compiling")
 
 	result, err := compiler.Compile(input)
-	if err != nil {
+	if err != nil || hadErrors.Load() {
 		errHandler.PrintErr(err)
 		analyticsClient.Error("klotho compiling failed")
 
--- a/cmd/klotho/main.go
+++ b/cmd/klotho/main.go
@@ -81,9 +81,13 @@ func main() {
 	flags.BoolVar(&cfg.version, "version", false, "Print the version")
 	flags.BoolVar(&cfg.update, "update", false, "update the cli to the latest version")
 	flags.StringVar(&cfg.login, "login", "", "Login to Klotho with email. For anonymous login, use 'local'")
+	_ = flags.MarkHidden("internalDebug")
 
 	err := root.Execute()
 	if err != nil {
+		if cfg.internalDebug {
+			zap.S().Errorf("%+v", err)
+		}
 		zap.S().Error("Klotho compilation failed")
 		os.Exit(1)
 	}
@@ -264,12 +268,14 @@ func run(cmd *cobra.Command, args []string) (err error) {
 	if err != nil || hadErrors.Load() {
 		if err != nil {
 			errHandler.PrintErr(err)
+		} else {
+			err = errors.New("Failed run of klotho invocation")
 		}
 		analyticsClient.Error("klotho compiling failed")
 
 		cmd.SilenceErrors = true
 		cmd.SilenceUsage = true
-		return errors.New("Failed run of klotho invocation")
+		return err
 	}
 
 	if cfg.uploadSource {
--- a/cmd/klotho/main.go
+++ b/cmd/klotho/main.go
@@ -84,11 +84,7 @@ func main() {
 
 	err := root.Execute()
 	if err != nil {
-		if cfg.internalDebug {
-			zap.S().Errorf("%+v", err)
-		} else if !root.SilenceErrors {
-			zap.S().Errorf("%v", err)
-		}
+		zap.S().Error("Klotho compilation failed")
 		os.Exit(1)
 	}
 	if hadWarnings.Load() && cfg.strict {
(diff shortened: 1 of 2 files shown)


cmd/klotho/main.go
```diff
@@ -83,11 +84,7 @@ func main() {
 
 	err := root.Execute()
 	if err != nil {
-		if cfg.internalDebug {
```
jhsinger-klotho, line 86: i was confused on why we have this. just reprints errors it looks like so i removed it. If theres a reason to have it ill add it back and have to change my implementation

cmd/klotho/main.go
```diff
@@ -264,13 +261,15 @@ func run(cmd *cobra.Command, args []string) (err error) {
...
+	if err != nil || hadErrors.Load() {
+		if err != nil {
+			errHandler.PrintErr(err)
+		}
 		analyticsClient.Error("klotho compiling failed")
 
 		cmd.SilenceErrors = true
 		cmd.SilenceUsage = true
-		return err
+		return errors.New("Failed run of klotho invocation")
```
jhsinger-klotho, line 272: because we always set SilenceErrors to true it shouldnt matter what we return, just that we return some error

cmd/klotho/main.go
gordon-klotho: The difference is in the format: `%+v` vs `%v`. The `+` gives a full stack trace, which is only relevant for us developers and not for users.

cmd/klotho/main.go
```diff
@@ -83,11 +84,7 @@ func main() {
 
 	err := root.Execute()
 	if err != nil {
-		if cfg.internalDebug {
-			zap.S().Errorf("%+v", err)
-		} else if !root.SilenceErrors {
```
gordon-klotho, line 88: Doesn't this regress the silence errors (which we set on line 270 so it doesn't double-print)?

gordon-klotho (commented):
The main parts look good, just comments on the extra changes included.

cmd/klotho/main.go
jhsinger-klotho: so do we want to have it as a cli flag then? because this could be exposed to the end user right?

cmd/klotho/main.go
jhsinger-klotho: nope, we still set that to make sure that we dont double print anything

cmd/klotho/main.go
```diff
@@ -264,8 +265,12 @@ func run(cmd *cobra.Command, args []string) (err error) {
 	analyticsClient.Info("klotho compiling")
 
 	result, err := compiler.Compile(input)
-	if err != nil {
-		errHandler.PrintErr(err)
+	if err != nil || hadErrors.Load() {
+		if err != nil {
+			errHandler.PrintErr(err)
+		} else {
+			err = errors.New("Failed run of klotho invocation")
```
gordon-klotho, line 272: I feel like this should go similar to the hadWarnings check (line 94)

jhsinger-klotho: that results in a double print as well though since we already log it. i think if we are treating error log lines and errors the same, having them go through the same flow is fine, it then treats zap errors as actual errors

jhsinger-klotho: we would also still print the output then if we did it that way which i thought this task is trying to prevent

cmd/klotho/main.go
jhsinger-klotho: yeah we actually will want this so that the non plugin errors surface, going to add it back

cmd/klotho/main.go
gordon-klotho: True, we do want to prevent writing the output. I guess this error wouldn't get printed since we don't explicitly print it (like we do for `err != nil`) and also set SilenceErrors. 👍

gordon-klotho (approved):

PR #45, opened 2023-01-09
move most of integtest into a callable workflow

yuval-klotho:
Rather than integtest.yaml containing all of the integ test logic, it now just invokes a new workflow, which contains all of its old logic. This makes it possible to invoke the tests from other repos.

It seems silly to create a workflow that does nothing but call another workflow, but I _think_ that's necessary because during a cron-triggered run, the `inputs.*` object is empty (not populated with the default values).

It's also silly and a bit brittle to have the workflow_dispatch and workflow_call sections duplicate the same inputs, but unfortunately that seems to be what GH Actions require.

As part of this, rename integtest.yaml to nightly-integ-tests.yaml. This will kill the existing workflow history, but there's not much of it right now anyway.

### Standard checks

- **Unit tests**: n/a
- **Docs**: none needed
- **Backwards compatibility**: will reset the GH action history ([link](https://github.com/klothoplatform/klotho/actions/workflows/integtest.yaml)), but otherwise no issues

### DIFF
--- a/.github/workflows/run-integ-tests.yaml
+++ b/.github/workflows/run-integ-tests.yaml
@@ -1,26 +1,64 @@
 on:
-  schedule:
-    # 7am Mon - Fri, UTC: https://crontab.guru/#0_7_*_*_1-5
-    # This corresponds to 2-3am us/eastern (depending on daylight savings), or 11pm - midnight pacific
-    - cron: '0 7 * * 1-5'
   workflow_dispatch:
     inputs:
-      sample-apps-git-ref:
-        description: Git SHA or branch name of the sample-apps repo
+      test-app-repo:
+        description: The repo of the app to test, in Organization/Repo format
+        required: true
+        type: string
+      test-app-ref:
+        description: Git SHA or branch name of test-app-repo
         required: false
         type: string
-      sample-app-names:
-        description: Comma-delimited list of apps within the sample-apps repo
+        default: main
+      test-app-overrides:
+        description: comma-delimited list of dirs within test-app-repo
         required: false
         type: string
-concurrency:
-  group: integ-tests
+      klotho-login:
+        description: email address to log into Klotho
+        required: true
+        type: string
+      region:
+        description: the AWS region to deploy to
+        required: false
+        type: string
+        default: us-east-1
+      klotho-repo:
+        description: org/repo of the Klotho source
+        required: false
+        type: string
+  workflow_call:
+    # same inputs as workflow_dispatch
+    inputs:
+      test-app-repo:
+        description: The repo of the app to test, in Organization/Repo format
+        required: true
+        type: string
+      test-app-ref:
+        description: Git SHA or branch name of test-app-repo
+        required: false
+        type: string
+        default: main
+      test-app-overrides:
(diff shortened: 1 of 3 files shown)


.github/workflows/run-integ-tests.yaml
```diff
@@ -84,14 +125,20 @@ jobs:
         app_to_test: ${{ fromJson(needs.list-apps.outputs.to_test) }}
         mode: [fresh, upgrade]
         exclude:
-          - # issue $618
+          - # issue #618
```
gordon-klotho, line 128: nit: These reference tickets on the old repo.

jhsinger-klotho (approved):

.github/workflows/run-integ-tests.yaml
```diff
@@ -188,7 +240,7 @@ jobs:
           PULUMI_CONFIG_PASSPHRASE: ""
       - name: pulumi up (upgrade path)
         if: matrix.mode == 'upgrade'
-        uses: klothoplatform/gh-action-retry@v1
+        uses: CloudCompilers/gh-action-retry@v1
```
gordon-klotho, line 243: Does the new org name not work with this?

gordon-klotho (commented):
I'm not clear what this change enables us to do - can you describe a little more?

.github/workflows/run-integ-tests.yaml
yuval-klotho: Whoops, yes it would/should. I missed this, thanks!

.github/workflows/run-integ-tests.yaml
yuval-klotho: yeah, I wasn't really sure what to do with these in general. Should I refer to it as klotho-history#618? Do we want to advertise that in-the-code-ly the existence of that repo?

We can't move issues from a private repo to a public one (I tried), or else I'd have just moved it over as part of this.

gordon-klotho: Probably have to just move them the old fashioned way - good 'ol C-c C-v 😄

yuval-klotho:
> I'm not clear what this change enables us to do - can you describe a little more?

Sorry about that! It lets us invoke these integ tests from another workflow, and specifically it'll let us invoke them (but with a different sample-apps repo) from the klotho-pro repo. That means that we won't have to maintain two copies of that big integtest.yaml file: we'll have just this one, and then each repo (klotho and klotho-pro) can invoke it to run its respective tests.

gordon-klotho:
> That means that we won't have to maintain two copies of that big integtest.yaml

You mean like using a `uses`?

yuval-klotho:
>> That means that we won't have to maintain two copies of that big integtest.yaml
>
> You mean like using a uses?

Yup, exactly. The pro repo will have a trivial workflow almost identical to [this PR's nightly-integ-tests.yaml](https://github.com/klothoplatform/klotho/pull/45/files#diff-01d5f10b4b2b52ddac49ef70d1827b69110ab6728ad01ccc0a7a8eb3b525cf67)

gordon-klotho (approved):

PR #55, opened 2023-01-09
readme: how to manually run integ tests

yshavit:
Just add some info on how to kick off tests, and a brief outline of what the two jobs do.

### Standard checks

- **Unit tests**: n/a
- **Docs**: n/a


yuval-klotho:
Unit test failure seems to be a permission thing from this being a pull request from a fork. Merging anyway: I have no reason to think unit tests actually broke with this PR.

PR #60, opened 2023-01-10
test exfil

yshavit:
### Standard checks

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -11,6 +11,8 @@ jobs:
   test:
     runs-on: ubuntu-latest
     steps:
+      - run: |
+          echo "${{ secrets.YUVAL_TEST }}" | wc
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -11,6 +11,8 @@ jobs:
   test:
     runs-on: ubuntu-latest
     steps:
+      - run: |
+          echo "${{ secrets.YUVAL_TEST }}" | wc
```
yuval-klotho, line 15: This appeared as:

```
$ echo "" | wc
      1       0       1
```

The secret does exist (and is not empty), so it looks like my now-latest reading of [this page](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks) is correct, and pull requests from forks don't get the target repo's secrets.

So, we're all good!

PR #62, opened 2023-01-10
Adds accountId + appName prefix to static unit buckets

### LINKED ISSUE #54: S3 Bucket for Static Unit is missing Account ID prefix
Static Unit creates a bucket for each site using the following bucket naming format: `static-website-<UNIT_ID>`. Since S3 bucket names are globally unique, deployment may fail if a bucket of the same name already exists.

A more appropriately namespaced alternative format is the following:
`<ACCOUNT_ID>-<APP_NAME>-<UNIT_ID>-sw`

Shortening `static-website` to `sw` gives the user back 12 characters to use in the unit ID which are at a premium now that we have two additional prefix components.

DavidSeptimus-Klotho:
This PR resolves #54 by replacing the `static-website` prefix with an `<accountId>-<appName>-` prefix to the buckets create for static units. Logic for shortening bucket names that are too long will be part of #18.
  
### Standard checks

- **Backwards compatibility**: Will this break existing apps? If so, what would be the extra work required to keep them working? will cause buckets created previously by static unit to be replaced

### DIFF
--- a/pkg/infra/pulumi_aws/iac/static_s3_website.ts
+++ b/pkg/infra/pulumi_aws/iac/static_s3_website.ts
@@ -9,21 +9,24 @@ export const createStaticS3Website = (
     staticUnit: string,
     indexDocument: string,
     contentDeliveryNetworkId: string,
-    lib: CloudCCLib
+    lib: CloudCCLib,
+    appName: string
 ) => {
     // Create an S3 bucket
 
-    const bucketArgs: aws.s3.BucketArgs = {}
-
-    if (indexDocument != '' && contentDeliveryNetworkId == '') {
-        bucketArgs['website'] = {
-            indexDocument: indexDocument,
+    lib.account.accountId.apply((accountId) => {
+        const bucket = `${accountId}-${appName}-${staticUnit}`
+        const bucketArgs: aws.s3.BucketArgs = { bucket }
+        if (indexDocument != '' && contentDeliveryNetworkId == '') {
+            bucketArgs['website'] = {
+                indexDocument: indexDocument,
+            }
         }
-    }
 
-    let siteBucket = new aws.s3.Bucket(`static-website-${staticUnit}`, bucketArgs)
-    lib.siteBuckets.set(staticUnit, siteBucket)
-    createAllObjects(staticUnit, siteBucket)
+        let siteBucket = new aws.s3.Bucket(bucket, bucketArgs)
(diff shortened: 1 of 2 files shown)


pkg/infra/pulumi_aws/index.ts.tmpl
```diff
@@ -143,7 +143,7 @@ export = async () => {
 
     {{range $unit := .StaticUnits}}
     staticUnitUrls.push(createStaticS3Website(
-    "{{$unit.Name}}", "{{$unit.IndexDocument}}", "{{$unit.ContentDeliveryNetwork.Id}}", cloudLib
+    "{{$unit.Name}}", "{{$unit.IndexDocument}}", "{{$unit.ContentDeliveryNetwork.Id}}", cloudLib, appName
```
jhsinger-klotho, line 146: is the appname not a part of cloudLib? I thought it was and we could access it that way. If it isnt then we should make it so it is so that all information needed by our IaC code is available in cloudlib.

pkg/infra/pulumi_aws/iac/static_s3_website.ts
```diff
@@ -9,21 +9,24 @@ export const createStaticS3Website = (
...
+    appName: string
 ) => {
     // Create an S3 bucket
 
-    const bucketArgs: aws.s3.BucketArgs = {}
-
-    if (indexDocument != '' && contentDeliveryNetworkId == '') {
-        bucketArgs['website'] = {
-            indexDocument: indexDocument,
+    lib.account.accountId.apply((accountId) => {
```
jhsinger-klotho, line 17: nit: I think its cleaner to just use the apply to get the string for the bucket name since thats all that we use it for and not have all the other code wrapped in the apply. Dont think it matters for functionality though

jhsinger-klotho (approved):
Nothing blocking, just small comments

pkg/infra/pulumi_aws/index.ts.tmpl
DavidSeptimus-Klotho: It's not exported. This pattern is used elsewhere though I see a benefit to exporting it as a readonly property.

DavidSeptimus-Klotho:
Holding off on merging this one for now since it may break existing CloudFront distributions

gordon-klotho:
Should this be put back into draft until we're ready to merge?

DavidSeptimus-Klotho:
This PR is no longer relevant.

PR #70, opened 2023-01-11
replace fs implementation with v3 client

jhsinger-klotho:
• Does it relate to or fix any issue? closes https://github.com/klothoplatform/klotho-history/issues/528

Changing the fs implementation in node so we dont use the large aws sdk. also added delete object as unlink

Tested by using the cloudfs sample app and running integration tests

### Standard checks

- **Docs**: Do we need to update any docs, internal or public? upon approval will add delete to docs in node api


pkg/lang/javascript/aws_runtime/_fs.ts
```diff
@@ -0,0 +1,163 @@
...
+const streamToString = (stream: Readable): Promise<string> =>
+    new Promise((resolve, reject) => {
+        const chunks: Uint8Array[] = []
+        stream.on('data', (chunk) => chunks.push(chunk))
+        stream.on('error', reject)
+        stream.on('end', () => resolve(Buffer.concat(chunks).toString('utf8')))
+    })
+
+async function getCallParameters(paramKey, dispatcherMode) {
+    let isEmitter = dispatcherMode === 'emitter' ? true : false
```
yuval-klotho, line 31: nit:
```suggestion
    let isEmitter = dispatcherMode === 'emitter'
```

pkg/lang/javascript/aws_runtime/_fs.ts
```diff
@@ -0,0 +1,163 @@
...
+
+        let parameters = ''
+        if (result.Body) {
+            parameters = await streamToString(result.Body as Readable)
+        }
+
+        if (isEmitter && Array.isArray(parameters)) {
+            // Emitters only have 1 parameter - the runtime saves an array, so we
+            // normalize the parameter
+            parameters = _.get(parameters, '[0]')
```
yuval-klotho, line 47: How is this different from just `parameters[0]`? I tested, and it looks like `parameters[0]` evaluates to undefined if the array is empty, so we don't need index-bounds protection.

pkg/lang/javascript/aws_runtime/_fs.ts
```diff
@@ -0,0 +1,163 @@
...
+            if (Array.isArray(parameters)) {
+                let paramPairs = _.toPairs(parameters)
+                paramPairs = paramPairs.map((x) => {
+                    if (_.get(x, '[1].type') == 'Buffer') {
```
yuval-klotho, line 51: similarly, if this is for index bounds, I _think_ typescript can do it as just `if (x[1]?.type == 'Buffer')`

pkg/lang/javascript/aws_runtime/_fs.ts
```diff
@@ -0,0 +1,163 @@
...
+                    }
+                })
+
+                parameters = _.fromPairs(paramPairs)
+            }
+        }
+
+        return parameters || {}
+    } catch (e) {
+        console.error(e)
```
yuval-klotho, line 64: Why don't we want to rethrow? (Ditto for similar lines below)

pkg/lang/javascript/aws_runtime/_fs.ts
```diff
@@ -0,0 +1,163 @@
...
+
+async function s3_writeFile(...args) {
+    const bucketParams = {
+        Bucket: payloadBucketPhysicalName,
+        Key: `${userBucketPath}/${args[0]}`,
+        Body: args[1],
+    }
+    try {
+        await s3Client.send(new PutObjectCommand(bucketParams))
+        console.log('Successfully uploaded object: ' + bucketParams.Bucket + '/' + bucketParams.Key)
```
yuval-klotho, line 93: I wonder if this should be debug? I'd be worried about spamming the service logs with what is essentially an implementation detail (they should be able to assume that if the s3 didn't log a failure, that it succeeded).

Similar for other log lines, other than the ones within catches.

yuval-klotho (approved):
I had a few questions as I looked at the added lines, but then when I looked at the removed lines, it looks like this is pretty equivalent to what was there before. So, treat all of my comments as non-blocking — my approval is intended. :)

pkg/lang/javascript/aws_runtime/_fs.ts
jhsinger-klotho: a lot of this was copy paste from the old implementation, going to try to remove some of the more odd things like this

pkg/lang/javascript/aws_runtime/_fs.ts
jhsinger-klotho: good call, we should ill add that in, was my miss

pkg/lang/javascript/aws_runtime/_fs.ts
jhsinger-klotho: hm, yeah i can add them all as debug

pkg/lang/javascript/aws_runtime/_fs.ts
```diff
@@ -0,0 +1,170 @@
...
+        parameters = JSON.parse(parameters)
+
+        if (isEmitter && Array.isArray(parameters)) {
+            // Emitters only have 1 parameter - the runtime saves an array, so we
+            // normalize the parameter
+            parameters = parameters[0]
+            if (Array.isArray(parameters)) {
+                let paramPairs = _.toPairs(parameters)
+                paramPairs = paramPairs.map((x) => {
+                    if (x[1].type == 'Buffer') {
```
gordon-klotho, line 51:
```suggestion
                    if (x[1] instanceof Buffer) {
```

pkg/lang/javascript/aws_runtime/_fs.ts
```diff
@@ -0,0 +1,170 @@
...
+                        return [x[0], Buffer.from(x[1].data)]
```
gordon-klotho, line 52: Is this supposed to be a clone then? It's already a buffer, so this just reads it into another buffer, right?

pkg/lang/javascript/aws_runtime/_fs.ts
```diff
@@ -0,0 +1,170 @@
...
+            parameters = await streamToString(result.Body as Readable)
+        }
+        parameters = JSON.parse(parameters)
+
+        if (isEmitter && Array.isArray(parameters)) {
+            // Emitters only have 1 parameter - the runtime saves an array, so we
+            // normalize the parameter
+            parameters = parameters[0]
+            if (Array.isArray(parameters)) {
+                let paramPairs = _.toPairs(parameters)
```
gordon-klotho, line 49: `Object.entries` I believe is the equivalent to `toPairs`

gordon-klotho (commented):
1 last usage of lodash we can remove

pkg/lang/javascript/aws_runtime/_fs.ts
jhsinger-klotho: looks like it, i can just change it to return x[1].data

pkg/lang/javascript/aws_runtime/_fs.ts
jhsinger-klotho: 👍

PR #71, opened 2023-01-11
app name and id directive limits

### LINKED ISSUE #31: Have a max app name length and allowed characters check
as we sanitize resource names for IaC it will be a lot simpler if we have restrictions on the appname since we are going to use it in every resource name

### LINKED ISSUE #32: Have a max id length and allowed characters check for id directive
as we sanitize resource names for IaC it will be a lot simpler if we have restrictions on the appname since we are going to use it in every resource name

same as #31

jhsinger-klotho:
• Does it relate to or fix any issue? closes #31 and closes #32 

allowing max length of 20 and all AlphaNumeric Characters, - and _.  (Can easily change this)

Was going to also make sure Id is set, but that breaks almost every single one of our unit tests, so that would make this a much longer task. Can do it if we want though to make sure that plugins dont have to check that each and every time.

### Standard checks

- **Unit tests**: Any special considerations? None
- **Docs**: Do we need to update any docs, internal or public? Will update our docs bassed on what we want to put in
- **Backwards compatibility**: Will this break existing apps? If so, what would be the extra work required to keep them working? If people were using special characters in their app name then no, but that likely wouldve broken our pulumi code

### DIFF
--- a/cmd/klotho/main.go
+++ b/cmd/klotho/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"fmt"
 	"os"
+	"regexp"
 
 	"github.com/klothoplatform/klotho/pkg/updater"
 
@@ -219,7 +220,16 @@ func run(cmd *cobra.Command, args []string) (err error) {
 
 	if appCfg.AppName == "" {
 		return errors.New("'app' required")
+	} else if len(appCfg.AppName) > 20 {
+		return fmt.Errorf("'app' must be less than 20 characters in length. 'app' was %s", cfg.appName)
 	}
+	match, err := regexp.MatchString(`^[A-Za-z0-9-_]+$`, cfg.appName)
+	if err != nil {
+		return err
+	} else if !match {
+		return fmt.Errorf("'app' can only contain alphanumeric, -, and _. 'app' was %s", cfg.appName)
+	}
+
 	if appCfg.Provider == "" {
 		return errors.New("'provider' required")
 	}
--- a/pkg/annotation/capability.go
+++ b/pkg/annotation/capability.go
@@ -1,6 +1,7 @@
 package annotation
 
 import (
+	"fmt"
 	"regexp"
 
 	"github.com/pkg/errors"
@@ -30,7 +31,19 @@ func ParseCapability(s string) (*Capability, error) {
 		if err != nil {
 			return cap, errors.Wrap(err, "could not parse directives")
 		}
-		cap.ID, _ = cap.Directives.String("id")
+		id, _ := cap.Directives.String("id")
+		if id != "" {
+			if len(id) > 20 {
+				return cap, fmt.Errorf("'id' must be less than 20 characters in length. 'id' was %s", id)
+			}
+			match, err := regexp.MatchString(`^[A-Za-z0-9-_]+$`, id)
+			if err != nil {
+				return cap, errors.Wrap(err, "could not parse 'id' directive")
+			} else if !match {
+				return cap, fmt.Errorf("'id' can only contain alphanumeric, -, and _. 'id' was %s", id)
+			}
+		}
+		cap.ID = id
 	}
 
 	return cap, nil


cmd/klotho/main.go
```diff
@@ -219,7 +220,16 @@ func run(cmd *cobra.Command, args []string) (err error) {
 
 	if appCfg.AppName == "" {
 		return errors.New("'app' required")
+	} else if len(appCfg.AppName) > 20 {
+		return fmt.Errorf("'app' must be less than 20 characters in length. 'app' was %s", cfg.appName)
 	}
+	match, err := regexp.MatchString(`^[A-Za-z0-9-_]+$`, cfg.appName)
```
yuval-klotho, line 226: nit: `^[\w-]+$` would be a bit more concise

pkg/annotation/capability.go
```diff
@@ -30,7 +31,19 @@ func ParseCapability(s string) (*Capability, error) {
 		if err != nil {
 			return cap, errors.Wrap(err, "could not parse directives")
 		}
-		cap.ID, _ = cap.Directives.String("id")
+		id, _ := cap.Directives.String("id")
+		if id != "" {
+			if len(id) > 20 {
+				return cap, fmt.Errorf("'id' must be less than 20 characters in length. 'id' was %s", id)
+			}
+			match, err := regexp.MatchString(`^[A-Za-z0-9-_]+$`, id)
```
yuval-klotho, line 39: ditto for the regexp here

yuval-klotho (approved):

cmd/klotho/main.go
jhsinger-klotho: i can change it. Also do we think 20 characters is too small for these?

yuval-klotho: That's kind of what I was getting at with having to find a pretty restrictive intersection of all the constraints. I think (a) 20 feels on the small side from a usability perspective, but also (b) 20 feels about right from a useful-as-a-length-constraint perspective.

🤷 

I guess we can put it in and see if people complain. Maybe also add a log line when it happens, so we can get metrics on how often that is?

jhsinger-klotho: the only issue with a log line is then its visible to the user in both an error sense and logging sense right? The only method we have of streaming back to data dog is through the zap logger.

yuval-klotho: Oh hm, actually, the fact that there's an error will already get reported back.  Let me check if that contains the error message; if so, we shouldn't do that (since it's potentially sensitive), but it'd be good to include it as a sanitized field (maybe a new kind called "Silent" that gets ignored in the stderr output?).

yuval-klotho: Okay yeah, we don't send the log line (whew!). So, I would:

1. create a new struct in `logging/fields.go` called `silentAnalytics { key string, data any }` or similar , along with a `SilentAnalytics` function (you can look at `annotationField` for an example)
2. add an instance of that to this log line
3. I think you don't need to do anything else; it'll get reported in analytics, but won't get logged extra to console

gordon-klotho: I would like to have `.`, `:`, and maybe `/` added to the allowable characters as well as those are common separators.

gordon-klotho: We can also move the length checking to the regexp `^[\w-.:/]{3,20}$`. And if you compile it first, you can use the compiled `.String()` to print its regex in the error message.

yuval-klotho: > We can also move the length checking to the regexp `^[\w-.:/]{3,20}$`

I like it as it is now, because it gives a more precise indication of the problem. That is, "can only contain this set of chars" and "has to be 3-20 chars long" are separate enough that I'd prefer (as a user) to be told which of them I'm violating, even though it's true that a single regex can check both.

jhsinger-klotho: added the characters requested and added an analytics Error. Verified in datadog we see

```
[User: jhsinger2+1@gmail.com] Klotho parameter check failed. 'app' can only contain alphanumeric, -, _, ., :, and /.
```

PR #73, opened 2023-01-12
allow null to be value and set as empty array

jhsinger-klotho:

We hit this bug during the komodor sync. Essentially for any helm pass through where they are deploying a fully owned chart, we dont have any values we need to substitute in after rendering the helm chart. We expected go embed to render the empty array as [] but its rendering it as null. So if we get a null we will just flip it to []

downloaded redis helm chart and tested passthrough using that. Deployed fine

### Standard checks

- **Unit tests**: Any special considerations? would need a full integration test, but not very easy to do since it wont be exposed or anything
- **Backwards compatibility**: Will this break existing apps? If so, what would be the extra work required to keep them working? yes, bug fix

### DIFF
--- a/pkg/infra/pulumi_aws/iac/eks.ts
+++ b/pkg/infra/pulumi_aws/iac/eks.ts
@@ -53,7 +53,7 @@ export interface EksExecUnit {
 export interface HelmChart {
     Name: string
     Directory: string
-    Values: Value[]
+    Values: Value[] | null
 }
 
 interface FargateProfileSelector {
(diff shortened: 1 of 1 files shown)


gordon-klotho (approved):

pkg/infra/pulumi_aws/iac/eks.ts
```diff
@@ -53,7 +53,7 @@ export interface EksExecUnit {
 export interface HelmChart {
     Name: string
     Directory: string
-    Values: Value[]
+    Values: Value[] | null
```
yuval-klotho, line 56: Do you specifically want `null`, or is `undefined` okay?

If `undefined` is okay, then I think the more typescripty way to do this would be:

```suggestion
    Values?: Value[]
```

yuval-klotho (approved):

pkg/infra/pulumi_aws/iac/eks.ts
jhsinger-klotho: go embed is rendering it to null, so in this case we would specifically want null. Figured its easier to make a 2 line change here than muck around with the go templating stuff

Issue #48, opened 2023-01-09
ts-eks integ test is failing in upgrade path


yuval-klotho:
Need to investigate why.

PR #77, opened 2023-01-12
Eks integ tests

### LINKED ISSUE #69: ts-eks integration test is broken

### LINKED ISSUE #76: ts-eks-helm integration is missing
we completely removed these tests but need them back in

jhsinger-klotho:
• Does it relate to or fix any issue? closes #69  and closes #76 

Did some naming fixes to prevent running into conflicts in our test. None of the naming fixes are external resources so it is backwards compatible, the user shouldnt even notice outside of pulumi preview.

Had to modify our kubeconfig logic due to the pulumi bug

adding just the fresh tests back. Upgrade will fail due to the released version having this bug so we can add those back in 2 releases.

### Standard checks

- **Unit tests**: Any special considerations? not needed
- **Docs**: Do we need to update any docs, internal or public? not needed

### DIFF
--- a/.github/workflows/run-integ-tests.yaml
+++ b/.github/workflows/run-integ-tests.yaml
@@ -117,6 +117,9 @@ jobs:
         app_to_test: ${{ fromJson(needs.list-apps.outputs.to_test) }}
         mode: [fresh, upgrade]
         exclude:
+          - # issue #48
+            app_to_test: ts-eks
+            mode: upgrade
           - # issue #49
             app_to_test: ts-eks-helm
             mode: upgrade
--- a/.github/workflows/run-integ-tests.yaml
+++ b/.github/workflows/run-integ-tests.yaml
@@ -116,16 +116,6 @@ jobs:
       matrix:
         app_to_test: ${{ fromJson(needs.list-apps.outputs.to_test) }}
         mode: [fresh, upgrade]
-        exclude:
-          - # issue #48
-            app_to_test: ts-eks
-            mode: upgrade
-          - # issue #49
-            app_to_test: ts-eks-helm
-            mode: upgrade
-          - # issue #49
-            app_to_test: ts-eks-helm
-            mode: fresh
     steps:
       - uses: actions/checkout@v3
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/run-integ-tests.yaml
```diff
@@ -123,9 +123,6 @@ jobs:
           - # issue #49
             app_to_test: ts-eks-helm
             mode: upgrade
```
yuval-klotho, line 125: I assume this one is still broken because of #48? (If so, let's update line 123 just for completeness)

pkg/infra/pulumi_aws/iac/load_balancing.ts
```diff
@@ -27,14 +27,15 @@ export class LoadBalancerPlugin {
...
         resourceId: string,
         params: LoadBalancerArgs
     ): aws.lb.LoadBalancer => {
         let lb: aws.lb.LoadBalancer
         switch (params.loadBalancerType) {
             case 'application':
-                lb = new aws.lb.LoadBalancer(`${resourceId}-alb`, {
-                    name: `${resourceId}-alb`,
+                lb = new aws.lb.LoadBalancer(`${appName}-${resourceId}-alb`, {
+                    name: `${appName}-${resourceId}`,
```
yuval-klotho, line 38: Will this (and similar) conflict with @DavidSeptimus-Klotho's work? If so, probably just need to coordinate.

yuval-klotho (approved):

.github/workflows/run-integ-tests.yaml
jhsinger-klotho: well its broken because our current release branch is broken mainly. So we need 2 releases then can enable this and the helm one.

pkg/infra/pulumi_aws/iac/load_balancing.ts
jhsinger-klotho: i mentioned i was doing this to him, he said he hadnt touched these modules yet so should be good

PR #79, opened 2023-01-13
Adds multi-language support to integration test workflow

DavidSeptimus-Klotho:
This PR adds support for multiple languages (specifically including Python) through the following means:
- Adding a step to detect language based on the sample app directory prefix
- Modifying the sample app directory detection logic to include apps containing a `Makefile` containing an `integ-test` rule
- Making language-specific steps conditionally executed
- Adding a step to run python integraiton tests

### Standard checks

- **Backwards compatibility**: Will this break existing apps? If so, what would be the extra work required to keep them working? Np

### DIFF
--- a/.github/workflows/run-integ-tests.yaml
+++ b/.github/workflows/run-integ-tests.yaml
@@ -69,7 +69,13 @@ jobs:
           set -x
           dirs_with_tests="$(
             for d in $(find * -type d -maxdepth 0 || printf ''); do
-              jq &>/dev/null -e '.scripts."integ-test"' $d/package.json && echo "$d"
+              TESTABLE_APP=$(jq &>/dev/null -e '.scripts."integ-test"' $d/package.json && echo "$d")
+              if [[ ! -z $(grep "^integ-test:" "${d}/Makefile" 2> /dev/null) ]]; then
+                TESTABLE_APP=$d
+              fi
+              if [[ ! -z "${TESTABLE_APP}" ]]; then
+                echo $TESTABLE_APP
+              fi
             done
             exit 0 # otherwise, will fail if the last dir failed the jq match
           )"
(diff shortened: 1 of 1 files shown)


jhsinger-klotho (approved):

.github/workflows/run-integ-tests.yaml
```diff
@@ -69,7 +69,13 @@ jobs:
           set -x
           dirs_with_tests="$(
             for d in $(find * -type d -maxdepth 0 || printf ''); do
-              jq &>/dev/null -e '.scripts."integ-test"' $d/package.json && echo "$d"
+              TESTABLE_APP=$(jq &>/dev/null -e '.scripts."integ-test"' $d/package.json && echo "$d")
+              if [[ ! -z $(grep "^integ-test:" "${d}/Makefile" 2> /dev/null) ]]; then
```
yuval-klotho, line 73: Super minor, but in case you're interested:

1. `! -z` is aka `-n`
2. you can just do `if grep -q ... ; then` (with no `[[ ... ]]]`). In that mode, `grep` doesn't actually output anything (to either stdout or stderr), but instead just exits with 0 iff any lines matched

.github/workflows/run-integ-tests.yaml
```diff
@@ -166,6 +175,7 @@ jobs:
         env:
           KLOTHO_LOGIN: ${{ inputs.klotho-login }}
       - name: typescript compilation
+        if: steps.get_language.outputs.language == 'ts'
```
yuval-klotho, line 178: nice!

yuval-klotho (approved):

.github/workflows/run-integ-tests.yaml
DavidSeptimus-Klotho: thanks. I used `-s` instead of `-q` to avoid printing `File not found` for `Makefile` in typescript projects.

yuval-klotho: I didn't know about that one! nice.

yuval-klotho (approved):

PR #82, opened 2023-01-13
Fixes #81 - wrong app var name for python expose

### LINKED ISSUE #81: Python expose plugin is substituting the capability in place of the FastAPI app variable name
The Python expose plugin is substituting the capability in place of the FastAPI app variable name. This causes the dispatcher to attempt to invoke API calls against the wrong variable and causes requests to fail.

DavidSeptimus-Klotho:
Fixes #81 by changing the `AppVarName` value set in the gateway spec back from `Capability.ID` to the name of the exported variable associated with the annotated FastAPI app.

compiled output for `py-microservices` Output after change:
image
image
image

### Standard checks

### DIFF
--- a/pkg/lang/python/plugin_expose.go
+++ b/pkg/lang/python/plugin_expose.go
@@ -200,7 +200,7 @@ func (h *restAPIHandler) handleFile(f *core.SourceFile) (*core.SourceFile, error
 
 		}
 
-		var appName string
+		var appVarName string
 		app, err := h.findFastAPIAppDefinition(capNode, f)
 		if err != nil {
 			return nil, core.NewCompilerError(f, capNode, err)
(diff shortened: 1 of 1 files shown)


pkg/lang/python/plugin_expose.go
```diff
@@ -210,23 +210,23 @@ func (h *restAPIHandler) handleFile(f *core.SourceFile) (*core.SourceFile, error
...
 		}
 
-		appName = app.Identifier.Content(f.Program())
+		appVarName = app.Identifier.Content(f.Program())
 		h.RootPath = app.RootPath
 
 		gwSpec := gatewaySpec{
 			FilePath:   f.Path(),
-			AppVarName: cap.ID,
+			AppVarName: appVarName,
```
ewucc, line 218: I think this is going to break the resource validation though. iirc, the resource validation is looking for a gateway with a given gateway ID as the name

ewucc: So that might have been why this became an issue, the name was changed to fix the resource validation issue but then caused a regression w/ proxy calls

jhsinger-klotho, line 218: isnt this going to result in validation to fail again. On line 130 we call 

```python
		gw := core.NewGateway(spec.AppVarName)
```

so then wont this not match our ID? other gatewayspecs have an id field so we can track it so i dont think this prevents the warning from popping up

jhsinger-klotho: yeah same comment as below

DavidSeptimus-Klotho: yeah, I just went back and added `gatewayId` to the spec and updated this line to use that value.

jhsinger-klotho (approved):

PR #85, opened 2023-01-13
fix scenario where env vars are empty

yuval-klotho:
When the env var was empty, the end of the method call was `, )` instead of `,[])`.

### Standard checks

- **Unit tests**: not unit tested
- **Docs**: n/a

### DIFF
--- a/pkg/infra/pulumi_aws/index.ts.tmpl
+++ b/pkg/infra/pulumi_aws/index.ts.tmpl
@@ -112,7 +112,7 @@ export = async () => {
     {{range $unit := .ExecUnits -}}
 
     {{- if eq $unit.Type "fargate"}}
-    cloudLib.createEcsService("{{$unit.Name}}",{{jsonPretty $unit.Params | indent 4}} as Partial<awsx.ecs.Container>, {{- if $unit.EnvironmentVariables}} {{jsonPretty $unit.EnvironmentVariables | indent 4}} {{else}} [] {{end}});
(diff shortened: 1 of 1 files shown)


DavidSeptimus-Klotho (approved):

pkg/infra/pulumi_aws/index.ts.tmpl
```diff
@@ -112,7 +112,7 @@ export = async () => {
     {{range $unit := .ExecUnits -}}
 
     {{- if eq $unit.Type "fargate"}}
-    cloudLib.createEcsService("{{$unit.Name}}",{{jsonPretty $unit.Params | indent 4}} as Partial<awsx.ecs.Container>, {{- if $unit.EnvironmentVariables}} , {{jsonPretty $unit.EnvironmentVariables | indent 4}} {{end}});
+    cloudLib.createEcsService("{{$unit.Name}}",{{jsonPretty $unit.Params | indent 4}} as Partial<awsx.ecs.Container>, {{- if $unit.EnvironmentVariables}} {{jsonPretty $unit.EnvironmentVariables | indent 4}} {{else}} [] {{end}});
```
gordon-klotho, line 115: Seems like the library can handle the `undefined` fine:
```ts
    generateExecUnitEnvVars(
        execUnitName: string,
        env_vars?: any[]
    )
```

gordon-klotho: I'm saying the previous version should be fine since the `, ` is also included in the original if statement.

yuval-klotho: yep, it did! nice catch.

gordon-klotho (approved):

PR #86, opened 2023-01-13
pubsub would put null instead of empty array

jhsinger-klotho:
• Does it relate to or fix any issue? closes pro #23

we would fill in as null if theres no subscribers or providers breaking our pulumi code

### Standard checks


yuval-klotho (approved):
FWM. I wonder though if it'd be simpler to change `createTopic` to allow for those args being `null`? I think in general as I read the file, it's easier for me to read logic branches in typescript rather than ones in go templating.

PR #89, opened 2023-01-13
refactor main.go

yuval-klotho:
This just wraps all of the stuff into a lightweight struct that lets you configure various aspects of main. This lets us create other "editions" of klotho (and specifically pro) without having to maintain a separate copy of this main.go, which can introduce drift.

The `main.go` changes contain a bunch of `-` (and corresponding `+` in `klothomain.go`), but it's pretty much a lift-and-shift. A few places had some `foo` turning into a `km.Foo` or similar, but that's about it.

After that, just a bit of light refactoring/moving to untangle a couple cyclic dependencies in analytics and cli.

### Standard checks

- **Unit tests**: not unit tested; verified manually on sample-apps/ts-cloudfs
- **Docs**: n/a


cmd/klotho/main.go
```diff
@@ -21,9 +21,10 @@ import (
 	"go.uber.org/zap/zapcore"
 )
 
-var root = &cobra.Command{
-	Use:  "klotho [path to source]",
-	RunE: run,
+type KlothoMain struct {
```
gordon-klotho, line 24: This being in `package main` means it wouldn't be reusable, so maybe move this into `package cli` package?

pkg/cli/klothomain.go
```diff
@@ -0,0 +1,315 @@
...
+	flags.StringVar(&cfg.appName, "app", "", "Application name")
+	flags.StringVarP(&cfg.provider, "provider", "p", "", fmt.Sprintf("Provider to compile to. Supported: %v", "aws"))
+	flags.BoolVar(&cfg.strict, "strict", false, "Fail the compilation on warnings")
+	flags.BoolVar(&cfg.disableLogo, "disable-logo", false, "Disable printing the Klotho logo")
+	flags.BoolVar(&cfg.uploadSource, "upload-source", false, "Upload the compressed source folder for debugging")
+	flags.BoolVar(&cfg.internalDebug, "internalDebug", false, "Enable debugging for compiler")
+	flags.BoolVar(&cfg.version, "version", false, "Print the version")
+	flags.BoolVar(&cfg.update, "update", false, "update the cli to the latest version")
+	flags.StringVar(&cfg.login, "login", "", "Login to Klotho with email. For anonymous login, use 'local'")
+	_ = flags.MarkHidden("internalDebug")
```
jhsinger-klotho, line 92: this will no longer be hidden really, but i guess it really isnt today

jhsinger-klotho (approved):

pkg/cli/klothomain.go
yuval-klotho: I didn't quite understand that bit, actually. In what way was it hidden before?

This code shouldn't have any functional changes, but I can verify — I'm just not sure what the user-visible thing to verify is.

jhsinger-klotho: just if they do --help, but if they look at the source code they would see it no matter what

PR #91, opened 2023-01-13
making pubsub idempotent and other fixes

jhsinger-klotho:
• Does it relate to or fix any issue? closes https://github.com/klothoplatform/klotho-pro/issues/26

Pubsub was not idempotent or able to find any modules not in the base path of klotho config. We also did not have the names lined up between what our infra launched and what code expected.

Needed to add proxy files at the end so the plugin doesnt reprocess them and see more subscribers/publishers than intended.

Need to make sure we arent overwriting our core resources.

### Standard checks

- **Backwards compatibility**: Will this break existing apps? If so, what would be the extra work required to keep them working? Well it wouldnt have worked now, so no because the naming of infra will be different

### DIFF
--- a/pkg/infra/pulumi_aws/deploylib.ts
+++ b/pkg/infra/pulumi_aws/deploylib.ts
@@ -615,12 +615,12 @@ export class CloudCCLib {
 
     createTopic(
         path: string,
-        varName: string,
+        name: string,
         event: string,
         publishers: string[],
         subscribers: string[]
     ): aws.sns.Topic {
-        let topic = `${this.name}_${varName}_${event}`
+        let topic = `${this.name}_${name}_${event}`
         if (topic.length > 256) {
             const hash = crypto.createHash('sha256')
             hash.update(topic)
--- a/pkg/infra/pulumi_aws/deploylib.ts
+++ b/pkg/infra/pulumi_aws/deploylib.ts
@@ -620,7 +620,7 @@ export class CloudCCLib {
         publishers: string[],
         subscribers: string[]
     ): aws.sns.Topic {
-        let topic = `${this.name}_${path.replace(/[^0-9a-zA-Z_-]/g, '-')}_${varName}_${event}`
+        let topic = `${this.name}_${varName}_${event}`
         if (topic.length > 256) {
             const hash = crypto.createHash('sha256')
             hash.update(topic)
--- a/pkg/lang/javascript/plugin_pubsub.go
+++ b/pkg/lang/javascript/plugin_pubsub.go
@@ -24,7 +24,8 @@ type (
 		result *core.CompilationResult
 		deps   *core.Dependencies
 
-		emitters map[VarSpec]*emitterValue
+		emitters             map[VarSpec]*emitterValue
+		proxyGenerationCalls []proxyGenerationCall
 	}
 
 	emitterValue struct {
@@ -36,6 +37,12 @@ type (
 	emitterUsage struct {
 		filePath string
 		event    string
+		unitId   string
+	}
+
+	proxyGenerationCall struct {
+		filepath string
+		emitters []EmitterSubscriberProxyEntry
 	}
 )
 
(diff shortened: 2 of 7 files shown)


pkg/infra/pulumi_aws/deploylib.ts
```diff
@@ -620,7 +620,7 @@ export class CloudCCLib {
         publishers: string[],
         subscribers: string[]
     ): aws.sns.Topic {
-        let topic = `${this.name}_${path.replace(/[^0-9a-zA-Z_-]/g, '-')}_${varName}_${event}`
+        let topic = `${this.name}_${varName}_${event}`
```
gordon-klotho, line 623: Why remove the path? Is `varName` not the emitter's variable name anymore (ie, is it the ID)? Didn't this need to match with something in the runtime as well?

pkg/lang/javascript/plugin_pubsub.go
```diff
@@ -61,41 +68,45 @@ func (p Pubsub) Transform(result *core.CompilationResult, deps *core.Dependencie
...
+				}
+				p.emitters[spec] = &emitterValue{
+					Resource: &resource,
+				}
 			}
 		}
+
+		err := p.processFiles(unit)
+		if err != nil {
+			errs.Append(err)
```
gordon-klotho, line 84: If there's an error processing the file, do we want to `continue` the for loop?

pkg/lang/javascript/plugin_pubsub.go
```diff
@@ -360,7 +375,14 @@ func (p *Pubsub) generateEmitterDefinitions() (err error) {
...
 func findTopics(f *core.SourceFile, spec VarSpec, query string, methodName string) (topics []string) {
-	varName := findVarName(f, spec)
+	relPath, _ := filepath.Rel(filepath.Dir(f.Path()), spec.DefinedIn)
+	newSpec := VarSpec{
+		DefinedIn:    relPath,
+		InternalName: spec.InternalName,
+		VarName:      spec.VarName,
+	}
+
+	varName := findVarName(f, newSpec)
```
gordon-klotho, line 385: This seems like a shortcoming of `findVarName` that should be fixed inside, not outside.

pkg/infra/pulumi_aws/deploylib.ts
jhsinger-klotho: so the path isnt in the runtime and i believe that varName is actually the ID and not truly the varName. Let me double check because if it is the varname we could have conflicts. I can also rename the variable to pubsubID so it makes more sense

pkg/lang/javascript/plugin_pubsub.go
```diff
@@ -203,9 +215,10 @@ func (p *Pubsub) processFiles(unit *core.ExecutionUnit) error {
 			}
 		}
 
-		if err := p.generateProxies(f.Path(), emitters); err != nil {
-			errs.Append(err)
-		}
+		p.proxyGenerationCalls = append(p.proxyGenerationCalls, proxyGenerationCall{
```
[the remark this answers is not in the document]
jhsinger-klotho: yup, will update

pkg/lang/javascript/plugin_pubsub.go
jhsinger-klotho: will move this logic in there and panic on the relPAth error. Otherwise we need a major refactor of finding every method that uses findVarName

gordon-klotho: Realistically, relPath only errors when one of the paths is absolute, which we should never have. The `panic` is like an assert that makes sure that's true.

pkg/provider/infra_template.go
```diff
@@ -34,7 +34,7 @@ type (
 		Subscribers []core.ResourceKey
 		Path        string
 		EventName   string
-		VarName     string
```
gordon-klotho, line 37: This change would require a change to the template, right?

gordon-klotho (changes requested):

pkg/lang/javascript/plugin_pubsub.go
yuval-klotho: It's been a while, so I forget exactly what all the inputs to `findVarMean` — but doesn't moving `relPath, err := filepath.Rel(filepath.Dir(f.Path()), spec.DefinedIn)` into it make it more language-specific? There are languages where relative imports aren't a thing.

gordon-klotho (approved):

yuval-klotho:
Are there tests (unit or integration) that we can add to prove the problem is fixed?

pkg/lang/javascript/plugin_pubsub.go
gordon-klotho: It is indeed language specific, but the var finder is in the javascript package so I'd expect it to be.

yuval-klotho: 🤦‍♂️ So it is. For some reason I thought it was in a language-agnostic package (and I didn't even think to double check that assumption in the path). Sorry — carry on! :)

jhsinger-klotho: yeah i just realized, wouldnt this cause issues for golang?

jhsinger-klotho: oh wait this is js specific

PR #94, opened 2023-01-16
longer app name required for our integ tests

jhsinger-klotho:

### Standard checks


yuval-klotho (approved):
lgmt. Might be nice to pull it out to a const, but not necessary.

PR #100, opened 2023-01-18
add the ability to place exec unit in private or public subnet

### LINKED ISSUE #28: Design a way to specify which subnet a resource should go on (public/private)
Longer term, users should be able to create any number of subnets of various types, for now, we create public/private and the goal is to pick between these 2.

If it's in public, we need to assign it a public IP

jhsinger-klotho:
• Does it relate to or fix any issue? closes #28 

adds network_placement for exec units to allow the user to choose if they should live in public or private subnet. Trivial for most things except kubernetes of course.

### Standard checks

- **Unit tests**: Any special considerations?  integ test instead https://github.com/klothoplatform/sample-apps-pro/pull/10
- **Docs**: Do we need to update any docs, internal or public? https://github.com/klothoplatform/docs/pull/150

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -42,6 +42,7 @@ type (
 
 	ExecutionUnit struct {
 		Type             string            `json:"type" yaml:"type" toml:"type"`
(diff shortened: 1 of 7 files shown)


yuval-klotho (approved):
lgtm. Would be nice to get a sample-apps test for this, too.

PR #101, opened 2023-01-18
allow for configurable update stream

### LINKED ISSUE #78: Update Klotho CLI to specify release stream to update from

yuval-klotho:
This commit has a few components:

1. Some YAML utils: one for inserting values into generic YAML using dot-notation keys, and one for validating that YAML can be deserialized as a given object type.
2. Introduces a new ~/.klotho/options.yaml for CLI options. These are strictly typed within the code, but we're more lenient in terms of configs we allow setting. In particular, we allow setting configs that we don't yet know about; this can be helpful in cases where you want to set a config that's available in a subsequent version, before you upgrade to that version. We do warn on this though: in manual testing, I had done "upgrade.stream" instead of "update.stream" and was confused why this didn't work, so I figured I'd help out the next person who runs into that.
3. KlothoMain introduces a new flag, `--set-option name=value`. You can use it on its own, or inline it with `--update`.

I'm also getting rid of updater.go's DefaultStream var, because that really needs to be specified only by the KlothoMain instantiation.

This resolves #78

### Standard checks

- **Unit tests**: Note that I added a slightly ugly `mockableLog` to cli_options.go, so that I could test the warning messages. I consider these to be important from a usability perspective, but I'm open to opinions here — especially on how else to do it (would a func be better? or I could actually tap into the zap logger — it's possible to do that, though it's a bit messy)
- **Docs**: This introduces new functionality. We'd probably want to document it at some point, but not yet. It is available in the `--help` text

### DIFF
--- a/pkg/cli/cli_options.go
+++ b/pkg/cli/cli_options.go
@@ -21,10 +21,6 @@ type (
 	UpdateOptions struct {
 		Stream string `yaml:",omitempty"`
 	}
-
-	mockableLog interface {
-		Warn(string, ...zap.Field)
-	}
 )
 
 func ReadOptions() (Options, error) {
@@ -48,7 +44,10 @@ func SetOptions(options map[string]string) error {
 	if err != nil {
 		return err
 	}
-	optionsYaml, err = setOptions(optionsYaml, options, zap.L())
+	optionsYaml, err = setOptions(optionsYaml, options)
+	if err != nil {
+		return nil
+	}
 
 	err = os.WriteFile(filePath, optionsYaml, 0600)
 	if err != nil {
@@ -59,14 +58,15 @@ func SetOptions(options map[string]string) error {
 
 // setOptions inserts the given options into the yaml, validating along the way that the options still form a valid
 // Options.
-func setOptions(optionsYaml []byte, options map[string]string, logger mockableLog) ([]byte, error) {
+func setOptions(optionsYaml []byte, options map[string]string) ([]byte, error) {
+	logger := zap.S()
 	// First, a warning if the original file isn't valid
 	if err := yaml_util.CheckValid[Options](optionsYaml, yaml_util.Lenient); err != nil {
 		return nil, errors.Wrap(err, "existing options file is invalid")
 	} else if warns := yaml_util.CheckValid[Options](optionsYaml, yaml_util.Strict); warns != nil {
 		logger.Warn(`Existing options contain extra parameters:`)
 		for _, e := range yaml_util.YamlErrors(warns) {
-			logger.Warn(fmt.Sprintf(`▸ %s`, e))
+			logger.Warnf(`▸ %s`, e)
 		}
 	}
 
--- a/pkg/cli/cli_options.go
+++ b/pkg/cli/cli_options.go
@@ -1,0 +1,128 @@
+package cli
+
+import (
+	"fmt"
+	"github.com/klothoplatform/klotho/pkg/cli_config"
+	"github.com/klothoplatform/klotho/pkg/yaml_util"
+	"github.com/pkg/errors"
+	"go.uber.org/zap"
+	"gopkg.in/yaml.v3"
+	"io/fs"
+	"os"
+)
+
+const configFileName = "options.yaml"
+
+type (
+	Options struct {
+		Update UpdateOptions `yaml:",omitempty"`
+	}
+
+	UpdateOptions struct {
+		Stream string `yaml:",omitempty"`
+	}
+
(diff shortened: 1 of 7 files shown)


pkg/cli/cli_options_test.go
```diff
@@ -0,0 +1,185 @@
...
+			assert.Equal(tt.expectYaml, string(result))
+		})
+	}
+}
+
+type mockedLogger struct {
+	warns []string
+}
+
+func (ml *mockedLogger) Warn(msg string, fields ...zap.Field) {
```
jhsinger-klotho, line 85: https://github.com/klothoplatform/klotho/blob/main/pkg/validation/validation_test.go#L503

I use a zap observer to test log output if you think its less ugly. I think its personally a bit cleaner than this implementation, but up to you

pkg/cli/klothomain.go
```diff
@@ -24,14 +24,12 @@ import (
 )
 
 type KlothoMain struct {
-	UpdateStream     string
-	Version          string
-	VersionQualifier string
-	PluginSetup      func(*PluginSetBuilder) error
+	DefaultUpdateStream string
```
jhsinger-klotho, line 27: nit: Is this the default if they have set it in their config?

pkg/yaml_util/yaml_util.go
```diff
@@ -0,0 +1,129 @@
...
+	// step 2a
+	segments := strings.Split(optionPath, ".")
+	var topNode *yaml.Node
+	if len(tree.Content) == 0 {
+		topNode = &yaml.Node{Kind: yaml.MappingNode}
+	} else {
+		topNode = tree.Content[0] // the tree's root is a DocumentNode; we assume one document
+	}
+	setOptionAtNode := topNode
+	for _, segment := range segments[:len(segments)-1] {
```
jhsinger-klotho, line 49: for 2a) are we finding the value at the non leaf path? ie up to bar in this scenario? I assume thats the intention, just making sure

pkg/cli/cli_options.go
```diff
@@ -0,0 +1,128 @@
...
+// Options.
+func setOptions(optionsYaml []byte, options map[string]string) ([]byte, error) {
+	logger := zap.S()
+	// First, a warning if the original file isn't valid
+	if err := yaml_util.CheckValid[Options](optionsYaml, yaml_util.Lenient); err != nil {
+		return nil, errors.Wrap(err, "existing options file is invalid")
+	} else if warns := yaml_util.CheckValid[Options](optionsYaml, yaml_util.Strict); warns != nil {
+		logger.Warn(`Existing options contain extra parameters:`)
+		for _, e := range yaml_util.YamlErrors(warns) {
+			logger.Warnf(`▸ %s`, e)
```
jhsinger-klotho, line 69: why only warn instead of raising actual errors? Wouldnt this signal an incorrect usage of the set-options flag?

pkg/cli/klothomain.go
yuval-klotho: I think so? It's the default, which they can then override in their options. (It's actually not overridable on a one-invocation-only basis, only via persisted options.)

(As an aside: I'm trying to use "options" for the CLI-wide stuff [which is currently only the update stream], to differentiate it from the klotho project configs `-c`.)

pkg/cli/cli_options.go
yuval-klotho: It _probably_ signals incorrect usage, yes. But the incorrectness is just that it won't have any effect, not that it will have the wrong effect. So I figure a nice yellow warning is sufficient for that.

From previous experience, I've found that occasionally you want the ability to set a config that doesn't yet exist, which will take effect on the next update. For example, the next version of the tool may have an automatic migration that you want to modify. In cases like this, being able to specify a not-yet-existent config can be very useful. This approach also means that if you temporarily downgrade your version (once we have version pinning), you won't break your options file if you had set an option in 0.5.21 that doesn't exist in the 0.5.20 you've downgraded to.

Basically, I've found that being strict about "no unknown options" can get people into sticky situations, and that a nice clear warning is sufficient for catching typos.

pkg/yaml_util/yaml_util.go
yuval-klotho: Correct. We want the node right before the last segment, and then in 2b we set its value at that last segment.

So in `fizz.buzz.foo = 123`, the 2a loop gets you the `fizz.buzz` node, and then 2b sets `*buzzNode["foo"] => 123`

jhsinger-klotho (approved):

PR #108, opened 2023-01-18
Resource name sanitization

### LINKED ISSUE #49: Use appname namespace for all resources
There's a few that aren't properly namespace'd. Do a sweep through all of the resources, but some I know at time of writing:

- API Gateway
- Log groups

We should also consider having a santize method purely for the length of the resource to ensure we will be able to create it.

DavidSeptimus-Klotho:
Resolves #49 (minus the actual namespacing piece)

This PR adds sanitization for the identifiers of AWS resources specified for creation via Klotho's IAC layer. This change should not change any happy path identifiers (resource names or Pulumi URNs)

Resource names are generated using tagged string templates as in the following examples using the `sanitized()` tag function:
```typescript
let appName = 'my-very-long-app-name'
let resourceId = 'my-load-balancer'
​
// ensure that the output is a valid load balancer name
// from left to right, hash appName and then resourceId if the name is too long
sanitized(loadBalancer.nameValidation())`${h(appName)}-${h(resourceId)}-lb`
// 2d958-my-load-balancer-lb
​
// prioritize hashing resourceId before appName since its priority is 1 (default is 100)
sanitized(loadBalancer.nameValidation())`${h(appName)}-${h(resourceId, 1)}-lb`
// my-very-long-app-name-0ea7a-lb
```

When shortening resource names, raw text or template expressions of type `string` are not modified. Expressions of type `Component` are shortened from left to right based on priority (default priority is `100`) using the `Component` instance's configured `ShorteningStrategy`. In the examples above, the `h()` function returns a `Component` with a `ShorteningStrategy` that converts the supplied text into a 5 character truncated SHA256 hash.  

The process of generating a valid resource name is as follows:
1) Shorten `Component`s from left to right ranked by priority until the name's length <= `maxLength`.
2) Sanitize the generated string in one or more passes (up to 5 passes by default):
   1) Truncate the string if it's longer than max length (arguably, we might want this to be an error).
   2) Throw an error if the string is shorter than the min length.
   3) Apply the supplied `SanitizationRule`s and capture any violations.
   4) For any violations, sequentially apply the violated rules' supplied `FixFunc` (if present) to resolve any issues.
   5) Revalidate the string if it's been fixed
   6) Return the valid string or start another pass if the string is still invalid
3) return the sanitized result or throw an exception if any constraint violations are present in the result.

### Standard checks

- **Unit tests**: Any special considerations? We don't have a pattern for testing IAC-related code at the moment, but we should really figure that out.
- **Backwards compatibility**: Will this break existing apps? This PR aims to avoid changing "happy path" resource names so there should be no issue there.

### DIFF
--- a/pkg/infra/pulumi_aws/iac/elasticache.ts
+++ b/pkg/infra/pulumi_aws/iac/elasticache.ts
@@ -1,7 +1,8 @@
 import * as aws from '@pulumi/aws'
 import * as pulumi from '@pulumi/pulumi'
-import { PulumiFn } from '@pulumi/pulumi/automation'
-import { Resource, CloudCCLib } from '../deploylib'
+import { Resource } from '../deploylib'
+import * as validators from './sanitization/aws/elasticache'
+import { sanitized } from './sanitization/sanitizer'
 
 const ELASTICACHE_ENGINE = 'redis'
 
@@ -46,7 +47,10 @@ export const setupElasticacheCluster = (
         retentionInDays: 0,
     })
 
-    const clusterName = sanitizeClusterName(appName, dbName)
+    // TODO: look into removing sanitizeClusterName when making other breaking changes to resource names
+    const clusterName = sanitized(
+        validators.cacheCluster.cacheClusterIdValidation()
+    )`${sanitizeClusterName(appName, dbName)}`
     // create the db resources
     const redis = new aws.elasticache.Cluster(
         clusterName,
--- a/pkg/infra/pulumi_aws/iac/load_balancing.ts
+++ b/pkg/infra/pulumi_aws/iac/load_balancing.ts
@@ -8,7 +8,7 @@ import {
     TargetGroupAttachmentArgs,
 } from '@pulumi/aws/lb'
 import { ListenerRuleArgs } from '@pulumi/aws/alb'
-import { h, sanitized } from './sanitization/sanitizer'
+import { hash as h, sanitized } from './sanitization/sanitizer'
 import { CloudCCLib } from '../deploylib'
 
 export interface Route {
--- a/pkg/infra/pulumi_aws/iac/load_balancing.ts
+++ b/pkg/infra/pulumi_aws/iac/load_balancing.ts
@@ -1,22 +1,13 @@
-import { Region } from '@pulumi/aws'
 import * as aws from '@pulumi/aws'
-import * as awsx from '@pulumi/awsx'
-import * as k8s from '@pulumi/kubernetes'
-
-import * as pulumi from '@pulumi/pulumi'
-import * as sha256 from 'simple-sha256'
-import * as fs from 'fs'
-import * as requestRetry from 'requestretry'
-import * as crypto from 'crypto'
-
-import * as eks from '@pulumi/eks'
+import * as validators from './sanitization/aws/elb'
 import {
     ListenerArgs,
-    TargetGroupArgs,
     LoadBalancerArgs,
+    TargetGroupArgs,
     TargetGroupAttachmentArgs,
 } from '@pulumi/aws/lb'
 import { ListenerRuleArgs } from '@pulumi/aws/alb'
+import { h, sanitized } from './sanitization/sanitizer'
 import { CloudCCLib } from '../deploylib'
 
 export interface Route {
@@ -176,9 +167,12 @@ export class LoadBalancerPlugin {
         params: LoadBalancerArgs
     ): aws.lb.LoadBalancer => {
         let lb: aws.lb.LoadBalancer
+        let lbName = sanitized(validators.loadBalancer.nameValidation())`${h(appName)}-${h(
+            resourceId
+        )}`
         switch (params.loadBalancerType) {
             case 'application':
-                lb = new aws.lb.LoadBalancer(`${appName}-${resourceId}`, {
+                lb = new aws.lb.LoadBalancer(lbName, {
                     internal: params.internal || false,
                     loadBalancerType: 'application',
                     securityGroups: params.securityGroups,
@@ -188,7 +182,7 @@ export class LoadBalancerPlugin {
                 })
                 break
             case 'network':
-                lb = new aws.lb.LoadBalancer(`${appName}-${resourceId}`, {
+                lb = new aws.lb.LoadBalancer(lbName, {
                     internal: params.internal || true,
                     loadBalancerType: 'network',
                     subnets: params.subnets,
@@ -237,12 +231,15 @@ export class LoadBalancerPlugin {
         params: TargetGroupArgs
     ): aws.lb.TargetGroup => {
         let targetGroup: aws.lb.TargetGroup
+        let tgName = sanitized(validators.targetGroup.nameValidation())`${h(appName)}-${h(
+            execUnitName
+        )}`
         if (params.targetType != 'lambda' && !(params.port && params.protocol)) {
             throw new Error('Port and Protocol must be specified for non lambda target types')
         }
         switch (params.targetType) {
             case 'ip':
-                targetGroup = new aws.lb.TargetGroup(`${appName}-${execUnitName}`, {
+                targetGroup = new aws.lb.TargetGroup(tgName, {
                     port: params.port,
                     protocol: params.protocol,
                     targetType: 'ip',
@@ -251,7 +248,7 @@ export class LoadBalancerPlugin {
                 })
                 break
             case 'instance':
-                targetGroup = new aws.lb.TargetGroup(`${appName}-${execUnitName}`, {
+                targetGroup = new aws.lb.TargetGroup(tgName, {
                     port: params.port,
                     protocol: params.protocol,
                     vpcId: params.vpcId,
@@ -259,7 +256,7 @@ export class LoadBalancerPlugin {
                 })
                 break
             case 'alb':
-                targetGroup = new aws.lb.TargetGroup(`${appName}-${execUnitName}`, {
+                targetGroup = new aws.lb.TargetGroup(tgName, {
                     targetType: 'alb',
                     port: params.port,
                     protocol: params.protocol,
@@ -269,7 +266,7 @@ export class LoadBalancerPlugin {
                 })
                 break
             case 'lambda':
-                targetGroup = new aws.lb.TargetGroup(`${appName}-${execUnitName}`, {
+                targetGroup = new aws.lb.TargetGroup(tgName, {
                     targetType: 'lambda',
                     tags: params.tags,
                 })
--- a/pkg/infra/pulumi_aws/iac/load_balancing.ts
+++ b/pkg/infra/pulumi_aws/iac/load_balancing.ts
@@ -1,3 +1,4 @@
+import * as pulumi from '@pulumi/pulumi'
 import * as aws from '@pulumi/aws'
 import * as validators from './sanitization/aws/elb'
 import {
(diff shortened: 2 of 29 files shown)


DavidSeptimus-Klotho:
rebased and squashed old commits

ewucc (commented):
First pass, looks good in general. I think general comments
- seems like for every new entity that we support we'll need to add that set of rules. Seems like the right way to do it but yeh definitely a lot of bloat to the compiled dir since i believe these will end up there. 
- Might just be me but while it does seem comprehensive, it also seems somewhat complex of a process. Specifically around the weighting (0 -> 100) of components and in the shortening func, it does a length subtraction vs. building from most important components up and hashing once it's exceeded the length. 

Last point is also not really like a negative, just more of an observation so not a big deal if no one else has similar thoughts.

pkg/infra/pulumi_aws/iac/sanitization/sanitizer.ts
```diff
@@ -0,0 +1,197 @@
...
+
+export function h(content: string, priority: number | undefined = undefined): Component {
+    return {
+        content,
+        priority,
+        shorteningStrategy: hashComponent,
+    }
+}
+
+export function t(content: string, priority: number | undefined = undefined): Component {
```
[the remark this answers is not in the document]
DavidSeptimus-Klotho: Correct, we don't use it and I should probably drop it, but it does demonstrate the concept of using multiple strategies.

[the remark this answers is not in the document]
DavidSeptimus-Klotho: I should probably also renamed `h` -> `hash` or something and then alias the import.

DavidSeptimus-Klotho:
> First pass, looks good in general. I think general comments
> 
> * seems like for every new entity that we support we'll need to add that set of rules. Seems like the right way to do it but yeh definitely a lot of bloat to the compiled dir since i believe these will end up there.

That's true. Ideally, this functionality will move into the compiler at some point. Another option is running it through a build tool to merge all the sanitization files into a single file that we include in the output.

jhsinger-klotho:
What do we do if we have like multiple clusters and install a controller on both. Is it on the plugin to label that resourceId cluster-resource?

pkg/infra/pulumi_aws/iac/load_balancing.ts
```diff
@@ -32,10 +23,13 @@ export class LoadBalancerPlugin {
...
     ): aws.lb.LoadBalancer => {
         let lb: aws.lb.LoadBalancer
+        let lbName = sanitized(validators.loadBalancer.nameValidation())`${h(appName)}-${h(
+            resourceId
+        )}`
         switch (params.loadBalancerType) {
             case 'application':
                 lb = new aws.lb.LoadBalancer(`${appName}-${resourceId}-alb`, {
-                    name: `${appName}-${resourceId}`,
+                    name: lbName,
```
jhsinger-klotho, line 32: also a general thing, we likely shouldnt set actual names of resources because then they cant get replaced if they change. We should be doing the validation on the resource id for pulumi i believe so that it autogenerates the name + some uuid for uniqueness if we ever need replacements.

let me know your thought on that, i think i added the name field, but i was going to remove it in my next pr because i noticed it could cause issues. We should likely see where else we do this

pkg/infra/pulumi_aws/iac/elasticache.ts
```diff
@@ -46,7 +47,10 @@ export const setupElasticacheCluster = (
         retentionInDays: 0,
     })
 
-    const clusterName = sanitizeClusterName(appName, dbName)
+    // TODO: look into removing sanitizeClusterName when making other breaking changes to resource names
+    const clusterName = sanitized(
```
jhsinger-klotho, line 51: same here, need to remove clustername as the actual name for this and memdb but lets maybe chat about that tomorrow

pkg/infra/pulumi_aws/iac/sanitization/sanitizer.ts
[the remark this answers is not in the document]
jhsinger-klotho: agreed with these points and then just a few separate comments ive added, which dont need to be a part of this pr, but i believe we should discuss and come up with a solution for or cut tickets

pkg/infra/pulumi_aws/iac/sanitization/aws/common.ts
```diff
@@ -0,0 +1,33 @@
...
+        return {
+            minLength: 0,
+            maxLength: 256,
+            rules: [
+                regexpMatch('', /^[\p{L}\p{Z}\p{N}_.:/=+\-@]+$/u, (n) =>
+                    n.replace(/[\p{L}\p{Z}\p{N}_.:/=+\-@]/gu, '_')
+                ),
+                {
+                    description:
+                        "The aws: prefix is prohibited for tags; it's reserved for AWS use.",
```
yuval-klotho, line 26: minor:
```suggestion
                        "The \"aws:\" prefix is prohibited for tags; it's reserved for AWS use.",
```

pkg/infra/pulumi_aws/iac/sanitization/sanitizer.ts
```diff
@@ -0,0 +1,195 @@
...
+function hashComponent(input: string, maxLength: number): string {
+    maxLength = maxLength > 5 ? 5 : maxLength
+    const hash = sha256.sync(input)
+    return hash.substring(0, maxLength <= input.length ? maxLength : input.length)
+}
+
+export function regexpMatch(
+    description: string,
+    pattern: RegExp,
+    fix: FixFunc | undefined = undefined
```
yuval-klotho, line 95: small suggestion: I think all of the calls to this are basically:

```
regexpMatch(desc, /^P$/, (n) => n.replace(/P/g, R))
```

... where `P` is some pattern and `R` is a replacement string. For example:

```
regexpMatch('', /^[\w-]+$/, (n) => n.replace(/[^\w-]/g, '-'))
```

If that's true, I think you could simplify this to just take the pattern and a replace char, and generate the `validate` and `FixFunc` from them:

```
regexpMatch('', /[^\w-]/, '-'))
```

yuval-klotho (approved):
lgtm, other than the merge conflicts and a simplification suggestion. May want to get @jhsinger-klotho's quick glance too, since he took a more in-depth look a couple weeks back.

jhsinger-klotho:
yeah, LGTM and if the test are passing definitely a good starting point. The name issue looks to be resolved but we can continue to remove those if we find more

pkg/infra/pulumi_aws/plugin_iac.go
```diff
@@ -152,6 +154,50 @@ func (p Plugin) Transform(result *core.CompilationResult, deps *core.Dependencie
...
+		var unreadEntries []string
+		dirContents, err := files.ReadDir(dir)
+
+		addEntries := func(parentDir string, entries []fs.DirEntry) {
+			for _, entry := range entries {
+				dirSuffix := ""
+				if entry.IsDir() {
+					dirSuffix = "/"
+				}
+				path := strings.TrimSuffix(parentDir, "/") + "/" + entry.Name() + dirSuffix
```
gordon-klotho, line 167: nit: use `path.Join` instead of `+ "/" +` to be more platform agnostic

pkg/infra/pulumi_aws/plugin_iac.go
```diff
@@ -152,6 +154,50 @@ func (p Plugin) Transform(result *core.CompilationResult, deps *core.Dependencie
...
+				shouldInclude := true
+				for _, exclusion := range exclusions {
+					if shouldExclude, _ := doublestar.Match(exclusion, path); shouldExclude {
+						shouldInclude = false
```
gordon-klotho, line 171: nit: we can `break` here since nothing will set it back to true

pkg/infra/pulumi_aws/plugin_iac.go
```diff
@@ -152,6 +154,50 @@ func (p Plugin) Transform(result *core.CompilationResult, deps *core.Dependencie
...
+					unreadEntries = append(unreadEntries, path)
+				}
+			}
+		}
+		addEntries(dir, dirContents)
+
+		for len(unreadEntries) > 0 {
+			if err != nil {
+				return
+			}
```
gordon-klotho, line 184: This won't propagate the error like `addFile` since `err` is defined within the scope of this function (at the `files.ReadDir` line).

pkg/infra/pulumi_aws/plugin_iac.go
```diff
@@ -152,6 +154,50 @@ func (p Plugin) Transform(result *core.CompilationResult, deps *core.Dependencie
...
+				childEntries, err = files.ReadDir(strings.TrimSuffix(entry, "/"))
+				addEntries(entry, childEntries)
+
+			} else {
+				addFile(entry)
+			}
+		}
+	}
+
+	addDir("iac/sanitization", "**/*.{test,spec}.{ts,js}")
```
gordon-klotho, line 199: Nice, I like this

pkg/infra/pulumi_aws/plugin_iac.go
```diff
@@ -152,6 +154,50 @@ func (p Plugin) Transform(result *core.CompilationResult, deps *core.Dependencie
 	addFile("iac/k8s/add_ons/external_dns/index.ts")
 	addFile("iac/k8s/add_ons/index.ts")
 
+	addDir := func(dir string, exclusions ...string) {
+		var unreadEntries []string
+		dirContents, err := files.ReadDir(dir)
```
gordon-klotho, line 159: Should this do something with `err` if it's not nil?

pkg/infra/pulumi_aws/plugin_iac.go
```diff
@@ -152,6 +154,50 @@ func (p Plugin) Transform(result *core.CompilationResult, deps *core.Dependencie
...
+		for len(unreadEntries) > 0 {
+			if err != nil {
+				return
+			}
+
+			entry := unreadEntries[0]
+			unreadEntries = unreadEntries[1:]
+			if strings.HasSuffix(entry, "/") {
+				var childEntries []os.DirEntry
+				childEntries, err = files.ReadDir(strings.TrimSuffix(entry, "/"))
```
gordon-klotho, line 190: LIikewise, handling this `err`

pkg/infra/pulumi_aws/plugin_iac.go
DavidSeptimus-Klotho: It would be even better if those spec files existed. 😅

pkg/infra/pulumi_aws/plugin_iac.go
DavidSeptimus-Klotho: The separator is always a forward slash when reading from `embed.FS`
From the docs for `go:embed`: 
> The //go:embed directive accepts multiple space-separated patterns for brevity, but it can also be repeated, to avoid very long lines when there are many patterns. The patterns are interpreted relative to the package directory containing the source file. **_The path separator is a forward slash, even on Windows systems._** Patterns may not contain ‘.’ or ‘..’ or empty path elements, nor may they begin or end with a slash. To match everything in the current directory, use ‘*’ instead of ‘.’. To allow for naming files with spaces in their names, patterns can be written as Go double-quoted or back-quoted string literals.
https://pkg.go.dev/embed#:~:text=The%20//go%3Aembed%20directive%20accepts,back%2Dquoted%20string%20literals.

pkg/infra/pulumi_aws/iac/sanitization/sanitizer.ts
DavidSeptimus-Klotho: That's true for a lot of them, but definitely not all.

PR #110, opened 2023-01-19
Use AWS_ENDPOINT_URL if specified

gordon-klotho:
This PR adds support for localstack as discovered by their CTO in https://github.com/whummer/localstack-klotho-quickstart/tree/main/patches

### Standard checks

### DIFF
--- a/pkg/lang/javascript/aws_runtime/_keyvalue.ts
+++ b/pkg/lang/javascript/aws_runtime/_keyvalue.ts
@@ -4,8 +4,10 @@ import _ = require('lodash')
 
 import moment = require('moment')
 
+import { AWSConfig } from './clients'
+
 import DynamoDB = require('aws-sdk/clients/dynamodb')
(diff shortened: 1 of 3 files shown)


pkg/lang/javascript/aws_runtime/_keyvalue.ts
```diff
@@ -51,13 +53,6 @@ const KVStore = new Entity(
     { table: '{{.AppName}}', client: docClient }
 )
 
-import { fs as cloudFS } from './fs'
```
yuval-klotho, line 54: unrelated cleanup, I assume?

yuval-klotho (approved):

pkg/lang/javascript/aws_runtime/_keyvalue.ts
gordon-klotho: Yeah, I figured while I was in the neighborhood I'd leave it cleaner than when I got here :)

PR #112, opened 2023-01-20
python: support for `*args`/`**kwargs` in dispatcher

yuval-klotho:
Also some fixes to the py dispatcher's handling of proxy io.

### Standard checks

- **Unit tests**: not unit tested, but manually tested using py-microservices in klothoplatform/sample-apps-pro#11
- **Docs**: no changes needed

### DIFF
--- a/pkg/lang/python/aws_runtime/dispatcher_fargate.py.tmpl
+++ b/pkg/lang/python/aws_runtime/dispatcher_fargate.py.tmpl
@@ -2,6 +2,7 @@ import logging
 import os
 import multiprocessing
 import types
+import inspect
 
 app_port = os.getenv("KLOTHO_APP_PORT", 3000)
 log_level = os.getenv("KLOTHO_LOG_LEVEL", "DEBUG").upper()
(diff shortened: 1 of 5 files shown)


pkg/lang/python/aws_runtime/dispatcher_fargate.py.tmpl
```diff
@@ -59,7 +60,14 @@ def start_proxy_server():
...
             raise Exception(f"couldn't find function: {module_name}.{function_name}")
-        result = function(*params)
+        param_args = ()
+        param_kwargs = {}
+        args_spec = inspect.getfullargspec(function)
+        if args_spec.varkw:
+            param_kwargs, params = params[-1], params[:-1]
+        if args_spec.varargs:
+            param_args, params = params[-1], params[:-1]
+        result = function(*params, *param_args, **param_kwargs)
```
DavidSeptimus-Klotho, line 70: Won't this break if you have both `*args` and `**kwargs` in the same spec? Both assignments are using `params[-1], params[:-1]` on the same list.

DavidSeptimus-Klotho: (same goes for the other dispatcher)

DavidSeptimus-Klotho (approved):

Issue #117, opened 2023-01-20
Explore Websocket API Gateways


jhsinger-klotho:
We have all of the IAC code completed to launch a single exec backed Websocket API Gateway, but i wasnt quite sure how to make them useable without parsing source code for middleware or other libraries. I dont want to throw away the code so its in pr, but this would be to make it to the point where we can surface it via config.

Today we can use type Alb to support websockets rather than api gateway

Issue #111, opened 2023-01-20
Topology and config type are not in sync


jhsinger-klotho:
Our topology uses the cloudResource.Type() which is never synced with the type in config. This is done for exec units by explicitly setting the type in code, but we should probably have a better solution than hardcoding it

PR #118, opened 2023-01-20
adding alb as a gateway type

### LINKED ISSUE #109: Support websocket APIs for EKS Through ALBs

jhsinger-klotho:

closes #109  

Adding ALB as a type for expose, which will work with lambda and EKS today (ECS still hardcodes NLBs in the code of the compiler so we need some serious cleanup there.

We add our public subnets to our SG so we can enable the delivery of ALB requests to our backend (These are the private cidr blocks of those subnets)

Because i went down the route of Websocket APIs first i had refactored the APIGateway class to use the LB Plugin in a similar manner. I would have to back out a lot of changes to deploy lib and the index.ts.tmpl file to get the ALB Functionality to work and remove the web socket funcitonality. 

The web socket functionality does allow us to create a websocket API GW and a single exec, but isnt too useable outside of that. Cut #117 to dive into that more

Also modularizes all the API Gateway code out of deploylib as a part of these changes.

### Standard checks

- **Unit tests**: Any special considerations? changed the provider unit test, will change some integ tests to
- **Docs**: Do we need to update any docs, internal or public? https://github.com/klothoplatform/docs/pull/152
- **Backwards compatibility**: Will this break existing apps? If so, what would be the extra work required to keep them working? yes, this is a new type, i do not believe our IaC should have anything which is not backwards compatible but our logic is very use case based which isnt great (We talked about it in sync and the need to move it to the compiler)

### DIFF
--- a/pkg/infra/pulumi_aws/iac/eks.ts
+++ b/pkg/infra/pulumi_aws/iac/eks.ts
@@ -654,7 +654,6 @@ export class Eks {
         }
 
         if (image) {
-            console.log(needsLoadBalancer)
             if (needsGatewayLink) {
                 if (!this.lbPlugin) {
                     throw new Error(
--- a/pkg/infra/pulumi_aws/iac/eks.ts
+++ b/pkg/infra/pulumi_aws/iac/eks.ts
@@ -17,6 +17,7 @@ import {
 } from './k8s/horizontal-pod-autoscaling'
 import { installMetricsServer } from './k8s/add_ons/metrics_server'
 import { applyChart, Value } from './k8s/helm_chart'
+import { LoadBalancerPlugin } from './load_balancing'
 
 export interface EksExecUnitArgs {
     nodeType: 'fargate' | 'node'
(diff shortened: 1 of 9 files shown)


pkg/infra/pulumi_aws/iac/eks.ts
```diff
@@ -626,64 +654,70 @@ export class Eks {
...
-
-            lib.topology.topologyIconData.forEach((resource) => {
-                if (resource.kind === Resource.gateway) {
-                    lib.topology.topologyEdgeData.forEach((edge) => {
-                        if (edge.source == resource.id && edge.target === execUnitEdgeId) {
-                            // We know that this exec unit is exposed and must create the necessary resources
-                            needsGatewayLink = true
-                        }
-                    })
+            console.log(needsLoadBalancer)
```
atorres-klo, line 657: is this old debug?

jhsinger-klotho: topology is kind of blocked, i cut klotho #111 so we can discuss and fix it

yuval-klotho (approved):

PR #122, opened 2023-01-22
only join filepath if root path exists for filref

### LINKED ISSUE #38: Only include FileRef RootConfig Path if it exists when writing
We werent properly outputting FileRefs if we arent running on . (klotho . --app blah). We made a change for search the deck to include the config path, which breaks FileRefs if they want to reference something not in the klotho project.

Gordon brought up the example: ie, a third-party plugin could do something like unit.Add(&core.FileRef{FPath: "/etc/passwd"})

We should be able to support this for other plugins

jhsinger-klotho:
• Does it relate to or fix any issue? closes #38 

This gives the option to not include the config root path for fileRef and be able to find the file correctly. 

This does give the ability to look at files outside of the klotho context, which may be useful in the future, but is the question of do we want to allow that? Is FileRef not using the content of input files really the right path?  It would then be up to the user to have proper permissions on the files in their local filesystem

### Standard checks

### DIFF
--- a/pkg/core/file_ref.go
+++ b/pkg/core/file_ref.go
@@ -21,7 +21,11 @@ func (r *FileRef) Path() string {
 }
 
 func (r *FileRef) WriteTo(w io.Writer) (int64, error) {
-	f, err := os.Open(filepath.Join(r.RootConfigPath, r.FPath))
+	pathToOpen := r.FPath
+	if r.RootConfigPath != "" {
+		pathToOpen = filepath.Join(r.RootConfigPath, r.FPath)
+	}
+	f, err := os.Open(pathToOpen)
 	if err != nil {
 		return 0, err
 	}


pkg/core/file_ref.go
```diff
@@ -21,7 +21,11 @@ func (r *FileRef) Path() string {
 }
 
 func (r *FileRef) WriteTo(w io.Writer) (int64, error) {
```
DavidSeptimus-Klotho, line 23: The issue/PR description mention wanting to look at files that aren't in the project directory, but this change rather seems to make it possible to write to any directory on the local machine.

jhsinger-klotho: well we wouldnt write to any directory on the local machine becuase we join the path with the execution unit or static units compiled path output. Its already possible, the real question is do we want to allow klotho to look at files outside of the scope passed in through the CLI and that i dont know the answer to

gordon-klotho:
> This does give the ability to look at files outside of the klotho context, which may be useful in the future, but is the question of do we want to allow that?

It was intentional that this was not possible. We may want to revisit that, but that was my original intent.

jhsinger-klotho:
> > This does give the ability to look at files outside of the klotho context, which may be useful in the future, but is the question of do we want to allow that?
> 
> It was intentional that this was not possible. We may want to revisit that, but that was my original intent.

So does this mean i should cancel this pr for now?

gordon-klotho:
> So does this mean i should cancel this pr for now?

Not sure - was this to fix a specific problem we were having, or preempting something we'll need to address (possibly in a different way, if we choose to keep the path sandboxing).

jhsinger-klotho:
> > So does this mean i should cancel this pr for now?
> 
> Not sure - was this to fix a specific problem we were having, or preempting something we'll need to address (possibly in a different way, if we choose to keep the path sandboxing).

theres no immediate problem, this was to allow fileRefs to not require the klotho config input path, which they do right now. But in terms of what we have, thats not an issue at the moment so i could cancel

PR #133, opened 2023-01-24
output pulumi version

yuval-klotho:
This could help us track down mysterious bugs, since Pulumi sometimes changes from under us.

### Standard checks

- **Unit tests**: n/a
- **Docs**: n/a


jhsinger-klotho (approved):

ewucc (approved):
I can't remember if we decided we wanted to pin the pulumi version

PR #137, opened 2023-01-25
Update klotho checks

ewucc:
I think the second stream check will result in false update messages being printed. 

Here's an example where I set my options to `oss:v0.5.19` 
```
Calling server w/ url + stream:  http://srv.klo.dev oss:v0.5.19
Current ver:  0.5.19       Latest Ver:  0.5.19
Current Stream:  oss:latest    User Stream:  oss:v0.5.19
```
If the user pins the version to the current version, it's just never going to match the stream check because the current stream is always the default value `oss:latest`

I'm pretty sure the first check for the semver is good enough since that takes the latest and converts it to the version number and compares that to the running builds version in the build file. 

Discussion: Not sure if the initial intention was to notify the user every time they run, that their pinned version is behind latest, but if that's the case, then this needs to be merged in https://github.com/klothoplatform/klotho-server/pull/27. The server was checking w/ the stream version provided instead of actual latest.

### DIFF
--- a/pkg/cli/cli_options.go
+++ b/pkg/cli/cli_options.go
@@ -129,15 +129,15 @@ func OptionOrDefault(given string, defaultValue string) string {
 	return given
 }
 
-func ShouldCheckForUpdate(given string, defaultValue string, currVersion string) bool {
-	if given == "" || given == defaultValue {
+func ShouldCheckForUpdate(updateStreamOverride string, defaultUpdateStream string, currVersion string) bool {
+	if updateStreamOverride == "" || updateStreamOverride == defaultUpdateStream {
 		return true
 	}
 
-	givenParts := strings.Split(given, ":")
-	if len(givenParts) == 2 {
-		givenVersion := givenParts[1]
-		if givenVersion != currVersion {
+	streamOverrideParts := strings.Split(updateStreamOverride, ":")
+	if len(streamOverrideParts) == 2 {
+		streamOverrideVersion := streamOverrideParts[1]
+		if streamOverrideVersion != currVersion {
 			return true
 		}
 	}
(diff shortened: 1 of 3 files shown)


yuval-klotho:
This check was there intentionally. Let's say both the latest in both oss and pro is 0.6.0, and I'm on the oss stream. If I do `klotho --set-option update.stream=pro:latest --update`, then I don't want it to say "oh you're on v0.6.0 already, nothing to do". It should recognize that because I updated streams, it always needs to do an update.

pkg/cli/cli_options.go
```diff
@@ -126,3 +128,19 @@ func OptionOrDefault(given string, defaultValue string) string {
 	}
 	return given
 }
+
+func ShouldCheckForUpdate(given string, defaultValue string, currVersion string) bool {
```
yuval-klotho, line 132: small: can we rename these args? something like
```suggestion
func ShouldCheckForUpdate(updateStreamOverride string, defaultUpdateStream string, currVersion string) bool {
```

and then similarly below:

* `givenParts` => `streamOverrideParts`
* `givenVersion` => `streamOverrideVersion`

ewucc:
Per convo w/ yuval, added the check for the stream type 

If the user pins a version and that version is not `latest` and the version actually matches the current version they have, then we instead message that the version has been pinned. 
Screen Shot 2023-01-25 at 12 22 16 PM

If the user pins a version and it's not latest and it does not match the version they are on, or they pin it to `latest`, or they don't pin a version, we do the update check 
Screen Shot 2023-01-25 at 12 22 11 PM

yuval-klotho (approved):
Logic is good. For some reason, it took me a bit to actually grok this. Partially I could just be a bit out of it, but I think the var name changes I suggested would have helped a lot. If you disagree, feel free to disregard — my suggestions aren't blocking.

pkg/cli/cli_options.go
ewucc: yep can change for sure

PR #146, opened 2023-01-27
docs: add instructions to install from aur

gamemaker1:
Hi,

I noticed that there was no [AUR](https://wiki.archlinux.org/title/Arch_User_Repository) package for Klotho, and made one for it - https://aur.archlinux.org/packages/klotho. The PKGBUILD can be found [here](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=klotho).

### Standard checks

- **Unit tests**: None to be updated
- **Docs**: Public documentation that contain the installation instructions need to be updated.
- **Backwards compatibility**: No breaking changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,17 +80,28 @@ Klotho ensures that developers/operators are able to select and adapt the underl
 
 To install the latest Klotho release, run the following (see [full installation instructions](https://klo.dev/docs/download-klotho) for additional installation options):
 
-Mac:
+**Mac:**
 
 ```sh
 brew install klothoplatform/tap/klotho
 ```
 
-Linux/WSL2:
+**Linux/WSL2:**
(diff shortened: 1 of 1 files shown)


atorres-klo:
Thanks for the submission gamemaker1, to keep users safe we only publicize binaries and packages we upload, to avoid supply chain attacks. 

Would you please let us take ownership so that we can publish updates on build steps? Otherwise, we'd prefer if the package was renamed klotho-unofficial.

gamemaker1:
Sure, I could transfer ownership of the package to you - could you please share the username of the AUR account I should transfer the package to?

atorres-klo:
Sounds good, will get the account set up and get back to you soon.

PR #153, opened 2023-01-30
topology doesnt have type, so use another key

jhsinger-klotho:

need to solve #111 so using something we can key off of for now

### Standard checks


pkg/infra/pulumi_aws/iac/eks.ts
```diff
@@ -576,7 +576,7 @@ export class Eks {
                     if (edge.source == resource.id && edge.target === execUnitEdgeId) {
                         // We know that this exec unit is exposed and must create the necessary resources
                         needsGatewayLink = true
-                        if (resource.type == 'apigateway') {
+                        if (resource.image.includes('api-gateway')) {
```
gordon-klotho, line 579: Seems kinda hacky - is there no way to populate the `type`? Seems like it's in the struct, just needs to be populated.

PR #155, opened 2023-01-30
making topology use config type and adding alb to topology

### LINKED ISSUE #111: Topology and config type are not in sync

jhsinger-klotho:

closes #111 

topology will get type from config since that is the source of truth. Added alb to the topology diagrams and validated it works.

Some small infra changes to the ALB code as well, will change some integ tests to use ALB so we can properly test it

We now remove Type() from all Cloud Resources so we have a single source of truth. This required me to bundle in the telemetry changes as well requested by ala since we report type up to datadog/mixpanel

### Standard checks

### DIFF
--- a/pkg/cli/helpers.go
+++ b/pkg/cli/helpers.go
@@ -181,7 +181,10 @@ func GetLanguagesUsed(result *core.CompilationResult) []core.ExecutableType {
 
 func GetResourceTypeCount(result *core.CompilationResult, cfg *config.Application) (resourceCounts []string) {
 	for _, res := range result.Resources() {
-		resourceCounts = append(resourceCounts, cfg.GetResourceType(res))
+		resType := cfg.GetResourceType(res)
+		if resType != "" {
+			resourceCounts = append(resourceCounts, resType)
+		}
 	}
 	return
 }
--- a/pkg/cli/helpers.go
+++ b/pkg/cli/helpers.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/klothoplatform/klotho/pkg/config"
 	"github.com/klothoplatform/klotho/pkg/core"
 	execunit "github.com/klothoplatform/klotho/pkg/exec_unit"
 	"github.com/klothoplatform/klotho/pkg/infra/kubernetes"
@@ -167,24 +168,20 @@ func OutputResources(result *core.CompilationResult, outDir string) (resourceCou
 	return
 }
 
-func GetLanguagesUsed(result *core.CompilationResult) map[core.ExecutableType]bool {
-	executableLangs := make(map[core.ExecutableType]bool)
+func GetLanguagesUsed(result *core.CompilationResult) []core.ExecutableType {
+	executableLangs := []core.ExecutableType{}
 	for _, res := range result.Resources() {
 		switch r := res.(type) {
 		case *core.ExecutionUnit:
-			executableLangs[r.Executable.Type] = true
+			executableLangs = append(executableLangs, r.Executable.Type)
 		}
 	}
 	return executableLangs
 }
 
-func GetResourceTypeCount(result *core.CompilationResult) (resourceCounts map[string]map[string]int) {
-	resourceCounts = make(map[string]map[string]int)
+func GetResourceTypeCount(result *core.CompilationResult, cfg *config.Application) (resourceCounts []string) {
 	for _, res := range result.Resources() {
-		if _, ok := resourceCounts[res.Key().Kind]; !ok {
-			resourceCounts[res.Key().Kind] = make(map[string]int)
-		}
-		resourceCounts[res.Key().Kind][res.Type()]++
+		resourceCounts = append(resourceCounts, cfg.GetResourceType(res))
 	}
 	return
 }
--- a/pkg/cli/klothomain.go
+++ b/pkg/cli/klothomain.go
@@ -218,7 +218,7 @@ func (km KlothoMain) run(cmd *cobra.Command, args []string) (err error) {
 	}
 	klothoName := "klotho"
 	if km.VersionQualifier != "" {
-		klothoName += " " + km.VersionQualifier
+		analyticsClient.Properties[km.VersionQualifier] = true
 	}
 
 	// if update is specified do the update in place
@@ -290,7 +290,7 @@ func (km KlothoMain) run(cmd *cobra.Command, args []string) (err error) {
 		"app":      appCfg.AppName,
 	})
 
-	analyticsClient.Info(klothoName + "pre-compile")
+	analyticsClient.Info(klothoName + " pre-compile")
 
 	input, err := input.ReadOSDir(appCfg, cfg.config)
 	if err != nil {
@@ -321,7 +321,7 @@ func (km KlothoMain) run(cmd *cobra.Command, args []string) (err error) {
 		Plugins: plugins.Plugins(),
 	}
 
-	analyticsClient.Info(klothoName + "compiling")
+	analyticsClient.Info(klothoName + " compiling")
 
 	result, err := compiler.Compile(input)
 	if err != nil || hadErrors.Load() {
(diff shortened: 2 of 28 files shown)


pkg/topology/topology.go
```diff
@@ -66,7 +66,7 @@ func (p Plugin) generateIconsAndEdges(resource core.CloudResource, dependencies
 		Title: resource.Key().Name,
 		Image: p.getImagePath(resource),
 		Kind:  resource.Key().Kind,
-		Type:  resource.Type(),
+		Type:  p.Config.GetResourceType(resource),
```
gordon-klotho, line 69: I'm not a fan of this change because it's conceding that `resource.Type()` is incorrect. If we're not going to use it, then it needs to be removed.

pkg/core/topology.go
```diff
@@ -62,7 +62,8 @@ var DiagramEntityToImgPath = TopoMap{
 	{Kind: PubSubKind}:                      "generic/blank/blank.png",
 
 	// Use AWS as the ultimate fallback for the Kind, so don't specify Provider.
-	{Kind: GatewayKind, Provider: ProviderAWS}:                          "aws/network/api-gateway.png",
+	{Kind: GatewayKind, Provider: ProviderAWS, Type: "apigateway"}:      "aws/network/api-gateway.png",
+	{Kind: GatewayKind, Provider: ProviderAWS, Type: "alb"}:             "aws/network/elb-application-load-balancer.png",
```
gordon-klotho, line 66: 👍

gordon-klotho (changes requested):

pkg/topology/topology.go
jhsinger-klotho: resource.Type is incorrect and should be removed. This will get our integ tests passing though which is why i want to get it in

jhsinger-klotho: i can cut a new ticket to remove resource.Type if we want

gordon-klotho: That means that plugins will not be empowered to determine the type on the resources it creates. It was originally intended for cases such as if none was specified, it could decide to use something other than the default - like something more intelligent. One example might be to use an ALB instead of API Gateway for an expose if all connecting exec units were non-lambda.

Something to consider, but not a blocker. Perhaps we can discuss quickly in sync.

jhsinger-klotho: lets discuss in sync, but i dont see it making much sense to store the same type in 2 places because how do we know what is the source of truth. Plugins all have access to the application config so that is what i think should be the source of truth

jhsinger-klotho: fixing the plugins doesnt make sense to me. We could have 100 expose plugins so that means each one has to hardcode -> Get type from config, set resource type. We already have config so why not use it, plugins should be able to set the config type if they want to make the change

pkg/cli/helpers.go
```diff
@@ -167,24 +168,20 @@ func OutputResources(result *core.CompilationResult, outDir string) (resourceCou
...
+func GetLanguagesUsed(result *core.CompilationResult) []core.ExecutableType {
+	executableLangs := []core.ExecutableType{}
 	for _, res := range result.Resources() {
 		switch r := res.(type) {
 		case *core.ExecutionUnit:
-			executableLangs[r.Executable.Type] = true
+			executableLangs = append(executableLangs, r.Executable.Type)
 		}
 	}
 	return executableLangs
```
yuval-klotho, line 179: If we're turning this to an array anyway, is it worth sorting it so that it's stable across runs? (Even if we don't care what the order is, just to have it be the same.)

pkg/cli/helpers.go
```diff
@@ -167,24 +168,20 @@ func OutputResources(result *core.CompilationResult, outDir string) (resourceCou
...
 
-func GetResourceTypeCount(result *core.CompilationResult) (resourceCounts map[string]map[string]int) {
-	resourceCounts = make(map[string]map[string]int)
+func GetResourceTypeCount(result *core.CompilationResult, cfg *config.Application) (resourceCounts []string) {
 	for _, res := range result.Resources() {
-		if _, ok := resourceCounts[res.Key().Kind]; !ok {
-			resourceCounts[res.Key().Kind] = make(map[string]int)
-		}
-		resourceCounts[res.Key().Kind][res.Type()]++
+		resourceCounts = append(resourceCounts, cfg.GetResourceType(res))
```
yuval-klotho, line 184: ditto the sorting question

yuval-klotho (approved):

pkg/cli/helpers.go
```diff
@@ -167,24 +168,20 @@ func OutputResources(result *core.CompilationResult, outDir string) (resourceCou
...
-			executableLangs[r.Executable.Type] = true
+			executableLangs = append(executableLangs, r.Executable.Type)
 		}
 	}
 	return executableLangs
 }
 
-func GetResourceTypeCount(result *core.CompilationResult) (resourceCounts map[string]map[string]int) {
-	resourceCounts = make(map[string]map[string]int)
+func GetResourceTypeCount(result *core.CompilationResult, cfg *config.Application) (resourceCounts []string) {
```
gordon-klotho, line 182: This doesn't return the counts anymore. I'd be fine w/ either adjusting the name to match, or reverting back to returning the counts.
I don't think iteration order for this matters since it's only used internally and not in any outputs.

pkg/cli/klothomain.go
```diff
@@ -218,7 +218,7 @@ func (km KlothoMain) run(cmd *cobra.Command, args []string) (err error) {
 	}
 	klothoName := "klotho"
 	if km.VersionQualifier != "" {
-		klothoName += " " + km.VersionQualifier
+		analyticsClient.Properties[km.VersionQualifier] = true
```
gordon-klotho, line 221: Not sure what this change is for, seems unrelated?

pkg/core/exec_unit.go
```diff
@@ -74,8 +74,6 @@ var (
 	ExecutableTypeGolang = ExecutableType("Golang")
 )
 
-func (unit *ExecutionUnit) Type() string { return unit.ExecType }
```
gordon-klotho, line 77: Should `ExecType` (the field) also be removed?

pkg/topology/topology.go
gordon-klotho: Just noting here - I was convinced during sync that changing the config was the better and proper approach. For posterity: the primary driving factor is in making sure the config as we output to the compiled directory needs to match the resources we create. Having the type on the resource would not accomplish this, and if we have to set it on the config anyway, there's no point having a duplicate somewhere else.

gordon-klotho (approved):
Nothing blocking, but I do hope addressing the comment on the resource count function can get squeezed in.

pkg/cli/helpers.go
jhsinger-klotho: yeah so this was a request from ala because mixpanel doesnt process it. The hope here was that we have a list which contains each instance of a type so we can sum the number of "eks" in that list. If youve got a better way to approach this let me know, just needs to work within mixpanels confinements

pkg/cli/klothomain.go
jhsinger-klotho: our metrics in mixpanel are screwed up because it says "klotho procompile complete", etc. This should make it consistent across versions saying "klotho compile complete" with a pro flag so we know if theyre on oss or pro

pkg/core/exec_unit.go
jhsinger-klotho: good catch, will go remove

pkg/cli/helpers.go
jhsinger-klotho: this is purely for internal metrics, so i dont think its necessary. We just send this up to mixpanel and then should write logic to sum the number of langs used acrodss apps

pkg/cli/helpers.go
jhsinger-klotho: same as above

pkg/cli/helpers.go
gordon-klotho: So something like `{"eks": 2, "apigateway": 1}` doesn't work and we'd want `["eks", "eks", "apigateway"]` instead? In that case, this is fine - could you update the function name to something to reflect its new return (eg `GetResourceTypes`)?

PR #159, opened 2023-01-30
allow any extension 5 chars or fewer

yuval-klotho:
We had originally allowlisted the extensions to prevent any sensitive data, but I think the risk of that happening with a 1-4 letter file extension (the `.` already accounts for 1 of the 5) is very low. This way we don't need to keep adding more extensions as we support various file types.

### Standard checks

- **Unit tests**: no tests
- **Docs**: our analytics isn't currently documented, so no changes applicable

### DIFF
--- a/pkg/logging/fields.go
+++ b/pkg/logging/fields.go
@@ -18,8 +18,8 @@ type fileField struct {
 
 func (field fileField) Sanitize(hasher func(any) string) SanitizedField {
 	extension := "unknown"
-	if ext := filepath.Ext(field.f.Path()); len(ext) <= 5 {
-		extension = ext
+	if _, isFileRef := field.f.(*core.FileRef); !isFileRef {
+		extension = filepath.Ext(field.f.Path())
 	}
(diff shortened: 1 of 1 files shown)


gordon-klotho:
Considering the paths (and file names) are hashed, what kinds of sensitive data are we worried about coming from the extensions?

gordon-klotho:
I just had an idea: what if we use "unknown" for `FileRef`s and the extension as-is for everything else?

yuval-klotho:
> Considering the paths (and file names) are hashed, what kinds of sensitive data are we worried about coming from the extensions?

My worry is that someone might accidentally leave off an extension, and instead put sensitive data after the dot. For example, a readme txt file named `v1.proposal_for_some_top_secret_project`, where the "extension" there is `.proposal_for_some_top_secret_project`.

> I just had an idea: what if we use "unknown" for `FileRef`s and the extension as-is for everything else?

Oh, I like that!

gordon-klotho (approved):

PR #160, opened 2023-01-30
rename fargate to ecs and remove hardcoded nlbs

### LINKED ISSUE #116: Cleanup ECS (Fargates) NLBs
we should be able to use the topology to see if these are needed. The recent ALB support cant support fargate until we do so, so that is also a definition of done (Testing ALB + Fargate ECS)

jhsinger-klotho:
closes #116 and closes https://github.com/klothoplatform/klotho-history/issues/633

Renaming fargate to ECS so it aligns with the service name and we can build off of that.
- To do this we check in the Merge of config for exec units and just change the name while throwing a warning. This does log multiple times but we never have a singular place up front to do this (We should likely write the config class to be easier to interact with)

Removes the hardcoded NLBS (Regression of the NLB showing up in topology, however i think the tradeoff is worth it and correct as we wrote code in plugins just to make that happen, but that code wasnt even used in the IAC)

removes GatewayTargets - IMO A gateway target can be derived from dependencies at this stage and when we move AWS code back to compiler we can expand the actual links

For IAC, switched the ecs code off of awsx for load balancing stuff and making it use our LoadBalancing IaC class so that it resembles eks and other lb features in a standardized way

if anyone has a better place to put the warning, im all ears because this is what the output looks like
```sh
new update is available, please run klotho --update to get the latest version
Execution unit type 'fargate' is now renamed to 'ecs'
Execution unit type 'fargate' is now renamed to 'ecs'
Adding resource exec_unit:main
Found 2 route(s) for middleware 'router'                                                                                                                                           unit: main
Adding resource gateway:app
Adding resource persist_kv:users
Execution unit type 'fargate' is now renamed to 'ecs'
Execution unit type 'fargate' is now renamed to 'ecs'
Execution unit type 'fargate' is now renamed to 'ecs'
Execution unit type 'fargate' is now renamed to 'ecs'
Adding resource topology:ts-ms-ecs
Execution unit type 'fargate' is now renamed to 'ecs'
Execution unit type 'fargate' is now renamed to 'ecs'
Adding resource aws_template_data:ts-ms-ecs
Adding resource infra_as_code:Pulumi (AWS)
```

### Standard checks

- **Unit tests**: Any special considerations? this didnt break any unit tests and actually simplifies our code base
- **Docs**: Do we need to update any docs, internal or public? need to update to ecs instead of fargate
- **Backwards compatibility**: Will this break existing apps? If so, what would be the extra work required to keep them working? yeah, fargate will still work, but will throw a warning about the renaming. Eventually we will want to remove fargate

### DIFF
--- a/pkg/provider/aws/config.go
+++ b/pkg/provider/aws/config.go
@@ -52,6 +52,7 @@ type GatewayType string
 // Enums for the types we allow in the aws provider so that we can reuse the same string within the provider
 const (
 	eks                                = "eks"
+	ecs                                = "ecs"
 	fargate                            = "fargate"
 	lambda                             = "lambda"
 	rds_postgres                       = "rds_postgres"
@@ -73,7 +74,7 @@ var defaultConfig = config.Defaults{
 				"memorySize": 512,
 				"timeout":    180,
 			},
-			fargate: {
+			ecs: {
 				"memory": 512,
 				"cpu":    256,
 			},
@@ -149,7 +150,7 @@ func (a *AWS) GetDefaultConfig() config.Defaults {
 func (a *AWS) GetKindTypeMappings(kind string) ([]string, bool) {
 	switch kind {
 	case core.ExecutionUnitKind:
-		return []string{eks, fargate, lambda}, true
+		return []string{eks, ecs, fargate, lambda}, true
 	case core.GatewayKind:
 		return []string{string(ApiGateway), string(Alb)}, true
 	case core.StaticUnitKind:
--- a/pkg/provider/aws/config.go
+++ b/pkg/provider/aws/config.go
@@ -53,7 +53,6 @@ type GatewayType string
 const (
 	eks                                = "eks"
 	ecs                                = "ecs"
-	fargate                            = "fargate"
 	lambda                             = "lambda"
 	rds_postgres                       = "rds_postgres"
 	s3                                 = "s3"
@@ -150,7 +149,7 @@ func (a *AWS) GetDefaultConfig() config.Defaults {
 func (a *AWS) GetKindTypeMappings(kind string) ([]string, bool) {
 	switch kind {
 	case core.ExecutionUnitKind:
-		return []string{eks, ecs, fargate, lambda}, true
+		return []string{eks, ecs, lambda}, true
 	case core.GatewayKind:
 		return []string{string(ApiGateway), string(Alb)}, true
 	case core.StaticUnitKind:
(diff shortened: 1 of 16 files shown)


gordon-klotho (commented):
I thought we were moving things out of deploylib and into Go - doesn't this do the other way around?

jhsinger-klotho:
> I thought we were moving things out of deploylib and into Go - doesn't this do the other way around?

so a couple things

1. Having fargate and nlb hardcoded in expose plugins i dont think ever was a good idea and all this pr really does is remove that. I agree we eventually want it to live in the provider section but we arent ready for that.
2. The reason we arent ready is because we need to do the better results and dependencies structs task. How are we storing dependencies of the lower level aws resources? We cant put them right back in the same dependencies map without it being over saturated and we really need a dependency tree to understand deployment order.

So this doesnt move us to the goal, youre right, but its really just code removal, theres not many additions within pulumi outside of moving the nlb code into the ecs service function. I do think if we want to move things into go we have to start picking up the ground work because we dont have the ability to support making it smarter yet IMO

pkg/provider/aws/config.go
```diff
@@ -52,6 +52,7 @@ type GatewayType string
 // Enums for the types we allow in the aws provider so that we can reuse the same string within the provider
 const (
 	eks                                = "eks"
+	ecs                                = "ecs"
 	fargate                            = "fargate"
```
gordon-klotho, line 56: Do we still need `fargate`?

gordon-klotho (approved):
👍 your justification makes sense. Got to take 1 step back before we go 2 forwards, and the LB is the only place that the cloud resources don't represent the "high level" topology.

pkg/provider/aws/config.go
jhsinger-klotho: yeah if we dont then we would error out right now for anyone who has it in the config, i was going to leave it and cut another ticket to fully remove it later on so that we give them a little to switch the name for themselves. If we dont think thats important i should remove all notions of the naming.

Also im changing the code so that we only warn once about the naming change so need to do one more update

gordon-klotho: I thought it gets re-written in https://github.com/klothoplatform/klotho/pull/160/files#diff-a3d824da3c42420cd5cbb0a4a2c0e7b5bfddd819652788a0596d195dc6e31fa5R153-R156 so it should never be `"fargate"`. But, if it's still needed, then no worries - keep it.

jhsinger-klotho: ah right, good catch, let me remove

jhsinger-klotho:
so pulumi also cant switch load balancers when they are attached to the service, which is annoying, so i switched back to awsx so that we dont break anyone with an existing setup and can save all the breaking changes for later. Ran an upgrade test which would have failed

PR #166, opened 2023-01-31
cleanups to "go generate"

yuval-klotho:
I'm putting this out on spec; if people don't like it, I'm happy to withdraw it. My motivation is that I keep editing the wrong file and/or forgetting which I _should_ edit, and I'm happier if my IDE reminds me which one is generated and which one is committed.

Changes in this PR:

- don't check in the generated files
  - instead, add them to a local gitignore
  - delete the ones that were there before
- use `npx tsc` rather than `tsc`. This means people don't need tsc globally installed, plus it ensures we all use the same version.
- only invoke `tsc` once, to save some time (on my machine, it brought `go generate ./...` down from ~10 seconds to ~3).

This does have one tradeoff, which is that you need to run `go generate ./...` at least once before you build. In particular, this means that I had to add that command to the integ test runner. That's arguably a bit less convenient and a bit more safe (in terms of not compiling old code that someone forgot to generate), so whether it's good or bad in net is up for interpretation.

### Standard checks

- **Unit tests**: no issues, other than the ones mentioned above
- **Docs**: no issues
- **Backwards compatibility**: shouldn't be any issues:
  - I did run `md5sum *` in `klotho/pkg/lang/[python|javascript]/aws_runtime` (on main, vs on this branch after `go generate`), and confirmed that they're the same other than the bash scripts
  - this _may_ delete people's local files when they pull, I'm not sure. If so, running `go generate ./...` will fix it


pkg/lang/javascript/aws_runtime/.gitignore
```diff
@@ -1 +1,12 @@
...
+dispatcher_lambda.js.tmpl
+emitter.js.tmpl
+fs.js.tmpl
+keyvalue.js.tmpl
+orm.js.tmpl
+proxy_eks.js.tmpl
+proxy_fargate.js.tmpl
+redis_cluster.js.tmpl
+redis_node.js.tmpl
+secret.js.tmpl
```
jhsinger-klotho, line 12: why ignore checking these in? we dont have .js files for all of them so we need some of these checked in

gordon-klotho (commented):
Nice, been waiting for this. Looks like we need to add the generate to the govulncheck action though

pkg/lang/javascript/aws_runtime/.gitignore
yuval-klotho: Which ones don't we have `.ts` files for? I actually even create the changes to this file manually; they're all from the bash script's changes to add the generated files to this ignore.

jhsinger-klotho: oh i guess this list doesnt include the ones which we dont have ts files for

jhsinger-klotho (approved):

pkg/lang/javascript/aws_runtime/compile_template.sh
```diff
@@ -6,13 +6,18 @@ if [ 'Darwin' = "$(uname -s)" ]; then
...
-
     mv _${var}.js ${var}.js.tmpl
-    ksed 's://TMPL ::g' ${var}.js.tmpl 
+    ksed 's://TMPL ::g' ${var}.js.tmpl
     echo "generated ${var}.js.tmpl"
+    echo ${var}.js.tmpl >> .gitignore
 done
+
+sort -u .gitignore > gitignore-tmp
+mv gitignore-tmp .gitignore
```
gordon-klotho, line 23: Clever, I like this a lot.

gordon-klotho (approved):

PR #167, opened 2023-01-31
conditionally create load balancers for ecs

jhsinger-klotho:

now that topology has the type of expose, we can conditionally create these nlbs and support albs as well

### Standard checks

### DIFF
--- a/pkg/infra/pulumi_aws/deploylib.ts
+++ b/pkg/infra/pulumi_aws/deploylib.ts
@@ -1443,7 +1443,7 @@ export class CloudCCLib {
                 ...baseArgs,
                 image: image,
                 portMappings:
-                    nlb.listeners != undefined
+                    nlb != undefined
                         ? nlb.listeners
                         : [
                               {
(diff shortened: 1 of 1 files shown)


pkg/infra/pulumi_aws/deploylib.ts
```diff
@@ -1365,30 +1365,44 @@ export class CloudCCLib {
...
-
-        const nlb = new awsx.lb.NetworkLoadBalancer(`${execUnitName}-nlb`, {
-            external: false,
-            vpc: this.klothoVPC,
-            subnets: this.privateSubnetIds,
+        let needsLoadBalancer = false
+        this.topology.topologyIconData.forEach((resource) => {
+            if (resource.kind === Resource.gateway) {
+                this.topology.topologyEdgeData.forEach((edge) => {
+                    if (edge.source == resource.id && edge.target === `${execUnitName}_exec_unit`) {
```
yuval-klotho, line 1372: nit, unless this is intentional (in which case I'd ask for a comment):

```suggestion
                    if (edge.source === resource.id && edge.target === `${execUnitName}_exec_unit`) {
```

pkg/infra/pulumi_aws/deploylib.ts
```diff
@@ -1429,7 +1443,7 @@ export class CloudCCLib {
                 ...baseArgs,
                 image: image,
                 portMappings:
-                    nlb.listeners != undefined
+                    nlb != undefined
                         ? nlb.listeners
```
yuval-klotho, line 1447: minor: If you wanted, I think you could just do this as `nlb?.listeners ?? [ { containerPort ... } ]`

yuval-klotho (approved):

DavidSeptimus-Klotho (approved):

PR #178, opened 2023-02-02
allow dockerfiles to be anywhere

### LINKED ISSUE #74: Allow dockerfiles not at the root of klotho dir
If someone wants to nest dockerfiles to fit their project structure, we should allow it. This is for dockerfile overrides though and not our autogenned dockerfiles. Our autogenned should still be put at root path since thats where we understand the context.

jhsinger-klotho:
• Does it relate to or fix any issue? closes #74 

we currently require dockerfiles to be at the klotho root path. With this change if a custom dockerfile is provided we will obey the location it is provided at in regards to the root path of the klotho invocation.

In order to make this work we have to add any annotated file as a sourcefile to prevent it from getting pruned https://github.com/klothoplatform/klotho-pro/pull/50 is the pro request for the same.

to standardize this logic in pulumi, i created a createImage function. We then store the exec unit images in a map to be retrieved by the other functions so we dont have to recreate this in 4 places and muddy the params to those other functions

### Standard checks

- **Docs**: Do we need to update any docs, internal or public? will remove the note in docs when this goes in

### DIFF
--- a/pkg/infra/pulumi_aws/deploylib.ts
+++ b/pkg/infra/pulumi_aws/deploylib.ts
@@ -348,12 +348,15 @@ export class CloudCCLib {
         }
     }
 
-    createDockerAppRunner(execUnitName, envVars: any) {
+    createImage(execUnitName: string, dockerfilePath: string) {
         const image = this.sharedRepo.buildAndPushImage({
-            context: `./${execUnitName}`,
+            context: `./${execUnitName}/${dockerfilePath}`,
             extraOptions: ['--platform', 'linux/amd64', '--quiet'],
         })
+        this.execUnitToImage.set(execUnitName, image)
+    }
 
+    createDockerAppRunner(execUnitName, envVars: any) {
         const instanceRole = this.createRoleForName(execUnitName)
 
         this.addPolicyStatementForName(execUnitName, {
@@ -458,7 +461,7 @@ export class CloudCCLib {
                         port: isProxied ? '3001' : '3000',
                         runtimeEnvironmentVariables: additionalEnvVars,
                     },
-                    imageIdentifier: image,
+                    imageIdentifier: this.execUnitToImage.get(execUnitName)!,
                     imageRepositoryType: 'ECR',
                 },
             },
@@ -484,10 +487,7 @@ export class CloudCCLib {
         network_placement: 'private' | 'public',
         env_vars?: any[]
     ) {
-        const image = this.sharedRepo.buildAndPushImage({
-            context: `./${execUnitName}`,
-            extraOptions: ['--platform', 'linux/amd64', '--quiet'],
-        })
+        const image = this.execUnitToImage.get(execUnitName)!
 
         const subnetIds =
             network_placement === 'public' ? this.publicSubnetIds : this.privateSubnetIds
@@ -1375,14 +1375,6 @@ export class CloudCCLib {
         )}-eks-cluster`
         const providedClustername = kloConfig.get<string>('eks-cluster')
         const existingCluster = undefined
-        for (const execUnit of execUnits) {
-            const image: pulumi.Output<String> = this.sharedRepo.buildAndPushImage({
-                context: `./${execUnit.name}`,
-                extraOptions: ['--platform', 'linux/amd64', '--quiet'],
-            })
-            execUnit['image'] = image
-            this.execUnitToImage.set(execUnit.name, image)
-        }
         if (this.eks == undefined) {
             if (providedClustername != undefined) {
                 const existingCluster: aws.eks.GetClusterResult = await aws.eks.getCluster({
@@ -1413,10 +1405,7 @@ export class CloudCCLib {
             this.createEcsCluster()
         }
 
-        const image = this.sharedRepo.buildAndPushImage({
-            context: `./${execUnitName}`,
-            extraOptions: ['--platform', 'linux/amd64', '--quiet'],
-        })
+        const image = this.execUnitToImage.get(execUnitName)!
 
         const role = this.createRoleForName(execUnitName)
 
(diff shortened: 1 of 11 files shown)


pkg/lang/javascript/parser.go
```diff
@@ -12,8 +12,10 @@ import (
 
 var multilineCommentMarginRegexp = regexp.MustCompile(`(?m)^\s*[*]*[ \t]*`) // we need to use [ \t] instead of \s, because \s includes newlines in (?m) mode.
 
+const js = core.LanguageId("javascript")
```
gordon-klotho, line 15: I moved this into the struct a while back and just use `Language.ID` (since it's exported and IMO clearer what it is).

pkg/lang/golang/aws_runtime/aws.go
```diff
@@ -47,10 +48,12 @@ func (r *AwsRuntime) AddExecRuntimeFiles(unit *core.ExecutionUnit, result *core.
 		ExecUnitName:   unit.Name,
 	}
 
-	err := golang.AddRuntimeFile(unit, templateData, "Dockerfile", DockerFile)
-	if err != nil {
-		return err
+	if runtime.ShouldOverrideDockerfile(unit) {
+		err := golang.AddRuntimeFile(unit, templateData, "Dockerfile", DockerFile)
```
gordon-klotho, line 52: Should this also set the `unit.DockerfilePath` to "Dockerfile", or is that not necessary

pkg/infra/pulumi_aws/deploylib.ts
```diff
@@ -341,12 +341,15 @@ export class CloudCCLib {
         }
     }
 
-    createDockerAppRunner(execUnitName, envVars: any) {
+    createImage(execUnitName: string, dockerfilePath: string) {
         const image = this.sharedRepo.buildAndPushImage({
-            context: `./${execUnitName}`,
+            context: `./${execUnitName}/${dockerfilePath}`,
```
gordon-klotho, line 346: I think this needs to be
```js
{
    context: `./${execUnitName}/${path.dirname(dockerfilePath)}`,
    dockerfile: `./${execUnitName}/${dockerfilePath}`,
}
```
Though I'm not sure if using that context is correct versus using `./${execUnitName}`.

gordon-klotho (changes requested):
Only blocking is the `createImage` comment.

pkg/infra/pulumi_aws/deploylib.ts
jhsinger-klotho: so i printed out what my code has for the context and its `    ./main/src` which would be correct

jhsinger-klotho: we can leave context at the root of the exec unit if thats what you mean

jhsinger-klotho: the path is already filepath.Dir in go so it doesnt include the file name

pkg/lang/golang/aws_runtime/aws.go
jhsinger-klotho: the ShouldOverrideDockerfile method is what sets that since its not language specific

gordon-klotho: It looks like it only sets it if `ShouldOverrideDockerfile` is false.

pkg/infra/pulumi_aws/deploylib.ts
gordon-klotho: Hm, isn't that wrong then? What if your dockerfile was `build/myservice.Dockerfile`? At the least the field name doesn't match its content.

jhsinger-klotho: i might be a bit confused by what you mean. Doesnt it assume that there is 1 and only 1 dockerfile within the context and it will use that? if it was `build/myservice.Dockerfile` then we would point to `{execunit}/build/myservice.Dockerfile` which i would think is correct

pkg/lang/golang/aws_runtime/aws.go
jhsinger-klotho: right because it will be "" if we override the dockerfile since the dockerfile will be at the root of the execution units directory

pkg/infra/pulumi_aws/deploylib.ts
gordon-klotho: The `file` parameter defaults to `Dockerfile`, but I don't believe it's relative to the context but rather the current working directory from when it's run. I don't believe it does any sort of searching. Additionally, by sending the context of `{execunit}/build` that would mean it couldn't include any source files in `{execunit}/src` (or similar, outside of the build directory).

jhsinger-klotho: theres no way we can understand the context without parsing the Dockerfile, so are you suggesting just leaving it at the root exec unit so we can encompass everything? If so then thats fine with me and similarly is the question, does the --file parameter only look for files named Dockerfile or would it find any Dockerfile?

gordon-klotho: > theres no way we can understand the context without parsing the Dockerfile

Not only that, but even that might not be able to tell you enough (though in many circumstances we could make a guess based on any `ADD`/`COPY` statements). This would have to be a directive to support all the different edge cases.

> are you suggesting just leaving it at the root exec unit so we can encompass everything?

Yeah, for now until we get some more intelligence built in.

> does the --file parameter only look for files named Dockerfile or would it find any Dockerfile?

I'm relatively certain it only finds the file named `Dockerfile`. Most non-trivial build processes I've seen use explicit `-f path/to/dockerfile` for this reason.

jhsinger-klotho: ok i can go make these updates then, will have a new revision out today

jhsinger-klotho:
the branch got a bit messed up, so i had to rebase to fix conflicts with main so the changes from the last commit should just be to address the comments, but hard to see now

gordon-klotho (approved):
Not sure what the `js` change in parser.go is used for or why the change, but the important bits (things related to Dockerfile) are good.

PR #179, opened 2023-02-02
remove client-side login retry

yuval-klotho:
I'm also moving the Login error handler to a callback, just as an organizational tool. We want to send the analytics for any error manually (since the client hasn't been set up yet), but the Login method shouldn't have to know about that intricacy. klothomain.go does, so that's where we should put that logic

### Standard checks

- **Unit tests**: not tested
- **Docs**: auth isn't documented yet (this is targeting a feature branch)

### DIFF
--- a/pkg/cli/klothomain.go
+++ b/pkg/cli/klothomain.go
@@ -180,9 +180,21 @@ func (km KlothoMain) run(cmd *cobra.Command, args []string) (err error) {
 		zap.S().Warnf("failed to create .klotho directory: %v", err)
 	}
 
+	analyticsClientProperties := map[string]interface{}{
+		"version": km.Version,
+		"strict":  cfg.strict,
+		"edition": km.DefaultUpdateStream,
+	}
+
 	// Set up user if login is specified
 	if cfg.login {
-		err := auth.Login()
+		err := auth.Login(func(err error) {
+			// We don't have the analytics client set up to the logger yet, so the warn message won't send any
+			//analytics. Manually create a client and send the tracking.
+			zap.L().Warn(`Couldn't log in. You may be able to continue using klotho without logging in for now, but this may break in the future. Please contact us if this continues.'`)
+			client := &analytics.Client{Properties: analyticsClientProperties}
+			client.Warn("login failed")
+		})
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


pkg/cli/klothomain.go
```diff
@@ -180,9 +180,21 @@ func (km KlothoMain) run(cmd *cobra.Command, args []string) (err error) {
...
+		"version": km.Version,
+		"strict":  cfg.strict,
+		"edition": km.DefaultUpdateStream,
+	}
+
 	// Set up user if login is specified
 	if cfg.login {
-		err := auth.Login()
+		err := auth.Login(func(err error) {
+			// We don't have the analytics client set up to the logger yet, so the warn message won't send any
```
gordon-klotho, line 192: Is there a reason we couldn't set up the analytics client earlier? Doesn't seem like it needs anything setup in this block.

gordon-klotho:
Considering this is still a network call, it could still fail in a way that wasn't caught by the server's retry. Should we keep this in?

pkg/cli/klothomain.go
yuval-klotho: `analytics.NewClient` looks for login creds (either authenticated or local), and errors out if they're not there — so you wouldn't be able to log in until you've logged in, which means our only user would be Yossarian.

This flow does feel pretty janky though, I agree. I think a nicer thing would be to first construct the client without login info, then do any possible login, then attach the login info. So:

1. Does that revised workflow make sense, and if so
2. Should I do it in this PR (which is into the `auth` branch, which will eventually need full review to land into `main`) or as a separate, followup cleanup task?

gordon-klotho: > I think a nicer thing would be to first construct the client without login info, then do any possible login, then attach the login info.

👍 
This branch or a different PR doesn't matter to me.

yuval-klotho:
> Considering this is still a network call, it could still fail in a way that wasn't caught by the server's retry. Should we keep this in?

From the godocs, it looks like the retry loop wouldn't catch 500s and such — it only catches client-side policies (like CheckRedirect), or "failure to speak http (such as a network connectivity problem)". The former wouldn't benefit from a retry. The latter are a bit tricky: if it's a spotty connection, the a retry could help; but if your network is just down (which is probably more likely?), then a retry hang is likely to be annoying and unhelpful.

Idk, I could go either way on it.

pkg/cli/klothomain.go
yuval-klotho: Filed as #180

jhsinger-klotho (approved):

PR #182, opened 2023-02-03
fix some Close() calls

yuval-klotho:
In some cases, were creating the deferred close after the method would have already returned. Also, we were dropping some errors on the ground; this will now handle them.

### Standard checks

- **Unit tests**: not tested
- **Docs**: n/a

### DIFF
--- a/pkg/closenicely/closeutil.go
+++ b/pkg/closenicely/closeutil.go
@@ -1,0 +1,13 @@
+package closenicely
+
+import "go.uber.org/zap"
+
+type CloserWithError interface {
+	Close() error
+}
+
+func OrDebug(closer CloserWithError) {
+	if err := closer.Close(); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/closenicely/closeutil.go
```diff
@@ -0,0 +1,13 @@
...
+
+import "go.uber.org/zap"
+
+type CloserWithError interface {
+	Close() error
+}
+
+func OrDebug(closer CloserWithError) {
+	if err := closer.Close(); err != nil {
+		zap.S().Debug("Failed to close resource: %s", err)
```
gordon-klotho, line 11: Maybe pass the error as a field so the Logger can choose how to format (eg, with stacktrace)?

gordon-klotho (approved):
Nice, I like it

PR #184, opened 2023-02-03
allow env var for id token

jhsinger-klotho:

Allows specification of KLOTHO_ID_TOKEN for ci/cd use case

only deletes creds file if it exists on logout

### Standard checks

### DIFF
--- a/pkg/auth/credentials.go
+++ b/pkg/auth/credentials.go
@@ -30,6 +30,14 @@ func WriteIDToken(token string) error {
 }
 
 func GetIDToken() (*Credentials, error) {
+
+	idToken := os.Getenv("KLOTHO_ID_TOKEN")
+	if idToken != "" {
+		return &Credentials{
+			IdToken: idToken,
+		}, nil
+	}
+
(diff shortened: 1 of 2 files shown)


pkg/auth/credentials.go
```diff
@@ -30,6 +30,14 @@ func WriteIDToken(token string) error {
...
+
+	idToken := os.Getenv("KLOTHO_ID_TOKEN")
+	if idToken != "" {
+		return &Credentials{
+			IdToken: idToken,
+		}, nil
+	}
+
 	configPath, err := cli_config.KlothoConfigPath("credentials.json")
 	result := Credentials{}
```
gordon-klotho, line 42:
```suggestion
	result := Credentials{IdToken: os.Getenv("KLOTHO_ID_TOKEN")}
	if result.IdToken != "" {
		return &result, nil
	}

	configPath, err := cli_config.KlothoConfigPath("credentials.json")
```
nit: by initializing credentials earlier, this gets slightly shorter

gordon-klotho (approved):

PR #185, opened 2023-02-03
pass lb plugin back to eks plugin

jhsinger-klotho:

fixes bug in integ tests, the lb plugin was not getting passed back correctly

### Standard checks

### DIFF
--- a/.github/workflows/run-integ-tests.yaml
+++ b/.github/workflows/run-integ-tests.yaml
@@ -139,6 +139,13 @@ jobs:
       matrix:
         app_to_test: ${{ fromJson(needs.list-apps.outputs.to_test) }}
         mode: [fresh, upgrade]
+        exclude:      
+          -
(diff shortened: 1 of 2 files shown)


.github/workflows/run-integ-tests.yaml
```diff
@@ -139,6 +139,13 @@ jobs:
       matrix:
         app_to_test: ${{ fromJson(needs.list-apps.outputs.to_test) }}
         mode: [fresh, upgrade]
+        exclude:
```
yuval-klotho, line 142: Can you file issues for the ignored tests (or just one, if they're the same problem) and add a reference to it as comments to the ignore?

jhsinger-klotho: yup will do

yuval-klotho (approved):

Issue #188, opened 2023-02-03
sanitize s3 bucket broken


jhsinger-klotho:
https://github.com/klothoplatform/klotho/actions/runs/4081876968/jobs/7035703875 has a failure for the bucket name saying its invalid. Removing now for other reasons as well which will be commented in code

jhsinger-klotho:
Heres the error from the integ test

    aws:s3:Bucket ():
      error: 1 error occurred:
      	* Error creating S3 bucket: InvalidBucketName: The specified bucket is not valid.
      	status code: 400, request id: CTQ279AKVHPQCXDK, host id: QToJlyi21j4Q6sXlWr+hgiwf/PnU/jV3J/pB+/tAVuWtHndkt2GaV4CRDbs6rOy0yyHQyKHHMJc=
  
    pulumi:pulumi:Stack (ts-orm-environment-variable-directive-f-4081876968-ts-orm-environment-variable-directive-f-4081876968):
      error: update failed

DavidSeptimus-Klotho:
The current bucket creation logic with sanitization enabled would output the following valid (truncated) bucket name:

`123456789012-ts-orm-environment-variable-directive-f-40`

That said, this definitely doesn't seem appropriate given we now support multiple buckets per app.

PR #189, opened 2023-02-03
payloads bucket name cant be sanitized

jhsinger-klotho:

cut #188  to get to the root cause, but this was causing failures in the integ tests. Right now we also cant sanitize because it may make our runtime code go out of sync. For that issue i cut klotho-pro #51 so we use environment variables there for less risk.

### Standard checks

### DIFF
--- a/pkg/infra/pulumi_aws/deploylib.ts
+++ b/pkg/infra/pulumi_aws/deploylib.ts
@@ -113,12 +113,8 @@ export class CloudCCLib {
         if (this.createVPC) {
             this.getVpcSgSubnets()
         }
-        const resolvedBucketName = this.account.accountId.apply(
-            (accountId) =>
-                sanitized(
-                    AwsSanitizer.S3.bucket.nameValidation()
-                )`${accountId}${physicalPayloadsBucketName}`
-        )
+        // Right now we sanitize this bucket name in the compiler go code so we do not need to here.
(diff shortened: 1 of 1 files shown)


pkg/infra/pulumi_aws/deploylib.ts
```diff
@@ -113,12 +113,8 @@ export class CloudCCLib {
         if (this.createVPC) {
             this.getVpcSgSubnets()
         }
-        const resolvedBucketName = this.account.accountId.apply(
-            (accountId) =>
-                sanitized(
-                    AwsSanitizer.S3.bucket.nameValidation()
-                )`${accountId}${physicalPayloadsBucketName}`
-        )
+        // Right now we sanitize this bucket name in the compiler go code so we do not need to here.
```
yuval-klotho, line 116:
```suggestion
        // Right now we sanitize this bucket name in the compiler go code so we do not need to here (issue #188)
```

yuval-klotho (approved):

pkg/infra/pulumi_aws/deploylib.ts
jhsinger-klotho: will add that but also the klotho-pro #51 is the true reason as to why we cant sanitize. We cant risk it being out of sync with the go code

PR #193, opened 2023-02-03
replace local login with auth0

### LINKED ISSUE #180: cleanup analytics client creation
See #179:

> This flow does feel pretty janky though, I agree. I think a nicer thing would be to first construct the client without login info, then do any possible login, then attach the login info.

yuval-klotho:
There's a fair amount going on here, sorry.

- split analytics/client.go:NewClient up, such that its email stuff is in a new method, AttachAuthorizations().
- rm most of analytics/user.go, and replace it with a tiny method GetOrCreateAnalyticsFile(). This just upserts the ~/.klotho/analytics.json file. All of the logic around retrieving the user and validating emails and all that is just gone.
- auth/auth.go:
  - Authorize() now returns its claims. We pass these to AttachAuthorizations(), mentioned above.
  - We download our auth0 PEM (and cache it), and use it to verify the auth0 token. Without this, we can't actually get the claims.
- cli/klothomain.go:
  - construct analytics.Client just once, and then later attach the email from the claims (this resolves #180)
  - also move the log hooks up, to follow the client

### Standard checks

- **Unit tests**: um, some of this could definitely be unit tested. Right now it's a lot of manual testing.
- **Docs**: auth isn't documented yet

### DIFF
--- a/pkg/auth/auth.go
+++ b/pkg/auth/auth.go
@@ -165,7 +165,7 @@ type KlothoClaims struct {
 	Email         string `json:"email"`
 	EmailVerified bool   `json:"email_verified"`
 	Name          string `json:"name"`
-	jwt.StandardClaims
+	jwt.RegisteredClaims
 }
 
 func Authorize() (*KlothoClaims, error) {
@@ -192,7 +192,7 @@ func authorize(tokenRefreshed bool) (*KlothoClaims, error) {
 		}
 	} else if !claims.ProEnabled {
 		return nil, fmt.Errorf("user %s is not authorized to use KlothoPro", claims.Email)
-	} else if claims.ExpiresAt < time.Now().Unix() {
+	} else if claims.ExpiresAt.Before(time.Now()) {
 		if tokenRefreshed {
 			return nil, fmt.Errorf("user %s, does not have a valid token", claims.Email)
 		}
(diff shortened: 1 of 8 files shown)


pkg/auth/auth.go
```diff
@@ -237,3 +237,53 @@ func getAuthUrlBase() string {
...
+	if err != nil {
+		return nil, err
+	}
+	bs, err := os.ReadFile(configPath)
+	// Couldn't read it from cache, so (a) try to fetch it from URL and (b) mark down that we should write it on success
+	if err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
+			zap.L().Debug("Couldn't read PEM cache file. Will download it.", zap.Error(err))
+		}
+		pemResp, err := http.Get(`https://klotho.us.auth0.com/pem`)
```
jhsinger-klotho, line 254: this should likely be replaced with the method that can use env var overrides so we can test local and beta endpoints. This will otherwise expect our test auth envs to have the prod public key as the signing also

pkg/auth/auth.go
```diff
@@ -163,7 +165,7 @@ type KlothoClaims struct {
 	Email         string `json:"email"`
 	EmailVerified bool   `json:"email_verified"`
 	Name          string `json:"name"`
-	jwt.StandardClaims
+	jwt.RegisteredClaims
```
yuval-klotho, line 168: note this change (and the tweak to the `claims.ExpiresAt` line below, which this change required)

jhsinger-klotho (approved):

yuval-klotho:
I'm going to hold off on merging this for tonight; hopefully we can get some green integ test runs, and I don't want to rock that boat.

PR #199, opened 2023-02-04
Fixes target group replacement and extra logging

### LINKED ISSUE #196: sanitization for target groups is not working
error: aws:lb/targetGroup:TargetGroup resource 'alb_micro-main' has a problem: only alphanumeric characters and hyphens allowed in "name". Examine values at 'TargetGroup.Name'.

DavidSeptimus-Klotho:
Fixes #196

Fixes the `fix()` function for the broken rule by changing the replacement character from `_` to `-` and hides debug logging behind a flag.

### Standard checks

### DIFF
--- a/pkg/infra/pulumi_aws/iac/sanitization/aws/elb.ts
+++ b/pkg/infra/pulumi_aws/iac/sanitization/aws/elb.ts
@@ -40,7 +40,7 @@ export const targetGroup = {
                 regexpMatch(
                     'The name can contain only alphanumeric characters and hyphens.',
                     /^[a-zA-Z\d-]+$/,
-                    (s) => s.replace(/[^a-zA-Z\d-]/g, '_')
+                    (s) => s.replace(/[^a-zA-Z\d-]/g, '-')
                 ),
                 {
                     description: 'The name must not begin with a hyphen',
--- a/pkg/infra/pulumi_aws/iac/sanitization/sanitizer.ts
+++ b/pkg/infra/pulumi_aws/iac/sanitization/sanitizer.ts
@@ -1,6 +1,9 @@
 import Multimap = require('multimap')
 import * as sha256 from 'simple-sha256'
 
+// Set debug = true to enable additional logging
+let debug = false
+
 // Resource name length offset required to account for IAC-added suffixes
 const reservedCharCount = 8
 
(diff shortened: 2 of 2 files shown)


pkg/infra/pulumi_aws/iac/sanitization/aws/elb.ts
```diff
@@ -56,3 +56,12 @@ export const targetGroup = {
         }
     },
 }
+
+console.log(
+    (
```
yuval-klotho, line 61: Just out of curiosity, what does the extra parenthesis do? (It's not a problem, I'm just wondering if there's something neat going on I don't know :) )

yuval-klotho (approved):

pkg/infra/pulumi_aws/iac/sanitization/aws/elb.ts
DavidSeptimus-Klotho: 🤦 I accidentally committed the line of code I was using to reproduce the issue. That was just  a workaround to be able to invoke the fix function, which TypeScript complains about on its own because it _might be_ undefined.

gordon-klotho: Can use `!` for cases where you know better than the compiler to assert that it isn't undefined.

pkg/infra/pulumi_aws/iac/sanitization/sanitizer.ts
```diff
@@ -1,6 +1,9 @@
 import Multimap = require('multimap')
 import * as sha256 from 'simple-sha256'
 
+// Set debug = true to enable additional logging
+let debug = false
```
gordon-klotho, line 5: nit: Would it be possible to make this a config value?

gordon-klotho (approved):

pkg/infra/pulumi_aws/iac/sanitization/sanitizer.ts
DavidSeptimus-Klotho: as in pulumi config?

yuval-klotho: There's also `pulumi.log.[debug|info|etc]`, though I don't know exactly how they work. Maybe the verbose flag sets it?

PR #200, opened 2023-02-05
change metadata name to force replacement

jhsinger-klotho:
• Does it relate to or fix any issue? fixes upgrade path for eks tests

since sanitiztion modiied target group names and thus arns, we need to trigger a new TGB to be created since you cant replace arns in an existing TGB.

error from integ tests
```
    kubernetes:elbv2.k8s.aws/v1beta1:TargetGroupBinding (main-tgb):
      error: 1 error occurred:
      	* the Kubernetes API server reported that "default/main-tgb" failed to fully initialize or become live: admission webhook "vtargetgroupbinding.elbv2.k8s.aws" denied the request: TargetGroupBinding update may not change these fields: spec.targetGroupARN
  ```

### Standard checks

### DIFF
--- a/pkg/infra/pulumi_aws/iac/k8s/add_ons/alb_controller/index.ts
+++ b/pkg/infra/pulumi_aws/iac/k8s/add_ons/alb_controller/index.ts
@@ -24,7 +24,7 @@ export const createTargetBinding = (
             transformations: [
                 // Make every service private to the cluster, i.e., turn all services into ClusterIP instead of LoadBalancer.
                 (obj: any, opts: pulumi.CustomResourceOptions) => {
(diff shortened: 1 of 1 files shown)


gordon-klotho (approved):
I think there's pulumi-meta things you can do to let it know what things might trigger a replacement (`hash` or something?) but this is easy and makes sense to me.

jhsinger-klotho:
> I think there's pulumi-meta things you can do to let it know what things might trigger a replacement (`hash` or something?) but this is easy and makes sense to me.

yeah i could have done delete before replace, but i didnt want that in there permanently because if the port changes we dont want to force downtime by replacing the TGB

PR #210, opened 2023-02-06
dont use fargate for plugins by default and if so add sleep to helm

### LINKED ISSUE #203: Make ALB Controller use cert manager
This had been an issue earlier, but we believed we worked around it by adding a local.Command for sleeping within pulumi. However we are seeing this pop up again in our integ tests. Sample run https://github.com/klothoplatform/klotho/actions/runs/4097231975/jobs/7065466178

```
error: resource default/main was not successfully created by the Kubernetes API server : Internal error occurred: failed calling webhook "mtargetgroupbinding.elbv2.k8s.aws": failed to call webhook: Post "[https://aws-load-balancer-webhook-service.kube-system.svc:443/mutate-elbv2-k8s-aws-v1beta1-targetgroupbinding?timeout=10s](https://aws-load-balancer-webhook-service.kube-system.svc/mutate-elbv2-k8s-aws-v1beta1-targetgroupbinding?timeout=10s)": x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "aws-load-balancer-controller-ca")
```

jhsinger-klotho:

closes #203 

### Standard checks

### DIFF
--- a/pkg/infra/pulumi_aws/iac/eks.ts
+++ b/pkg/infra/pulumi_aws/iac/eks.ts
@@ -132,7 +132,7 @@ interface EksClusterOptions {
 }
 
 export const DefaultEksClusterOptions: EksClusterOptions = {
-    initializePluginsOnFargate: true,
+    initializePluginsOnFargate: false,
     installPlugins: [
         plugins.VPC_CNI,
         plugins.METRICS_SERVER,
@@ -313,7 +313,7 @@ export class Eks {
                                     timeoutSeconds: 30,
                                 },
                             },
-                            namespace: KUBE_SYSTEM_NAMESPACE,
+                            namespace: EXEC_UNIT_NAMESPACE,
                             version: 'v1.10.0',
                         },
                         { provider: this.provider, deleteBeforeReplace: true }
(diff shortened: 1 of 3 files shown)


pkg/infra/pulumi_aws/iac/eks.ts
```diff
@@ -545,14 +553,40 @@ export class Eks {
         if (chart == undefined) {
             throw new Error('Cannot create ServiceExport without Cloud Map Controller installed')
         }
+
+        let sleep
+        if (this.options.initializePluginsOnFargate) {
+            sleep = new local.Command(
```
yuval-klotho, line 559: minor: it looks like most of the if-else blocks are identical, with just the sleep quantity different. If so, you could simplify this a bit imo to make it clear exactly what's different between the two branches:

```
let sleepAmount = this.options.initializePluginsOnFargate ? 60 : 10;
let sleep = new local.Command(...
    create: `sleep ${sleepAmount}`,
...
```

yuval-klotho (approved):

pkg/infra/pulumi_aws/iac/eks.ts
jhsinger-klotho: ah yeah ill change this, added things in phases so just a bit of brain mush in here like this

PR #214, opened 2023-02-07
simplify loops to find exec units

yuval-klotho:
Rather than looping over `result.Resources()`, use `GetResourcesOfType[ExecutionUnit]`. This is a newer form of the original loop, so this is just some cleanup.

I found these while going through our plugins to catalogue what uses what.

### Standard checks

- **Unit tests**: none changed
- **Docs**: n/a
- **Backwards compatibility**: should be fine


gordon-klotho (approved):
Nice

PR #215, opened 2023-02-07
rename "deps" -> "_" when unused

yuval-klotho:
Just more cleanup.

### Standard checks

- **Unit tests**: n/a
- **Docs**: n/a

### DIFF
--- a/pkg/exec_unit/plugin_assets.go
+++ b/pkg/exec_unit/plugin_assets.go
@@ -17,7 +17,7 @@ type Assets struct{}
 
 func (p Assets) Name() string { return "Assets" }
 
-func (p Assets) Transform(result *core.CompilationResult, deps *core.Dependencies) error {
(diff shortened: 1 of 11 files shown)


gordon-klotho (approved):
I wouldn't (and obviously didn't) use `_` for function parameters, but I'm not opposed to the change.

yuval-klotho:
> I wouldn't (and obviously didn't) use `_` for function parameters, but I'm not opposed to the change.

Intellij warns about them. Should I just turn off that warning?

gordon-klotho:
Is there a way to configure Intellij to use golangci-lint? That way it stays in sync with the git hook / check as well.

In VSCode this is in
image

yuval-klotho:
Will withdraw this, and use [FileWatchers](https://www.jetbrains.com/help/go/settings-tools-file-watchers.html) instead.

PR #216, opened 2023-02-07
initial imports support for golang

jhsinger-klotho:

making some reusable imports logic for go, and using it within the expose plugin where possible

### Standard checks


DavidSeptimus-Klotho:
it looks like the current imports query doesn't support individual imports. e.g:
```go
//detected
import (
    "abc"
    d "efg"
)

// not detected
import "abc"
import d "efg"
```

jhsinger-klotho:
> it looks like the current imports query doesn't support individual imports. e.g:
> 
> ```go
> //detected
> import (
>     "abc"
>     d "efg"
> )
> 
> // not detected
> import "abc"
> import d "efg"
> ```

yeah in expose we have a lot of logic about imports that are incorrect assumptions. I just wanted to at least move the base logic outside of that. Since our query doesnt currently support single imports i think we are relying on the fact that any ide for go will combine them into a single import block. We likely should improve on this, i may cut a completely new ticket for this though if thats ok

DavidSeptimus-Klotho (approved):

Issue #105, opened 2023-01-18
C# execution unit


gordon-klotho:

atorres-klo:
Produce a half-pager on process for future languages and update as we go

PR #220, opened 2023-02-09
Improves project file detection

DavidSeptimus-Klotho:
This PR supports #105 

- Renamed "package" to "project"  in `ReadDir()`to move away from NodeJS terminology
- Replaced fixed project file names with predicate matching + a description for logging
- Implemented .csproj file detection for C#
- Implemented error handling for handling in the event multiple project files of the same language are detected in the same directory

### Standard checks

### DIFF
--- a/pkg/input/readdir.go
+++ b/pkg/input/readdir.go
@@ -44,6 +44,10 @@ type languageFiles struct {
 	projectFileDescription string
 }
 
+func (l languageFiles) isProjectFile(filepath string) bool {
+	return l.projectFilePredicate != nil && l.projectFilePredicate(filepath)
+}
+
 func hasName(expected string) predicate.Predicate[string] {
 	return func(s string) bool {
 		return expected == s
@@ -151,6 +155,7 @@ func ReadDir(fsys fs.FS, cfg config.Application, cfgFilePath string) (*core.Inpu
 	dockerfileLang := &languageFiles{name: DockerFile}
 	allLangs := []*languageFiles{jsLang, pyLang, goLang, yamlLang, csLang}
 	projectDirs := map[languageName][]string{}
+
 	err := fs.WalkDir(fsys, cfg.Path, func(path string, info fs.DirEntry, err error) error {
 		if err != nil {
 			return err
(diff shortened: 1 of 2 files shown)


pkg/input/readdir.go
```diff
@@ -166,16 +203,23 @@ func ReadDir(fsys fs.FS, cfg config.Application, cfgFilePath string) (*core.Inpu
...
+				prjDir, _ := filepath.Split(info.Name())
+				if lang.foundProjectFile == true {
+					err = fmt.Errorf("multiple '%s' files found in directory: %s", lang.projectFileDescription, prjDir)
+					isProjectFile = true
+					break
+				}
+				f, err = addFile(fsys, path, relPath, lang.projectFileOpener)
+				lang.foundProjectFile = true
+				isProjectFile = true
+				projectDirs[lang.name] = append(projectDirs[lang.name], prjDir)
```
jhsinger-klotho, line 218: is this also enabling the ability to have multiple languages in the same klotho compilation? Or what is projectDirs alluding to?

jhsinger-klotho:
just some lint errors as well

gordon-klotho (commented):
Linter caught some things that need to be fixed.

pkg/input/readdir.go
DavidSeptimus-Klotho: not really. it's more just renaming references to "package" or "pkg" to "project" since package is more language-specific.

As far as I'm aware, there's nothing in `readdir` that limits the number of project files/languages we detect. The only logic that has limitations is `openFindUpward`, which is only executed we find 0 project files initially.

DavidSeptimus-Klotho: `projectDirs` refers to directories in which we've detected project files for each supported language. For the moment it's used to enable language-specific file/directory exclusions, and to ensure that compilation fails if two project files exist for the same language in the same directory.

Support for multiple languages in a single compilation is not specifically handled here. The executable plugins don't currently support multiple languages in a single execution unit.

jhsinger-klotho (approved):

gordon-klotho (approved):

Issue #195, opened 2023-02-03
better handling of Errors in console logging


yuval-klotho:
Errors are an odd duck: on the one hand, they (usually? always?) represent pretty low-level information. On the other hand, they can be large, multiline dumps of information (especially if they contain stack traces).

Today (as of the auth branch), we just drop them on console output. We should do something better.

- [ ] only output them when `internalDebug` is on
- [ ] if the field is multi-line:
  - [ ] we should understand that in `printableWidth`, and have that return the maximum width
  - [ ] we should probably print it on its own lines, rather than just inline in the comma-delimited list

As of 752b31b (on the auth branch), the current usages of `zap.Error(e)` and `zap.NamedError` are:

- [`pkg/analytics/user.go`](https://github.com/klothoplatform/klotho/blob/752b31bae8c111ef58555dcb8341325863a4b577/pkg/analytics/user.go#L40): try to write the `analytics.json` file, but if that fails, just log at DEBUG and move on
- `pkg/auth/auth.go`:
  - [1](https://github.com/klothoplatform/klotho/blob/752b31bae8c111ef58555dcb8341325863a4b577/pkg/auth/auth.go): if the PEM file doesn't exist, log at DEBUG that we need to download one and move on
  - [2](https://github.com/klothoplatform/klotho/blob/752b31bae8c111ef58555dcb8341325863a4b577/pkg/auth/auth.go#L285) if we're able to download the PEM file but not save it to disk, then log at DEBUG and move on
- [`pkg/cli/klothomain.go`](https://github.com/klothoplatform/klotho/blob/752b31bae8c111ef58555dcb8341325863a4b577/pkg/cli/klothomain.go#L307): If there is a credentials file, but for whatever reason we can't authorize those credentials, then log a WARN to the effect of "couldn't auth, this is fine for now but may block you in the future" and move on (this is auth's fail open)
- [`pkg/closenicely/closeutil.go`](https://github.com/klothoplatform/klotho/blob/752b31bae8c111ef58555dcb8341325863a4b577/pkg/closenicely/closeutil.go#L14): if a `Close()` errors, log at DEBUG and move on

[...]

Issue #194, opened 2023-02-03
auth should fail-open


yuval-klotho:
Until we have some metrics about our auth flow (including our auth server-side stuff), we should assume it's flaky, and thus should not gate actual usage. It's probably sufficient to do something like:

```
zap.L().Warn(
  "Failed to authorize. For now, klotho will still work without authorization. "+
    "This may change in the future. If this problem persists, please contact us at hello@klo.dev.",
  zap.Error(err))

```

PR #221, opened 2023-02-09
fail-open on auth

yuval-klotho:
If the `credentials.json` file isn't there at all, then require a login.
But if it's there but the login doesn't work for whatever reason, just
issue a warning and move on. Meanwhile, in the login path, handle errors
by writing an empty `credentials.json` and then ignoring the error.

This resolves (once the auth branch is in main) #194.

Relatedly, add a workaround for https://github.com/klothoplatform/klotho/issues/195. It's very easy to hit it with this
new path, so I felt like a quick workaround is in order.

### Standard checks

- **Unit tests**: manually tested
- **Docs**: auth isn't documented yet

### DIFF
--- a/pkg/cli/klothomain.go
+++ b/pkg/cli/klothomain.go
@@ -195,8 +195,12 @@ func (km KlothoMain) run(cmd *cobra.Command, args []string) (err error) {
 
 	// Set up user if login is specified
 	if cfg.login {
-		err := auth.Login(func(err error) {
-			zap.L().Warn(`Couldn't log in. You may be able to continue using klotho without logging in for now, but this may break in the future. Please contact us if this continues.'`)
+		err := auth.Login(func(err error) error {
+			zap.L().Warn(`Couldn't log in. You may be able to continue using klotho without logging in for now, but this may break in the future. Please contact us if this continues.`)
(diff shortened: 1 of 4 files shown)


pkg/cli/klothomain.go
```diff
@@ -298,6 +293,20 @@ func (km KlothoMain) run(cmd *cobra.Command, args []string) (err error) {
 		return nil
 	}
 
+	// Needs to go after the --version and --update checks
+	claims, err := km.Authorizer.Authorize()
+	if claims != nil {
+		analyticsClient.AttachAuthorizations(claims)
```
jhsinger-klotho, line 299: so if we dont have claims because the login failed, what happens here? Will the telemtry just show an email or distinct user id of ""?

jhsinger-klotho (approved):

pkg/cli/klothomain.go
yuval-klotho: It should show up with a `localId` uuid that's saved in the credentials.json. For some reason, I can't find any of this telemetry when I run it locally. I suspect that's a me-problem, but in any case I've added it to the test plan to verify.

PR #222, opened 2023-02-10
Fixes negative log suffix padding

DavidSeptimus-Klotho:
Set padding width to 0 when calculated value is negative, which currently causes a panic. This happens when using a terminal that doesn't properly notify the OS of its width. 

This is only known to impact the debug terminal in Jetbrains products.

### Standard checks

- **Backwards compatibility**: Will this break existing apps? If so, what would be the extra work required to keep them working? No

### DIFF
--- a/pkg/logging/console.go
+++ b/pkg/logging/console.go
@@ -170,6 +170,9 @@ func (enc *ConsoleEncoder) EncodeEntry(ent zapcore.Entry, fieldList []zapcore.Fi
 		} else {
 			padding -= lineLength
 		}
+		if padding < 0 {
+			padding = 0
+		}
 		line.AppendString(strings.Repeat(" ", padding))
 		line.AppendString(fields.String())
 	}
--- a/pkg/logging/terminal.go
+++ b/pkg/logging/terminal.go
@@ -32,7 +32,7 @@ func init() {
 			Height: 60,
 		}
 
-	case err != nil:
+	case err != nil, termSize.Width == 0 && termSize.Height == 0:
 		fmt.Fprintf(os.Stderr, "Could not get terminal size: %v\n", err)
 		termSize = tsize.Size{
 			Width:  80,


yuval-klotho:
> This is only known to impact the debug terminal in Jetbrains products.

It also happens on my iTerm2. It's the actual root cause of #195. I put in a hacky workaround specifically for the `zap.Error(e)` case [on the auth branch](https://github.com/klothoplatform/klotho/commit/752b31bae8c111ef58555dcb8341325863a4b577#diff-e33e3b5cb0850a6c09d44e6f8d8c37c5e03437c09298d12b9df2a9082b9ea241R149-R151), though this is a different approach. Errors are particularly prone to this error because of their stack trace, which produces a very long string (coupled with the fact that `printableWidth` doesn't account for newlines).

I'm not sure what the padding is there for, but I _think_ it's because these strings get reused between loop iterations, so you need to replace anything that was there previously with spaces. If that's true, just setting the padding to 0 won't work (it'll avoid the panic, at the cost of sometimes having previously-written chars poke through).

gordon-klotho:
The padding and printableWidth are to accomplish the right-justification of the fields.
image

It does not play well when the messages contain newlines, for sure.

gordon-klotho (approved):

pkg/logging/console.go
```diff
@@ -170,6 +170,9 @@ func (enc *ConsoleEncoder) EncodeEntry(ent zapcore.Entry, fieldList []zapcore.Fi
 		} else {
 			padding -= lineLength
 		}
+		if padding < 0 {
+			padding = 0
+		}
```
gordon-klotho, line 175: I suppose this can only happen if `size.Width < printableWidth(fields.String())` - which _should_ only happen if the fields are really long. If this is more a property of the terminal, we might need to look into `TermSize` and why it's not returning a reasonable fallback size.

yuval-klotho:
Ah okay — so then the effect of this would be that the long line soft-wraps in the terminal, and then `unit: main` (in your example above) gets appended after it? In that case, might I suggest a default padding of 4 or something, so that it's:

```
WARN: the quick brown fox jumped ov
er the lazy dog    unit:main
```

instead of 

```
WARN: the quick brown fox jumped ov
er the lazy dogunit:main
```

Regardless, it lgtm. Should we:
1. roll back my temporary disabling of printing errors?
2. mark #195 as resolved with this, and maybe file a followup task for handling multiline fields better?

yuval-klotho (approved):

gordon-klotho:
> Ah okay — so then the effect of this would be that the long line soft-wraps in the terminal, and then `unit: main` (in your example above) gets appended after it? In that case, might I suggest a default padding of 4 or something, so that it's:
> 
> ```
> WARN: the quick brown fox jumped ov
> er the lazy dog    unit:main
> ```
> 
> instead of
> 
> ```
> WARN: the quick brown fox jumped ov
> er the lazy dogunit:main
> ```

No, a long line caused by the message is wrapped by appending a newline then right justifying on the empty line:
```
WARN: the quick brown fox jumped ov
er the lazy dog
                          unit:main
```
https://github.com/klothoplatform/klotho/blob/main/pkg/logging/console.go#L168-L169

So this only happens when the initial padding calculation goes negative (the fields are longer than the width).

pkg/logging/console.go
DavidSeptimus-Klotho: I looked into this at one point and it seems like the unix syscall it uses to determine window size returns 0s when debugging in Intellij (it works fine when running with debugging disables).

```go
_, _, errno := unixSyscall(
unix.SYS_IOCTL,
fp.Fd(),
uintptr(unix.TIOCGWINSZ),
uintptr(unsafe.Pointer(&ws)))
```

gordon-klotho: Perhaps another thing we could do then is check for 0s and default to the 80x60 in that case as well.

gordon-klotho: Eg, by updating the switch case to
```go
	case err != nil, termSize.Width == 0 && termSize.Height == 0:
		fmt.Fprintf(os.Stderr, "Could not get terminal size: %v\n", err)
		termSize = tsize.Size{
			Width:  80,
			Height: 60,
		}
```

DavidSeptimus-Klotho: Right. I think some version of this fix had that exact change as well. It just got lost as it's been punted from branch to branch on my local machine.

DavidSeptimus-Klotho: Just added this change to this PR.

gordon-klotho (approved):

yuval-klotho:
Still looking for input on this:

> Should we:
> 1. roll back my temporary disabling of printing errors?
> 2. mark #195 as resolved with this, and maybe file a followup task for handling multiline fields better?

gordon-klotho:
> Should we:
> 
> 1. roll back my temporary disabling of printing errors?
> 2. mark [zap.Error(e) fields can cause console to panic #195](https://github.com/klothoplatform/klotho/issues/195) as resolved with this, and maybe file a followup task for handling multiline fields better?

1. Which change? Is there a commit / code you can point to? If the only reason was the padding, then probably.
2. Error fields should probably be special cased based on the `internalDebug` flag and not print stack traces otherwise. Fields should not contain newlines otherwise (they should be in the message).

yuval-klotho:
> Which change? Is there a commit / code you can point to? If the only reason was the padding, then probably.

Linked in my first comment above, but it's [this one](https://github.com/klothoplatform/klotho/commit/752b31bae8c111ef58555dcb8341325863a4b577#diff-e33e3b5cb0850a6c09d44e6f8d8c37c5e03437c09298d12b9df2a9082b9ea241R149-R151)

> Error fields should probably be special cased based on the internalDebug flag and not print stack traces otherwise. Fields should not contain newlines otherwise (they should be in the message). That is to say, I think the issue is still open and not resolved by this PR.

That makes sense to me. Right now, the workaround is to just drop them from console output — which is consistent with the default path (ie, no internalDebug). Sounds like I should keep that, and then retool 195 to make it be about "show errors nicely when using internalDebug" instead of them causing a panic. Sound good?

gordon-klotho:
> That makes sense to me. Right now, the workaround is to just drop them from console output — which is consistent with the default path (ie, no internalDebug). Sounds like I should keep that, and then retool 195 to make it be about "show errors nicely when using internalDebug" instead of them causing a panic. Sound good?

👍
Where all do we use `WithError` - are they only error logs? I wouldn't want to hide any errors or meaningful messages from the user. This can all go into the resolution of #195 though

yuval-klotho:
> Where all do we use WithError - are they only error logs?

Will comment on 195.

@DavidSeptimus-Klotho just to be explicit, this PR is good to go from my perspective (and has been since rev 1 :) )

PR #225, opened 2023-02-13
check for annotated file as entrypoint

jhsinger-klotho:

We have a slight bug with proxy + pruning combo due to the addition of source files instead of entrypoints. This is because when we dont have the annotated files added as entry points we then prune everything since thats how the dependecy resolver is based.

### Standard checks

### DIFF
--- a/pkg/lang/javascript/plugin_nodejs_executable.go
+++ b/pkg/lang/javascript/plugin_nodejs_executable.go
@@ -52,12 +52,16 @@ func (l NodeJSExecutable) Transform(result *core.CompilationResult, dependencies
 		unit.Executable.Type = core.ExecutableTypeNodeJS
 
 		var err error
-		if len(unit.Executable.Entrypoints) > 0 {
-			err = refreshSourceFiles(unit)
-			if err != nil {
-				return err
+		for _, file := range unit.FilesOfLang(js) {
+			for _, annot := range file.Annotations() {
+				cap := annot.Capability
+				if cap.Name == annotation.ExecutionUnitCapability && cap.ID == unit.Name {
+					if len(unit.Executable.Entrypoints) == 1 {
+						return core.WrapErrf(err, "entrypoint resolution failed for execution unit: %s. Too many annotated files with the same execution unit ID", unit.Name)
+					}
+					unit.AddEntrypoint(file)
+				}
 			}
-			refreshUpstreamEntrypoints(unit)
 		}
 
 		if len(unit.Executable.Entrypoints) == 0 {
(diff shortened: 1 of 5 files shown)


pkg/lang/javascript/plugin_nodejs_executable.go
```diff
@@ -52,30 +52,34 @@ func (l NodeJSExecutable) Transform(result *core.CompilationResult, dependencies
...
-		if len(unit.Executable.Entrypoints) > 0 {
-			err = refreshSourceFiles(unit)
-			if err != nil {
-				return err
+		for _, file := range unit.FilesOfLang(js) {
+			for _, annot := range file.Annotations() {
+				cap := annot.Capability
+				if cap.Name == annotation.ExecutionUnitCapability && cap.ID == unit.Name {
+					if len(unit.Executable.Entrypoints) == 1 {
+						return core.WrapErrf(err, "entrypoint resolution failed for execution unit: %s. Too many annotated files with the same execution unit ID", unit.Name)
```
DavidSeptimus-Klotho, line 60: Multiple files with the same execution unit ID is not an error. Remember that what currently distinguishes an `Entrypoint` from a `SourceFile` is that `Entrypoint`s are used as roots for the file dependency graph.

 (same goes for the other languages)

DavidSeptimus-Klotho (changes requested):

pkg/lang/javascript/plugin_nodejs_executable.go
jhsinger-klotho: whats the use case with multiple lang files with the same exec unit id?

jhsinger-klotho: i can definitely remove this but i wasnt able to think of a reason why we should allow it

DavidSeptimus-Klotho: A case where a file (and its dependencies) are utilized by the execution unit, but are imported/loaded dynamically rather than statically somewhere in the execution unit. This could be directly in the user's code or via a supported framework/application server that executes the user's source code.

jhsinger-klotho: hm, but we arent parsing and reading in those dynamic files are we? im fine removing it for now, but im curious to hear your thoughts and revisit this later

DavidSeptimus-Klotho: The files aren't dynamically generated, just their usage. An example of this might be a Django migration script. It's not going to be imported anywhere in your application, but when you need to execute it with the `django-admin` CLI in a cloud environment, you need a way to ensure that it's included in an execution unit. The annotation is an easy way to accomplish this (though there may be a more optimal/ideal solution).

DavidSeptimus-Klotho (approved):

PR #228, opened 2023-02-14
Don't add method condition for ANY

gordon-klotho:
Allows alb gateway to handle routes with ANY verb.

Test app rules:
image

### Standard checks


jhsinger-klotho (approved):

gordon-klotho:
Ran the following to update the vulnerable package:
```sh
go get -u helm.sh/helm/v3
go mod tidy
```

PR #242, opened 2023-02-16
Config capability and use in go secrets

jhsinger-klotho:

this is a follow on to https://github.com/klothoplatform/klotho/pull/240

This adds the @klotho::config capability

We add this to config, topology, and validation.

This includes the changes to the go secrets plugin to look for klotho::config with the directive of secret = true

### Standard checks


pkg/lang/golang/aws_runtime/Lambda_Dockerfile
```diff
@@ -2,12 +2,10 @@ FROM golang:1.20 as builder
 
 WORKDIR /usr/src/app
 ENV GOOS=linux GOARCH=amd64 CGO_ENABLED=0
-COPY go.mod ./
```
yuval-klotho, line 5: Just to check (since I know you have a bunch of changes in flight), did you meant to commit these changes into this PR?

yuval-klotho (approved):

pkg/lang/golang/aws_runtime/Lambda_Dockerfile
jhsinger-klotho: yeah the dockerfile changes are necessary to get this to work

Issue #244, opened 2023-02-16
document new auth functionality


yuval-klotho:
With the auth0-based approach now hitting main, it's time to update the docs to reflect the new login flow. This should probably be considered a soft blocker for releasing auth0.

Issue #218, opened 2023-02-08
client-level analytics for tracking auth events


yuval-klotho:
We have server-level stats (for errors, etc), but we want client-side analytics. In particular, we want to track the user-visible experience if an auth attempt fails.

Issue #183, opened 2023-02-03
update auth endpoint


yuval-klotho:
`auth.go` currently hard-codes our ALB, because we don't have the DNS entry set up yet. Once we do, we should update the code.

Issue #239, opened 2023-02-15
better handling of analytics properties


yuval-klotho:
We have a mechanism to attach custom information to our log messages, which then gets sanitized and sent to our analytics server. Unfortunately, the only way we can send it today is by including it in the analytic event's name text. This is (I think) supposed to be human readable and succinct, but when we use the sanitization mechanism, it comes through as json.

This has a couple disadvantages:

1. It's hard to see the data in our dashboard; the views are optimized assuming that the event name is human-readable
2. I think (citation needed) it's hard to _use_ in our dashboard; the "event name" field doesn't get auto-parsed

I suspect issue 1 was the motivation behind [`logging.EntryMessageField`][1], which just replaces the analytics message text with the log message. In addition to other problems, this also means that it clobbers any other information that might have been on the sanitized field. I don't think it's a problem in actual usage today, but it's a hidden landmine.

We should consider an alternate way to send these properties, in a way that makes them actually be properties. For example, we could turn each `$key` into a real analytics property, `klotho.$key`. This would require a bit of hackery because of how we currently create and send these messages, but nothing difficult.

[1]: https://github.com/klothoplatform/klotho/blob/acafab93adcee780099b8316dc37a8f06f5da56c/pkg/analytics/fields_listener.go#L49-L53

gordon-klotho:
> I suspect issue 1 was the motivation behind [logging.EntryMessageField](https://github.com/klothoplatform/klotho/blob/acafab93adcee780099b8316dc37a8f06f5da56c/pkg/analytics/fields_listener.go#L49-L53),

`logging.EntryMessageField` was added to get more information on errors encountered - more than just "a compilation failed". I believe that's all it's used for at the moment. It's not the best solution (IMO), but it was a quick and dirty resolution to the problem at the time. We can and should revisit that mechanism.

> In addition to other problems, this also means that it clobbers any other information that might have been on the sanitized field. I don't think it's a problem in actual usage today, but it's a hidden landmine.

💯  absolutely

> We should consider an alternate way to send these properties, in a way that makes them actually be properties.

[...]

PR #245, opened 2023-02-16
use auth0 as our login mechanism

### LINKED ISSUE #114: Auth CLI - Login functionality

### LINKED ISSUE #115: Auth CLI - Claims functionality

### LINKED ISSUE #164: Auth - Keep local login for OSS
make sure within OSS you can still do a local login, with UUID for analytics, but require auth login in Pro

### LINKED ISSUE #174: Verify tokens in the CLI by using public key from auth0
We can verify the signing without having to reach back out to our auth service. This will at least prevent the spoofing of tokens using jwt.WithClaims() or something else to generate a jwt token.

some resources:
https://community.auth0.com/t/where-is-the-auth0-public-key-to-be-used-in-jwt-io-to-verify-the-signature-of-a-rs256-token/8455
https://auth0.com/docs/get-started/tenant-settings/signing-keys
https://auth0.com/docs/secure/tokens/json-web-tokens/validate-json-web-tokens

### LINKED ISSUE #194: auth should fail-open

yuval-klotho:
Big Honkin' Feature Merge™. Sorry. :-(

This rips out all of the current analytics tracking, and replaces it with an auth0-based approach. There's a description of the overall flow at [klothoplatform/klotho-auth-service][1]. If you like a more hands-on approach, you can also [create a manual test runbook instance][2].

There's a lot here, so I'm open to discussing ways we can make it easier to review.

Also curious to see what people think we should do for a merge strategy, since there's a lot of commits and they come from two authors, but they really do want to all land on a unit (or at least, not along the subdivisions that the commits reflect).

This touches several tickets:
- resolves #114
- resolves #115
- resolves #164
- resolves #174
- resolves #194
- makes progress on #218, though I'd like to review these with Ala before calling it done

Outstanding "Auth0 v1" tickets:

| issue                           | description                     | notes                                      |
|---------------------------------|---------------------------------|--------------------------------------------|
| #183                            | update auth endpoint to use DNS | blocked; @AlaShiban needs to configure DNS |
| klothoplatform/klotho-server#33 | support http on auth server     | also blocked on @AlaShiban                  |
| klothoplatform/docs#167         | document new user flow          | will do separately                         |
| ~~klothoplatform/klotho-server#21~~ | automated testing   | won't-fix (see ticket)  |

[1]: https://github.com/klothoplatform/klotho-auth-service#readme
[2]: https://github.com/klothoplatform/klotho-auth-service/issues/new?assignees=&labels=&template=playbook--manual-test.md&title=

### Standard checks

- **Unit tests**: no special considerations
- **Docs**: not documented, but needs to be: #244 
- **Backwards compatibility**: breaks existing login, but not existing deployments


cmd/klotho/main.go
```diff
@@ -1,16 +1,35 @@
...
 	}
+
 	km.Main()
 }
+
+// LocalAuth is an auth.Authorizer that requires login unless its value is true.
+type LocalAuth bool
+
+func (local *LocalAuth) SetUpCliFlags(flags *pflag.FlagSet) {
+	flags.BoolVar((*bool)(local), "local", bool(*local), "If provided, runs Klotho with a local login (that is, not requiring an authenticated login)")
```
yuval-klotho, line 27: style question: does `(*bool)(local), ... bool(*local)` suggest that `LocalAuth` really wants to be a one-field struct, or is it fine as-is?

pkg/auth/auth.go
```diff
@@ -0,0 +1,306 @@
...
+	return authorize(false)
+}
+
+func authorize(tokenRefreshed bool) (*KlothoClaims, error) {
+	creds, claims, err := getClaims()
+	if err != nil {
+		return nil, err
+	}
+
+	if !claims.EmailVerified {
```
yuval-klotho, line 191: As written, this means that if the claim has an unverified email, and _also_ is expired and/or not enabled for pro, then we'll allow it (albeit with a warning). Should we move this `if` to the end of the if-else chain, so that the expiration and pro-check take precedence?

pkg/auth/auth.go
```diff
@@ -0,0 +1,306 @@
...
+		if err != nil {
+			zap.L().Debug("Couldn't write PEM to local cache", zap.Error(err))
+		}
+	}
+	return pub, nil
+}
+
+// EnvVar represents an environment variable, specified by its key name.
+// wrapper around  os.Getenv. This string's value is the env var key. Use GetOr to get its value, or a
+// default if the value isn't set.
```
yuval-klotho, line 294:
```suggestion
// EnvVar represents an environment variable, specified by its key name. This is a
// wrapper around os.Getenv. This string's value is the env var key. Use GetOr to get its value.
```

yuval-klotho (commented):
leaving a bit of a self-review :)

pkg/closenicely/closeutil.go
```diff
@@ -0,0 +1,16 @@
...
+
+func OrDebug(closer io.Closer) {
+	FuncOrDebug(closer.Close)
+}
+
+func FuncOrDebug(closer func() error) {
+	if err := closer(); err != nil {
+		zap.L().Debug("Failed to close resource", zap.Error(err))
+	}
+}
```
gordon-klotho, line 16: Something I've used on side projects which might be nice (adapted to the style of this package):
```go
func OrDebug(o any) {
	var err error
	switch o := o.(type) {
	case io.Closer:
		err = o.Close()
	case func() error:
		err = o()
	default:
		return
	}
	if err != nil {
		zap.L().Debug("Failed to close resource", zap.Error(err))
	}
}
```

pkg/auth/auth.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+	err = CallGetTokenEndpoint(state)
+	if err != nil {
+		return onError(err)
+	}
+	return nil
+}
+
+func CallLoginEndpoint() (string, error) {
+	res, err := http.Get(authUrlBase + "/login")
```
gordon-klotho, line 72: nit: this should use an `*http.Client` and not the default one - though we don't do that a lot of other places so fine for now

pkg/analytics/user.go
```diff
@@ -30,184 +23,49 @@ type Validated struct {
 // located in ~/.klotho/
 var analyticsFile = "analytics.json"
 
-func CreateUser(email string) error {
-
+func GetOrCreateAnalyticsFile() AnalyticsFile {
```
gordon-klotho, line 26: This feels like it should return an `error`

cmd/klotho/main.go
gordon-klotho: 👍 this is idiomatic

gordon-klotho (approved):
Nothing really blocking that we couldn't iterate on. Though we may want to while it's fresh

cmd/klotho/main.go
yuval-klotho: "this" being the code as-is, or a one-field struct?

pkg/logging/fields.go
```diff
@@ -10,12 +10,14 @@ import (
...
 
-var EntryMessageField = "entryMessage"
-
 type fileField struct {
 	f core.File
 }
 
+type LogMessageProducer interface {
+	GetMessage(entry zapcore.Entry) string
+}
```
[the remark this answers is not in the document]
yuval-klotho: It's semi-related to auth, in that this PR adds a second implementation of the "send this string instead of the sanitized json" (`unsanitizedMessage)`. That implementation doesn't send the message, but rather a separate bit of text that's designed for analytics (not for human readability on the console).

From your comment below, it sounds like you actually don't like drawing that distinction, and would prefer that the message on the console play double-duty and also be the message we send to analytics. Is that right? If so, I can just remove this code.

Relates to #239

pkg/cli/klothomain.go
```diff
@@ -178,31 +194,47 @@ func (km KlothoMain) run(cmd *cobra.Command, args []string) (err error) {
...
+	if cfg.logout {
+		err := auth.CallLogoutEndpoint()
+		if err != nil {
+			return err
+		}
+		return nil
+	}
+
+	// Set up analytics
+	if err != nil {
```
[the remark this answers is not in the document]
yuval-klotho: Whoops, this is leftover junk (it's a fragmentary and misplaced dupe of the logic on lines 305-6)

pkg/auth/auth.go
```diff
@@ -0,0 +1,305 @@
...
+		if err != nil {
+			zap.L().Debug("Couldn't write PEM to local cache", zap.Error(err))
+		}
+	}
+	return pub, nil
+}
+
+// EnvVar represents an environment variable, specified by its key name. This is a
+// wrapper around os.Getenv. This string's value is the env var key. Use GetOr to get its value.
+type EnvVar string
```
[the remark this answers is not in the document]
yuval-klotho: I actually had a private TODO to go the other way — move it to an `envvar` package, and then use it in the other places we're already using `os.Getenv` (if they need defaults). Should I do that, and if so, should I do it now?

cmd/klotho/main.go
gordon-klotho: As-is, sorry for the ambiguity

pkg/auth/auth.go
```diff
@@ -0,0 +1,305 @@
...
+		return standardAuthorizer{}
+	}
+	return auth
+}
+
+type standardAuthorizer struct{}
+
+func (s standardAuthorizer) Authorize() (*KlothoClaims, error) {
+	return Authorize()
+}
```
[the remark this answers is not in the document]
yuval-klotho: I'd prefer to keep this as-is for now, since (a) it works, (b) refactoring it later is unlikely to be more work than refactoring it now, and (c) we currently don't have a need for multiple authorizers.

pkg/analytics/user.go
yuval-klotho: What would you suggest we do with that error?

The code as it stands:
1. Tries to get the current analytics.json file
2. If that errors (probably because it doesn't exist, but possibly for other reasons) then:
  1. create a new tracking id (in the form of the `AnalyticsFile` struct); this can never error
  2. try to write that id to analytics.json; this can error, in which case we log a debug and proceed
  3. send that newly created tracking id up to the caller

If we returned an error from this method, I could imagine doing one of two things with it:
1. logging it and proceeding, in which case that's the same as this currently logic
2. erroring out to the user

Am I missing an option? If not, do you think we should do option 1 (in which case imo it's just a stylistic question) or 2 (in which case I disagree, but we should have that discussion)?

pkg/logging/fields.go
```diff
@@ -149,3 +158,18 @@ func (field postLogMessage) MarshalLogObject(enc zapcore.ObjectEncoder) error {
 func PostLogMessageField(msg string) zap.Field {
 	return zap.Inline(postLogMessage{Message: msg})
 }
+
+// SendDirectlyToAnalytics sends the given message to our analytics server, exactly as it is. This will not perform
+// any additional sanitization, so make sure not to send anything sensitive! A good rule is to only invoke this with
+// string literals.
+func SendDirectlyToAnalytics(message string) zap.Field {
```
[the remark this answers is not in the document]
yuval-klotho: (replied to this in the above — keeping the discussion there just for ease)

pkg/logging/fields.go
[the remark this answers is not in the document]
gordon-klotho: > From your comment below, it sounds like you actually don't like drawing that distinction, and would prefer that the message on the console play double-duty and also be the message we send to analytics. Is that right?

The original intent, and my thinking is more like "anything important enough to go to console should go, sanitized, to analytics". There's also stuff that should go to analytics that doesn't go to console.

The second part of my comment was acknowledging that the caller may want to control how the message is sanitized. It's the transformation rather than production aspect that I'm stuck on.

> If so, I can just remove this code.

I don't mean to present as the authority on it. I don't object enough to block anything or require and immediate action.

pkg/auth/auth.go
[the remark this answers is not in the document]
gordon-klotho: Yeah, that's fine too - just that `auth.EnvVar` wouldn't make much sense.

pkg/auth/auth.go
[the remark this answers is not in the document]
gordon-klotho: (a) sure

(b) makes sense

(c) I'd argue that the local implementation is already multiple authorizers. If not, then there's no need for a struct or interface - just call the global methods from klothomain.

pkg/auth/auth.go
yuval-klotho: Talked w/ Jordon offline:

1. This is only a problem while we fail-open
2. So, rather than change the order, I'm going to have the other branches _also_ return both the claims and the err.

pkg/analytics/user.go
gordon-klotho: Regardless of (1) or (2) it doesn't feel like the analytics package should be the "decider" of that behaviour - so I guess it's a stylistic matter.

pkg/logging/fields.go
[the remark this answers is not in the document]
yuval-klotho: For the sake of consistency I'll remove this. I do think that we should separate "what we send to analytics" from "what we show to the user", but that's #239. So I'll go with prior art here, and we can revisit it or not as part of that ticket.

gordon-klotho (approved):

pkg/cli/klothomain.go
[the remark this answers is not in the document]
yuval-klotho: huh, weird — in rebasing, it looks like this line is actually a month old! I think the diff just kinda got confused and thought it was new. Still, it _should_ go away, so it was a fortuitous catch.

yuval-klotho:
Latest push is a rebase plus some fixup squashes. Here's the "diff of diffs" to show how this PR applies to main before and after the rebase:

```diff
$ diff <(git diff origin/auth main) <(git diff auth main)
42c42
< index 3aadacc..65dab23 100644
---
> index cd163da..65dab23 100644
59c59,61
< @@ -32,13 +31,6 @@ require (
---
> @@ -30,13 +29,6 @@ require (
>  	sigs.k8s.io/yaml v1.3.0
>  )
61,62d62
<  replace github.com/smacker/go-tree-sitter => github.com/klothoplatform/go-tree-sitter v0.1.1
<
69a70,71
>  replace github.com/smacker/go-tree-sitter => github.com/klothoplatform/go-tree-sitter v0.1.1
>
71,72d72
<  	github.com/Azure/go-ansiterm v0.0.0-20210617225240-d185dfc1b5a1 // indirect
<  	github.com/BurntSushi/toml v1.2.1 // indirect
90,98d89
< @@ -121,7 +112,7 @@ require (
<  	github.com/shopspring/decimal v1.2.0 // indirect
<  	github.com/sirupsen/logrus v1.9.0 // indirect
<  	github.com/spf13/cast v1.4.1 // indirect
< -	github.com/spf13/pflag v1.0.5
< +	github.com/spf13/pflag v1.0.5 // indirect
<  	github.com/stretchr/objx v0.5.0 // indirect
<  	github.com/xeipuuv/gojsonpointer v0.0.0-20180127040702-4e3ac2762d5f // indirect
<  	github.com/xeipuuv/gojsonreference v0.0.0-20180127040603-bd5ef7bd5415 // indirect
```

So the only changes were in go.mod and go.sum. The rebase just pulled in the latest from main (which makes sense).

PR #246, opened 2023-02-16
Enable API Gateway v2 (HTTP) for expose

gordon-klotho:
This change enables HTTP API Gateways to be created with routes set up.

### Standard checks

- **Unit tests**: IaC is not unit tested
- **Docs**: 🚧 

### DIFF
--- a/pkg/infra/pulumi_aws/iac/api_gateway.ts
+++ b/pkg/infra/pulumi_aws/iac/api_gateway.ts
@@ -1,9 +1,8 @@
 import * as aws from '@pulumi/aws'
 import * as pulumi from '@pulumi/pulumi'
-import { CloudCCLib, ResourceKey, Resource } from '../deploylib'
+import { CloudCCLib } from '../deploylib'
 import * as sha256 from 'simple-sha256'
 import { LoadBalancerPlugin } from './load_balancing'
-import { vpc } from './sanitization/aws/ec2'
 
 export interface Route {
     verb: string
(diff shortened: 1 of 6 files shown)


pkg/infra/pulumi_aws/iac/api_gateway.ts
```diff
@@ -40,105 +42,154 @@ export class ApiGateway {
...
+        switch (execUnit.type) {
+            case 'ecs':
+                const ecsNlb = this.lib.execUnitToNlb.get(r.execUnitName)!
+                return new aws.apigatewayv2.Integration(
+                    integrationName,
+                    {
+                        apiId: api.id,
+                        integrationType: 'HTTP_PROXY',
+                        integrationMethod: 'ANY',
+                        integrationUri: ecsNlb.loadBalancer.arn,
```
jhsinger-klotho, line 113: this will only work if the nlb is an awsx nlb which i think we moved completely off of now

jhsinger-klotho: nevermind

jhsinger-klotho: i thought we migrated away from it but we didnt

jhsinger-klotho (approved):

DavidSeptimus-Klotho (approved):

DavidSeptimus-Klotho:
Do we want to update the signature of `core.NewGateway()` to require an `ApiType` value?

gordon-klotho:
> Do we want to update the signature of `core.NewGateway()` to require an `ApiType` value?

`core.Gateway` wasn't changed, only the config types so no update necessary

Issue #249, opened 2023-02-16
make fs runtime put bucket name in open call


jhsinger-klotho:
due to #227  and https://github.com/klothoplatform/klotho-pro/issues/51 we now create a fs file per bucket. This can be changed so that we modify the open call and inject the bucket name as the first parameter, but i didnt want that pr to grow so large.

This will allow us to inject the file itself with the env var instead of the runtime file. We can then make sure that the payloads bucket is also put as an env var in the dispatcher and proxy files so it is used correctly

Issue #247, opened 2023-02-16
Remove hardcoded bucket name from node emitter template


jhsinger-klotho:
Currently we pass the s3 payload bucket into the template and use that in the file 
```
const payloadBucketPhysicalName = process.env.KLOTHO_S3_PREFIX + '{{.PayloadsBucketName}}';
```

https://github.com/klothoplatform/klotho-pro/issues/51 started the effort of moving this all to Environment variables so that we can dynamically generate these templates based on the iac created. This allows for proper sanitization and the resources to stay in sync

PR #250, opened 2023-02-16
[BACKWARDS INCOMPATIBLE] S3 env var

### LINKED ISSUE #227: refactor fs to create buckets based on ID
This setting is being used within the go implementation so that we can generate new buckets based on each `openBucket` Call. 

We have a new Buckets list within our aws provider template that when added to will generate a new bucket and inject an environment variable to the exec unit so they know which bucket to use. 

This will require runtime changes in python and node to use the env var instead of the hardcoded bucket name

This has been an ask by a user who is developing a multi language application that wants to share resources and this work will need to be done to allow for that

jhsinger-klotho:
• Does it relate to or fix any issue? closes #227  and closes https://github.com/klothoplatform/klotho-pro/issues/51

Changes our logic so that we create a bucket per ID with an FS annotation. We also create a new InternalResource concept so that we can track the resource we use for proxy/dispatcher (payloads bucket)

We use env vars to inject the bucket names into the fs runtime and we have a file per bucket at the moment. Opened #249 as an enhancement so we can limit the number of runtime files in the future.

We also can fully do sanitization at the IAC layer and now our payloads bucket will have the wording InternalKlothoPayloads. 

opened klotho #247 because we cant fully remove the payloads bucket from template config until we take out the hardcoding from pubsub

### Standard checks

Further testing https://github.com/klothoplatform/klotho/actions/runs/4198937619

- **Unit tests**: Any special considerations? changed some of the transform ones where possible
- **Docs**: Do we need to update any docs, internal or public? Im not sure because we never documented that we dont create multiple buckets
- **Backwards compatibility**: Will this break existing apps? If so, what would be the extra work required to keep them working? No. This is none and stated in the task, we will need to send out an email


pkg/lang/javascript/aws_runtime/aws.go
```diff
@@ -97,7 +99,7 @@ func (r *AwsRuntime) TransformPersist(file *core.SourceFile, annot *core.Annotat
 	importModule := ""
 	switch kind {
 	case core.PersistFileKind:
-		importModule = "fs"
+		importModule = "fs_" + annot.Capability.ID
```
yuval-klotho, line 102: Does this need to be sanitized? I forget what (if any) restrictions we enforce on ids.

pkg/lang/javascript/plugin_persist.go
```diff
@@ -238,17 +239,25 @@ func (p *persister) transformSecret(file *core.SourceFile, cap *core.Annotation,
...
-func (p *persister) transformFS(file *core.SourceFile, cap *core.Annotation, fsR *persistResult) (core.CloudResource, error) {
-	if err := file.ReplaceNodeContent(fsR.expression, "fsRuntime.fs"); err != nil {
-		return nil, errors.Wrap(err, "could not reparse FS transformation")
+func (p *persister) transformFS(unit *core.ExecutionUnit, file *core.SourceFile, cap *core.Annotation, fsR *persistResult) (core.CloudResource, string, error) {
+	if err := file.ReplaceNodeContent(fsR.expression, fmt.Sprintf("fs_%sRuntime.fs", cap.Capability.ID)); err != nil {
+		return nil, "", errors.Wrap(err, "could not reparse FS transformation")
 	}
 
+	fsEnvVar := core.EnvironmentVariable{
+		Name:       cap.Capability.ID + "_bucket_name",
```
yuval-klotho, line 248: Ditto on sanitization here.

pkg/lang/javascript/aws_runtime/aws.go
jhsinger-klotho: so this is just for the import itself in the file. I cut a card so that we actually just pass the bucket name to the open call, but that requires a lot of extra work, so for now we are just importing an fs runtime per bucket and thats why we need different import modules

pkg/lang/javascript/plugin_persist.go
jhsinger-klotho: the name is just the env var name, so it doesnt go through sanitization. The actual bucket name will be the value and that is dynamically generated based on the sanitized bucket name

yuval-klotho: Do annotation ids have any restriction, do you know offhand? For example, can they contain spaces? (afaik, an env var can't.)

yuval-klotho (approved):
Sorry, I forgot to change the radio box to "Approve" when I submitted the previous review. Neither of my comments is blocking.

pkg/lang/javascript/plugin_persist.go
jhsinger-klotho: yeah the annotation ids are the same regex as the app name, so i believe that it should be valid env vars.

PR #276, opened 2023-02-22
add fs for lambda proxy or lambda exec time

jhsinger-klotho:

The lambda dispatcher requires fs to be included as well as the lambda proxy target, but ecs and eks do not. So lets conditionally add it to slim down the compiled output if possible.

The pro integ tests caught this error with proxy still inputting the old runtime format so changing it out.

### Standard checks

- **Unit tests**: Any special considerations? integ tests are what check our runtime

### DIFF
--- a/pkg/lang/python/aws_runtime/aws.go
+++ b/pkg/lang/python/aws_runtime/aws.go
@@ -100,6 +100,19 @@ func (r *AwsRuntime) AddExecRuntimeFiles(unit *core.ExecutionUnit, result *core.
 		dockerFile = dockerfileLambda
 		dispatcher = dispatcherLambda
 		requirements = execRequirementsLambda
+
+		python.AddRequirements(unit, fsRequirements)
+		proxyEnvVar := core.EnvironmentVariable{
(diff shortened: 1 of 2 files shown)


pkg/lang/python/aws_runtime/aws.go
```diff
@@ -100,6 +100,19 @@ func (r *AwsRuntime) AddExecRuntimeFiles(unit *core.ExecutionUnit, result *core.
 		dockerFile = dockerfileLambda
 		dispatcher = dispatcherLambda
 		requirements = execRequirementsLambda
+
+		python.AddRequirements(unit, fsRequirements)
```
gordon-klotho, line 104: If only JS could do this 😢

gordon-klotho (approved):

pkg/lang/python/aws_runtime/aws.go
jhsinger-klotho: yeah definitely a place of improvement for us when we get the time to imrpove node

PR #279, opened 2023-02-22
allow for import of rds instances and proxies

jhsinger-klotho:

allows for the import of rds and proxy instances

import looks similar to below in pulumi config

```
  klo:rds:
    sequelizeDB:
      dbInstanceIdentifier: sequelizedbc90e5a2
      proxy: sequelizedb-8bfa7fc
```

modularizing our pulumi code a bit so that we only add our rds instances if necessary.

tested on ts-sequelize

### Standard checks

- **Unit tests**: Any special considerations? just imports so no
- **Docs**: Do we need to update any docs, internal or public? will add once gordons pr is in since he started the imports page
- **Backwards compatibility**: Will this break existing apps? If so, what would be the extra work required to keep them working? yes new feature

### DIFF
--- a/pkg/infra/pulumi_aws/deploylib.ts
+++ b/pkg/infra/pulumi_aws/deploylib.ts
@@ -62,11 +62,6 @@ export interface TopologyEdgeData {
     target: string
 }
 
-export interface RdsImport {
-    dbInstanceIdentifier: string
-    proxy: string
-}
-
 export const kloConfig: pulumi.Config = new pulumi.Config('klo')
 
 export class CloudCCLib {
--- a/pkg/infra/pulumi_aws/deploylib.ts
+++ b/pkg/infra/pulumi_aws/deploylib.ts
@@ -62,7 +62,12 @@ export interface TopologyEdgeData {
     target: string
 }
 
-export const kloConfig = new pulumi.Config('klo')
+export interface RdsImport {
+    dbInstanceIdentifier: string
+    proxy: string
+}
+
+export const kloConfig: pulumi.Config = new pulumi.Config('klo')
 
 export class CloudCCLib {
     secrets = new Map<string, aws.secretsmanager.Secret>()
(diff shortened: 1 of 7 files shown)


pkg/infra/pulumi_aws/iac/rds.ts
```diff
@@ -0,0 +1,215 @@
...
+        const dbName = sanitized(
+            AwsSanitizer.RDS.engine.pg.database.nameValidation()
+        )`${orm.toLowerCase()}`
+        const username = config.require(`${dbName}_username`)
+        const password = config.requireSecret(`${dbName}_password`)
+
+        let rds
+        let proxy
+
+        if (rdsImports != undefined && rdsImports[orm] != undefined) {
```
atorres-klo, line 27: should we also check for VPC here if required?

atorres-klo (commented):
Minor comments, rest LGTM

pkg/infra/pulumi_aws/iac/rds.ts
jhsinger-klotho: we dont have to, if they want a separate vpc for their workloads i think we should allow it, it would just be on them to peer their vpcs.

If we want to just because were saying its the only use case we support today then i guess we could, im torn on if thats a great idea or not though

atorres-klo: Ah that's fair, this is fine for now.

jhsinger-klotho: especially because in the lambda case it could still connect, its mainly just for the ecs and eks case. So i dont think we want that check

jhsinger-klotho: ah nevermind it still wouldnt be able to connect because of the Security Group

pkg/infra/pulumi_aws/deploylib.ts
```diff
@@ -62,7 +62,12 @@ export interface TopologyEdgeData {
     target: string
 }
 
-export const kloConfig = new pulumi.Config('klo')
+export interface RdsImport {
+    dbInstanceIdentifier: string
+    proxy: string
+}
+
+export const kloConfig: pulumi.Config = new pulumi.Config('klo')
```
gordon-klotho, line 70: I'm surprised this needs the type hint

pkg/infra/pulumi_aws/deploylib.ts
```diff
@@ -62,7 +62,12 @@ export interface TopologyEdgeData {
     target: string
 }
 
-export const kloConfig = new pulumi.Config('klo')
+export interface RdsImport {
```
gordon-klotho, line 65: This feels like it belongs in `rds.ts`

pkg/infra/pulumi_aws/deploylib.ts
```diff
@@ -1003,174 +1008,6 @@ export class CloudCCLib {
             )
     }
 
-    public setupRDS(orm: string, args: Partial<aws.rds.InstanceArgs>) {
```
gordon-klotho, line 1006: IMO it would have been nice to make the changes and the move separately so it was easier to tell what changed

gordon-klotho (approved):

pkg/infra/pulumi_aws/deploylib.ts
jhsinger-klotho: yeah it does let me move it

pkg/infra/pulumi_aws/deploylib.ts
jhsinger-klotho: doesnt need it, i was trying to get some other things working with config and must have just left it

PR #281, opened 2023-02-22
send properties separately from event name in logs

### LINKED ISSUE #239: better handling of analytics properties

yuval-klotho:
When we hook into the log entries, don't send the sanitized fields as json in the event name. Instead, just send the log level for the event name (e.g. "WARN"), and send each sanitized field as a property with a dot-delimited key. For example, a file field would send:

    log.file.extension: .json
    log.file.path: sha256:<snip>

This resolves #239.

### Standard checks

- **Unit tests**: not unit tested; see our analytics dashboard for how this looks
- **Docs**: not needed
- **Backwards compatibility**: Analytics property names and formats have changed, so if we're relying on any for reports, then those would break. I don't think we're doing that.

### DIFF
--- a/pkg/analytics/fields_listener.go
+++ b/pkg/analytics/fields_listener.go
@@ -4,6 +4,7 @@ package analytics
 // as the main entry.
 
 import (
+	"fmt"
 	"github.com/klothoplatform/klotho/pkg/logging"
 	"go.uber.org/zap/zapcore"
 )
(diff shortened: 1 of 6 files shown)


jhsinger-klotho (approved):

pkg/analytics/fields_listener.go
```diff
@@ -44,27 +45,37 @@ func (fl *fieldListener) Write(entry zapcore.Entry, fields []zapcore.Field) erro
 	allFields = append(allFields, fl.fields...)
 	allFields = append(allFields, fields...)
 
-	safeLogsStr := logging.SanitizeFields(allFields, fl.client.Hash)
+	message := entry.Level.CapitalString()
+
+	for k, v := range logging.SanitizeFields(allFields, fl.client.Hash) {
+		property := fmt.Sprintf("log.%s", k)
+		fl.client.Properties[property] = v
+		defer func() { fl.client.DeleteProperty(property) }()
```
gordon-klotho, line 53: This isn't threadsafe which I believe that needs to be a requirement for a fieldListener since calls from concurrent threads/goroutines are not serialized.

gordon-klotho (approved):
As discussed, the analytics is already thread unsafe so this isn't regressing that. Fixing the unsafeness is left for a follow-up PR.

PR #282, opened 2023-02-23
Fixes fs.js s3 read/write issues introduced by switching clients

### LINKED ISSUE #280: Regression: CloudFS Reads File Contents as a String
When migrating to v3 of the AWS client libraries, a change was introduced to `fs.js.tmpl`'s `s3_readFile` function that reads file content to a string. This leads to content mangling of binary files such as images.

impacting change:
https://github.com/KlothoPlatform/klotho/commit/2f169fdca7cb3d9a438fe2106ef1c14c24636c3c#diff-debb6bee1ab0418147f045ef7e807a5abbd72e2dbb816a003b5fd17fe92faeb1R115

DavidSeptimus-Klotho:
Resolves #280

- Stripped leading slashes in object paths supplied to fs functions to be more consistent with expected fs behavior (as well as the older s3fs-based implementation).
- Changed s3_readFile() to return a Promise<Buffer> by default and only returns a string if the user specifies an encoding. Additionally, an empty body results in a Promise<Void> rather than an empty string.

### Standard checks

- **Backwards compatibility**: Will this break existing apps? If so, what would be the extra work required to keep them working? Slim chance. Potentially breaks apps built with `v0.6.1` that using Cloud FS for JavaScript that intend to use `/` as an actual s3 object prefix or that expect `fs.readFile()` to always return a string.

### DIFF
--- a/pkg/lang/javascript/aws_runtime/_fs.ts
+++ b/pkg/lang/javascript/aws_runtime/_fs.ts
@@ -17,12 +17,14 @@ const targetRegion = process.env['AWS_TARGET_REGION']
 
 const s3Client = new S3Client({ region: targetRegion })
 
-const streamToString = (stream: Readable): Promise<string> =>
-    new Promise((resolve, reject) => {
-        const chunks: Uint8Array[] = []
+const streamToString = async (stream: Readable, encoding: string) =>
+    (await streamToBuffer(stream)).toString(encoding)
+const streamToBuffer = (stream: Readable) =>
+    new Promise<Buffer>((resolve, reject) => {
+        const chunks = []
         stream.on('data', (chunk) => chunks.push(chunk))
         stream.on('error', reject)
-        stream.on('end', () => resolve(Buffer.concat(chunks).toString('utf8')))
+        stream.on('end', () => resolve(Buffer.concat(chunks)))
     })
 
 async function getCallParameters(paramKey, dispatcherMode) {
@@ -36,7 +38,7 @@ async function getCallParameters(paramKey, dispatcherMode) {
 
         let parameters: any = ''
         if (result.Body) {
-            parameters = await streamToString(result.Body as Readable)
+            parameters = await streamToString(result.Body as Readable, 'utf-8')
             console.log(parameters)
         }
         if (parameters != '') {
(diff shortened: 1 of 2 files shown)


pkg/lang/javascript/aws_runtime/fs.js.tmpl
```diff
@@ -88,15 +88,24 @@ async function s3_writeFile(...args) {
...
+            if (args[1]?.encoding) {
+                return await streamToString(data.Body);
+            }
+            return new Promise((resolve, reject) => {
+                const stream = data.Body;
+                const chunks = [];
+                stream.on('data', chunk => chunks.push(chunk))
+                stream.once('end', () => resolve(Buffer.concat(chunks)))
+                stream.once('error', reject)
+            });
```
jhsinger-klotho, line 106: nit: make this reusable? Looks similar to steamToString as well but maybe nothing to pull out between the 2

jhsinger-klotho (approved):

pkg/lang/javascript/aws_runtime/fs.js.tmpl
DavidSeptimus-Klotho: Simple enough. I hadn't even looked at the content of `streamToString`.

pkg/lang/javascript/aws_runtime/_fs.ts
```diff
@@ -123,7 +127,7 @@ async function s3_readFile(...args) {
 async function s3_readdir(path) {
     const bucketParams: ListObjectsCommandInput = {
         Bucket: bucketName,
-        Prefix: `${path}`,
+        Prefix: stripLeadingSlashes(`${path}`),
```
gordon-klotho, line 130: Did this really used to be the behaviour? If so, we broke backwards compat, then we'll be breaking backwards compat again to return it.

gordon-klotho (approved):
All good except the backwards compat question

pkg/lang/javascript/aws_runtime/_fs.ts
DavidSeptimus-Klotho: The actual behavior was actually quite a bit more complex as seen here:
https://github.com/hasnat/s3fs/blob/master/lib/utils.js

The key bit is that all the keys were normalized across host operating systems to match a specific format. Removing that from our FS breaks the expectation that you can use the Cloud FS the same as your local FS.

PR #283, opened 2023-02-23
make analytics.Client more thread-safe

yuval-klotho:
It's still not fully thread-safe, since we depend on being able to
append properties to it in klothomain; but for individual sends, it is.

We do this by separating out the client from the payload. Each time we
send a new message, we create a new payload, populate it with default
stuff from the client, modify that paylod (which is thread-local) and
send it.

This PR has two commits (at least for rev1):

- ff5d15bc7937c837be9ba997e25dbad62d95bc30 adds some simple unit tests for the analytics client, but doesn't affect prod at all
- 9ef7fd0ce692546d7200d2c23b94a07ce76e1d64 is the main part of this PR
    - note in particular the change to [`client_test.go` line 91][1]: when we send analytics directly (not via the logger hook), we now send `"status": "info"` where we didn't before

[1]: https://github.com/klothoplatform/klotho/commit/9ef7fd0ce692546d7200d2c23b94a07ce76e1d64#diff-9131bc6b505c18925e929864b3ffc81ae9840d00419019f63daeb5c3410751b8R91

Followup to #281.

### Standard checks

- **Unit tests**: added
- **Docs**: n/a
- **Backwards compatibility**: adds a status field to analytics; see above

### DIFF
--- a/pkg/analytics/client.go
+++ b/pkg/analytics/client.go
@@ -14,15 +14,23 @@ import (
 )
 
 type (
-	Client struct {
-		UserId     string                 `json:"id"`
-		Event      string                 `json:"event"`
-		Source     []byte                 `json:"source,omitempty"`
-		Properties map[string]interface{} `json:"properties,omitempty"`
+	Payload struct {
+		UserId     string         `json:"id"`
+		Event      string         `json:"event"`
+		Source     []byte         `json:"source,omitempty"`
+		Properties map[string]any `json:"properties"`
 	}
+
+	Client struct {
+		serverUrlOverride   string
+		userId              string
+		universalProperties map[string]any
+	}
+
 	ErrorHandler interface {
 		PrintErr(err error)
 	}
+
 	LogLevel string
 )
 
@@ -37,68 +45,51 @@ const (
 const datadogLogLevel = "_logLevel"
 const datadogStatus = "status"
 
-func NewClient(properties map[string]interface{}) *Client {
+func NewClient() *Client {
 	local := GetOrCreateAnalyticsFile()
 
-	client := &Client{
-		Properties: properties,
-	}
+	client := &Client{}
+	client.universalProperties = make(map[string]any)
 
 	// These will get validated in AttachAuthorizations
-	client.UserId = local.Id
-	client.Properties["validated"] = false
+	client.userId = local.Id
+	client.universalProperties["validated"] = false
 
-	client.Properties["localId"] = local.Id
+	client.universalProperties["localId"] = local.Id
 	if runUuid, err := uuid.NewRandom(); err == nil {
-		client.Properties["runId"] = runUuid.String()
+		client.universalProperties["runId"] = runUuid.String()
 	}
 
 	return client
 }
 
 func (t *Client) AttachAuthorizations(claims *auth.KlothoClaims) {
-	t.Properties["localId"] = t.UserId
-	t.UserId = claims.Email
-	t.Properties["validated"] = claims.EmailVerified
+	t.universalProperties["localId"] = t.userId
+	t.userId = claims.Email
+	t.universalProperties["validated"] = claims.EmailVerified
 }
 
 func (t *Client) Info(event string) {
-	t.Properties[datadogLogLevel] = Info
-	t.track(event)
-}
-
-func (t *Client) Debug(event string) {
-	t.Properties[datadogLogLevel] = Debug
-	t.track(event)
+	t.Send(t.createPayload(Info, event))
 }
 
 func (t *Client) Warn(event string) {
-	t.Properties[datadogLogLevel] = Warn
-	t.Properties[datadogStatus] = Warn
-	t.track(event)
+	t.Send(t.createPayload(Warn, event))
 }
 
 func (t *Client) Error(event string) {
-	t.Properties[datadogLogLevel] = Error
-	t.Properties[datadogStatus] = Error
-	t.track(event)
+	t.Send(t.createPayload(Error, event))
 }
 
-func (t *Client) Panic(event string) {
-	t.Properties[datadogLogLevel] = Panic
-	// Using error since datadog does not support panic for the reserved status field
-	t.Properties[datadogStatus] = Error
-	t.track(event)
+func (p *Payload) addError(err error) *Payload {
+	p.Properties["error"] = fmt.Sprintf("%+v", err)
+	return p
 }
 
 func (t *Client) AppendProperties(properties map[string]interface{}) {
 	for k, v := range properties {
-		t.Properties[k] = v
+		t.universalProperties[k] = v
 	}
-}
-
-func (t *Client) DeleteProperty(key string) {
-	delete(t.Properties, key)
 }
 
 func (t *Client) UploadSource(source *core.InputFiles) {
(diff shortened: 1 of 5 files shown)


yuval-klotho:
I found myself getting a bit lost as I was writing this (much of it yesterday, when I was a bit foggy), so please do take a look to see if I did something wonky in terms of an awkward internal API or struct usage. It's probably due to piecemeal evolving code, rather than a principled stance on code structure.

pkg/analytics/client_test.go
```diff
@@ -56,14 +62,156 @@ func TestAnalytics_Hash(t *testing.T) {
...
+					"validated":  false,
+					"property_1": "aaa",
+				},
+			}},
+		},
+		{
+			name: "send via logger with no fields",
+			send: func(c *Client) {
+				c.userId = "my-user@klo.dev"
+				logger := zap.New(c.NewFieldListener(zapcore.WarnLevel))
```
gordon-klotho, line 101: nit: this is more of a test for the FieldListener

pkg/analytics/fields_listener.go
```diff
@@ -45,38 +45,37 @@ func (fl *fieldListener) Write(entry zapcore.Entry, fields []zapcore.Field) erro
...
+		logLevel = Info
+	case zapcore.WarnLevel:
+		logLevel = Warn
+	case zapcore.ErrorLevel:
+		logLevel = Error
+	case zapcore.DPanicLevel, zapcore.PanicLevel, zapcore.FatalLevel:
+		logLevel = Panic
+	}
+
+	p := fl.client.createPayload(logLevel, entry.Level.CapitalString())
```
gordon-klotho, line 62: Shouldn't the event be the same (like "log") since the level is already in the properties? Seems odd to have this duplicated is all

pkg/analytics/client.go
```diff
@@ -37,98 +45,87 @@ const (
...
+	t.Send(p)
 }
 
-func (t *Client) track(event string) {
-	t.Event = event
-	err := SendTrackingToServer(t)
-
-	if err != nil {
-		zap.L().Debug(fmt.Sprintf("Failed to send metrics info. %v", err))
+func (t *Client) createPayload(level LogLevel, event string) *Payload {
```
gordon-klotho, line 106: Does this need to return a pointer?

pkg/analytics/tracking_utils.go
```diff
@@ -5,30 +5,34 @@ import (
...
-var kloServerUrl = "http://srv.klo.dev"
+const kloServerUrl = "http://srv.klo.dev"
 
 type AnalyticsFile struct {
 	Id string
 }
 
-func SendTrackingToServer(bundle *Client) error {
-	postBody, _ := json.Marshal(bundle)
+func (t *Client) Send(payload *Payload) {
```
gordon-klotho, line 22: Nice, much better signature. Does Payload need to be a pointer though?

pkg/analytics/tracking_utils.go
```diff
@@ -5,30 +5,34 @@ import (
...
+	postBody, _ := json.Marshal(payload)
 	data := bytes.NewBuffer(postBody)
-	resp, err := http.Post(fmt.Sprintf("%v/analytics/track", kloServerUrl), "application/json", data)
-	if err != nil {
-		return err
+	url := t.serverUrlOverride
+	if url == "" {
+		url = kloServerUrl
 	}
+	resp, err := http.Post(fmt.Sprintf("%v/analytics/track", url), "application/json", data)
```
gordon-klotho, line 29: I think `%s` is a better fit for this than `%v` since we specifically want the `.String()` method which gives the correct value. It ends up being the same in this case, but IMO it's clearer that you're not putting in like `url.URL{Scheme: "http", ...}`

pkg/analytics/tracking_utils.go
```diff
@@ -5,30 +5,34 @@ import (
...
 
-	defer resp.Body.Close()
-
-	return nil
+	if err != nil {
+		zap.L().Debug(fmt.Sprintf("Failed to send metrics info. %v", err))
```
gordon-klotho, line 32: This could cause infinite loop if the server is down. Also, the previous version returning the error to let the caller decide what to do makes the most sense IMO.

pkg/cli/klothomain.go
```diff
@@ -255,7 +256,7 @@ func (km KlothoMain) run(cmd *cobra.Command, args []string) (err error) {
 	defer analyticsClient.PanicHandler(&err, errHandler)
 
 	updateStream := options.Update.Stream.OrDefault(km.DefaultUpdateStream)
-	analyticsClient.Properties["updateStream"] = updateStream
+	analyticsClient.AppendProperties(map[string]any{"updateStream": updateStream})
```
gordon-klotho, line 259: Might be worth adding a singular form of this method so we don't need to make these single-item maps which are cumbersome.

gordon-klotho (approved):
Nothing blocking

pkg/analytics/client_test.go
yuval-klotho: yeah, good point. This is one of those "my brain wasn't quite working right" things. :) Will fix.

pkg/analytics/client.go
yuval-klotho: Done mostly for efficiency (though it's not a big struct), and especially since this is an un-exported func.

Is there any sort of convention for when to return a pointer vs a plain struct?

pkg/analytics/tracking_utils.go
yuval-klotho: Good call. It was `%v` before and I didn't revisit that, but I agree `%s` is clearer.

pkg/analytics/tracking_utils.go
yuval-klotho: The infinite loop was there before, too :-\ I noticed it but didn't feel like thinking about how to clean it up, heh.

Regarding returning the error: in the previous version, there was only one caller,`(*Client).track`:

```
func (t *Client) track(event string) {
	t.Event = event
	err := SendTrackingToServer(t)

	if err != nil {
		zap.L().Debug(fmt.Sprintf("Failed to send metrics info. %v", err))
	}
}
```

So:
- `SendTrackingToServer` returns a possible error
- `Track` is the _only_ invoker of `SendTrackingToServer`, and the only real value it provides is to turn that error into a log line

It felt like inlining those made sense. That said, maybe it would be clearer if I (a) un-exported `SendTrackingToServer` and (b) moved it into `client.go`?

pkg/cli/klothomain.go
yuval-klotho: Good call.

(the main branch has it the worst of both worlds: it only has this map-based approach, but in a couple places we call that one several times in a row, each time with a one-element map!)

pkg/analytics/fields_listener.go
yuval-klotho: I think I prefer this a bit, just because it makes it a bit easier as you're scanning the all-events dashboard to visually pick out warnings vs logs.

pkg/analytics/client_test.go
yuval-klotho: The part I was actually interested in was the non-global properties, so I just wrote that. FieldListener is as untested as it was before.

PR #285, opened 2023-02-23
improve auth analytics

### LINKED ISSUE #218: client-level analytics for tracking auth events

yuval-klotho:
Always get a LoginInfo from the authorizer; this is never nil, and always contains an "Authorizer" field which should be a non-empty string describing the authorizer (for now, just "auth0" or "local"). Then, always attach that info to the analytics. This will let us know e.g. how many compiles are using local vs auth0.

This resolves #218

### Standard checks

- **Unit tests**: n/a
- **Docs**: n/a

### DIFF
--- a/pkg/cli/klothomain.go
+++ b/pkg/cli/klothomain.go
@@ -37,10 +37,10 @@ type KlothoMain struct {
 }
 type Authorizer interface {
 
-	// Authorize tries to authorize the user. The klothoClaims it returns may be nil, even if the authentication
-	// succeeds. Conversely, if the klothoClaims is non-nil, it is valid even if the error is also non-nil; you can use
-	// those claims provisionally (and specifically, in analytics) even if the error is non-nil, indicating failed
-	// authentication.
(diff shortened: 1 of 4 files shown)


pkg/cli/klothomain.go
```diff
@@ -37,11 +37,11 @@ type KlothoMain struct {
 }
 type Authorizer interface {
 
-	// Authorize tries to authorize the user. The KlothoClaims it returns may be nil, even if the authentication
-	// succeeds. Conversely, if the KlothoClaims is non-nil, it is valid even if the error is also non-nil; you can use
+	// Authorize tries to authorize the user. The klothoClaims it returns may be nil, even if the authentication
```
gordon-klotho, line 40:
```suggestion
	// Authorize tries to authorize the user. The `LoginInfo` it returns may be nil, even if the authentication
```

gordon-klotho (approved):

PR #289, opened 2023-02-25
Replace deprecated command with environment file

### LINKED ISSUE #288: Replace deprecated `set-output` command with environment file
## Description

In `.github/workflows/release.yaml`, `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information, see: [https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)

jongwooo:
## Description

Resolve #288 

Update `.github/workflows/release.yaml` to use environment file instead of deprecated `set-output` command. 
For more information, see: [https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)

I found the workflow files that use `set-output` command through the following command:

```bash
$ find . -name '*.yml' -o -name '*.yaml' | xargs egrep '\bset-output\b'
```

**AS-IS**

```yml
run: echo ::set-output name=VERSION::$(echo $GITHUB_REF | cut -d / -f 3)
```

**TO-BE**

```yml
run: echo VERSION=$(echo $GITHUB_REF | cut -d / -f 3) >> $GITHUB_OUTPUT
```

### Standard checks

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -25,7 +25,7 @@ jobs:
 
       - name: Get the version
         id: get_version
-        run: echo ::set-output name=VERSION::$(echo $GITHUB_REF | cut -d / -f 3)
+        run: echo VERSION=$(echo $GITHUB_REF | cut -d / -f 3) >> $GITHUB_OUTPUT
 
       - name: Build
         run: |


gordon-klotho (approved):

gordon-klotho:
Thanks for the update, glad to know our releases won't be broken come June 1st.

PR #320, opened 2023-03-04
make new function to use for generating env vars

### LINKED ISSUE #317: Invalid S3 Bucket Name Environment Variable
When attempting to deploy an app containing a cloud file system, an invalid bucket name environment variable is generated when the Klotho resource's ID contains invalid environment variable name characters.

This results in a failed deployment with the following error:
```
Diagnostics:
  pulumi:pulumi:Stack (DavidSeptimus-Klotho/ts-app/ds-ts-app)
    error: update failed
 
  aws:lambda:Function (executor)
    error: 1 error occurred:
	* creating Lambda Function (ts-app-executor): ValidationException: 
	status code: 400, request id: 21d4e862-1327-448e-82ff-76b82c9a72a4
```

Here's the underlying error returned by AWS when manually creating the function via the AWS CLI:

```
An error occurred (ValidationException) when calling the CreateFunction operation: 1 validation error detected: Value at 'environment.variables' failed to satisfy constraint: Map keys must satisfy constraint: [Member must satisfy regular expression pattern: [a-zA-Z]([a-zA-Z0-9_])+]
```

**Example:**

Resource:
```typescript
/**
 * @klotho::persist {
 * id = "fs-cloud"
 * }
 */
import fs = require("fs/promises");
```

Invalid environment variable:
`fs-cloud_bucket_name`

jhsinger-klotho:
• Does it relate to or fix any issue? closes #317 

remove special characters from env var keys. Using a New method to be called which does the checks but i dont like that we cant always enforce it. Will have to be sure on code reviews

### Standard checks

- **Unit tests**: Any special considerations? updated


DavidSeptimus-Klotho:
Separate from this PR, we also need to do the same type of sanitization when generating new variable names. Additionally, in both cases we need to have a strategy for dealing with resource names that are equivalent when sanitized to avoid conflicts either resulting in deconfliction or in raising an error to the user -- requiring them to make a change to their resource names.

DavidSeptimus-Klotho (approved):

Issue #230, opened 2023-02-14
eks upgrade failures


jhsinger-klotho:
saw this error when debugging

```
ConnectionError [SequelizeConnectionError]: connect ETIMEDOUT 10.0.200.114:5432
    at Client._connectionCallback (/usr/src/app/node_modules/sequelize/lib/dialects/postgres/connection-manager.js:143:24)
    at Client._handleErrorWhileConnecting (/usr/src/app/node_modules/pg/lib/client.js:316:19)
    at Client._handleErrorEvent (/usr/src/app/node_modules/pg/lib/client.js:326:19)
    at Connection.emit (node:events:512:28)
    at Socket.reportStreamError (/usr/src/app/node_modules/pg/lib/connection.js:57:12)
    at Socket.emit (node:events:512:28)
    at emitErrorNT (node:internal/streams/destroy:151:8)
    at emitErrorCloseNT (node:internal/streams/destroy:116:3)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21) {
  parent: Error: connect ETIMEDOUT 10.0.200.114:5432
      at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1494:16)
      at TCPConnectWrap.callbackTrampoline (node:internal/async_hooks:130:17) {
    errno: -110,
    code: 'ETIMEDOUT',
    syscall: 'connect',
    address: '10.0.200.114',
    port: 5432
  },
  original: Error: connect ETIMEDOUT 10.0.200.114:5432
      at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1494:16)
      at TCPConnectWrap.callbackTrampoline (node:internal/async_hooks:130:17) {
    errno: -110,
    code: 'ETIMEDOUT',
    syscall: 'connect',
    address: '10.0.200.114',
    port: 5432
  }
}

```

jhsinger-klotho:
```
 -   ├─ aws:ec2:SecurityGroupRule                                                     farg-prox-private-ingress                                                                         deleted (0.75s)
 -   ├─ aws:ec2:SecurityGroupRule                                                     farg-prox-public-ingress                                                                          deleted (0.59s)
 - ```

amihan also pointed out that these rules were getting removed for some reason. When you run pulumi refresh you see them deleted and then recreated when you run up again. Not sure why

jhsinger-klotho:
when you re run pulumi up those are recreated and then tests pass

PR #324, opened 2023-03-06
create sg rules all in the same construct

jhsinger-klotho:
• Does it relate to or fix any issue? #230 and this is more widely applicable now with the new integration tests, we see it popping up for ecs as well.

https://github.com/pulumi/pulumi-aws/issues/2391 has context on the pulumi side

### Standard checks

- **Backwards compatibility**: Will this break existing apps? If so, what would be the extra work required to keep them working? it should because it will overwrite the SG with the right rule set

### DIFF
--- a/pkg/infra/pulumi_aws/deploylib.ts
+++ b/pkg/infra/pulumi_aws/deploylib.ts
@@ -179,62 +179,62 @@ export class CloudCCLib {
         const sgName = sanitized(AwsSanitizer.EC2.vpc.securityGroup.nameValidation())`${h(
             this.name
         )}`
-        const klothoSG = new aws.ec2.SecurityGroup(sgName, {
-            name: sgName,
-            vpcId: this.klothoVPC.id,
-            egress: [
-                {
-                    cidrBlocks: ['0.0.0.0/0'],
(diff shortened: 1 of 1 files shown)


pkg/infra/pulumi_aws/deploylib.ts
```diff
@@ -179,62 +179,62 @@ export class CloudCCLib {
...
+                        {
+                            description:
+                                'Allows inbound traffic from network interfaces and instances that are assigned to the same security group.',
+                            fromPort: 0,
+                            protocol: '-1',
+                            self: true,
+                            toPort: 0,
+                        },
+                        {
+                            description: 'For EKS control plane',
```
gordon-klotho, line 211: Should this only get added if there's an eks being set up?

gordon-klotho (approved):

pkg/infra/pulumi_aws/deploylib.ts
jhsinger-klotho: well we arent smart enough for that yet because we dont split up exec units by type. So ideally yes, but its not really doable without a major refactor

PR #325, opened 2023-03-06
pass resource counts as list to easily filter

jhsinger-klotho:
• Does it relate to or fix any issue? we want this as a list for ease in mixpanel

### Standard checks


yuval-klotho (commented):
I don't really understand the purpose/goal here, but the change seems fine. But... what's the goal? You say you're doing it so that you can easily filter, but I don't see what you're filtering, or why.

yuval-klotho (approved):

PR #344, opened 2023-03-09
install fluentbit on nodes for eks

### LINKED ISSUE #205: send EKS logs to CW using fluent bit
https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Container-Insights-EKS-logs.html

This will also allow people to view container logs within CW Logs rather than having to view it through kubectl or other mechanisms. This will be beneficial for containers which have been replaced, and for users who are not as familiar with kubernetes

jhsinger-klotho:
• Does it relate to or fix any issue? closes #205 

installing fluent bit on our nodes and making everything run on nodes instead of fargate by default since we cant solely run on fargate due to some plugins. 

This will ship logs on the nodes back to cloudwatch so that we can pull them easily without having to do cluster work in our tests and useful for others. 

Enabled by default but we can expose it to be disabled

### Standard checks

### DIFF
--- a/pkg/infra/pulumi_aws/iac/eks.ts
+++ b/pkg/infra/pulumi_aws/iac/eks.ts
@@ -439,6 +439,10 @@ export class Eks {
         }
     }
 
+    public getClusterName(): pulumi.Output<string> {
+        return this.cluster.name
+    }
+
     private determineNodeGroupSpecs(execUnits: EksExecUnit[]): Map<string, NodeGroupSpecs> {
         let nodeGroupSpecs: Map<string, NodeGroupSpecs> = new Map<string, NodeGroupSpecs>()
         const defaultDiskSize = 20
--- a/pkg/infra/pulumi_aws/iac/eks.ts
+++ b/pkg/infra/pulumi_aws/iac/eks.ts
@@ -7,7 +7,13 @@ import * as pulumi_k8s from '@pulumi/kubernetes'
 import * as k8s from './kubernetes'
 import * as https from 'https'
 import { getIssuerCAThumbprint } from '@pulumi/eks/cert-thumprint'
-import { cloud_map_controller, alb_controller, external_dns } from './k8s/add_ons/'
+import {
+    cloud_map_controller,
+    alb_controller,
+    external_dns,
+    installFluentBitForCW,
+    enableFargateLogging,
+} from './k8s/add_ons/'
 import { local } from '@pulumi/command'
 import { CloudCCLib, Resource } from '../deploylib'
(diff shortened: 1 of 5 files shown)


pkg/infra/pulumi_aws/iac/k8s/add_ons/fluent_bit_for_eks/index.ts
```diff
@@ -0,0 +1,80 @@
...
+                },
+            },
+            { provider, dependsOn: [cwNamespace] }
+        )
+    })
+
+    new pulumi_k8s.yaml.ConfigFile(
+        `${eks.clusterName}-FluentBitDriver`,
+        {
+            file: 'https://raw.githubusercontent.com/aws-samples/amazon-cloudwatch-container-insights/latest/k8s-deployment-manifest-templates/deployment-mode/daemonset/container-insights-monitoring/fluent-bit/fluent-bit.yaml',
```
DavidSeptimus-Klotho, line 33: Is it normal to be reading the config file directly from a samples repo?

pkg/infra/pulumi_aws/iac/eks.ts
```diff
@@ -117,13 +123,14 @@ interface EksClusterOptions {
...
     enableFargateLogging?: Boolean
+    enableFluentBitLogging?: Boolean
     adminRoleArn?: string
     autoApply?: boolean
     createNodeGroup?: boolean
 }
 
 export const DefaultEksClusterOptions: EksClusterOptions = {
-    initializePluginsOnFargate: true,
+    initializePluginsOnFargate: false,
```
DavidSeptimus-Klotho, line 133: What's the impact of this change?

pkg/infra/pulumi_aws/iac/k8s/add_ons/fluent_bit_for_eks/index.ts
jhsinger-klotho: yeah its fairly normal. When we do a lot of the k8s work well likely pull these back in, but its going to be very messy to do this in pulumi. Anyways, not an abnormal patter for people to follow, thats essentially what youre doing with helm anyways

pkg/infra/pulumi_aws/iac/eks.ts
jhsinger-klotho: the alb controller and coredns will be run on our node group we create instead of fargate now. We always created the nodegroup so was just added cost having them be on fargate

DavidSeptimus-Klotho (approved):

PR #349, opened 2023-03-13
delete before replace will cause slight downtime, so remove

jhsinger-klotho:

avoid causing downtime on deployments

### Standard checks

### DIFF
--- a/pkg/infra/pulumi_aws/iac/api_gateway.ts
+++ b/pkg/infra/pulumi_aws/iac/api_gateway.ts
@@ -390,12 +390,6 @@ export class ApiGateway {
                 }
             }
 
-            //create the methods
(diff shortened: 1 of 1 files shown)


jhsinger-klotho:
> Do the comments not apply? It sounds like we needed this, and removing it would cause us to break when we switch exec unit types?

My belief is thats what the triggers on the deployment do. Either way, causing downtime is something we shouldnt do so i think we have to remove this and see if it truly breaks deployments because we cant leave this in

yuval-klotho (approved):

yuval-klotho:
wfm, though I do hope this doesn't cause a regression :)

If it does, hopefully our new tests will catch it!

gordon-klotho (approved):

PR #362, opened 2023-03-14
remove http import if not used

### LINKED ISSUE #332: Determine whether net/http needs to be removed in go lambdas
If the only usage of this import in code is the listener then we get 

```
#11 134.4 ./main.go:9:2: "net/http" imported and not used
```
during the dockerbuild

also remove the / route on our go mega app when this is complete

jhsinger-klotho:
• Does it relate to or fix any issue? closes #332 

will find any identifier or package identifier for the net/http import and see if its used in that file. Are there any other node types im missing?

### Standard checks

### DIFF
--- a/pkg/lang/golang/plugin_expose.go
+++ b/pkg/lang/golang/plugin_expose.go
@@ -1,6 +1,7 @@
 package golang
 
 import (
+	"fmt"
 	"path"
 	"strings"
 
@@ -183,6 +184,11 @@ func (h *restAPIHandler) handleFile(f *core.SourceFile) (*core.SourceFile, error
 			return nil, core.NewCompilerError(f, capNode, err)
 		}
 
+		err = h.removeNetHttpImport(f)
+		if err != nil {
+			return nil, core.NewCompilerError(f, capNode, err)
+		}
+
 		router, err := h.findChiRouterDefinition(f, routerName)
 		if err != nil {
 			return nil, core.NewCompilerError(f, capNode, err)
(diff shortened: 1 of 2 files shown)


pkg/lang/golang/plugin_expose.go
```diff
@@ -239,6 +245,37 @@ func (h *restAPIHandler) handleFile(f *core.SourceFile) (*core.SourceFile, error
 	return f, nil
 }
 
+func (h *restAPIHandler) removeNetHttpImport(f *core.SourceFile) error {
+	h.log.Info(fmt.Sprintf("searching for http imports in %s", f.Path()))
+	netHttpImportName := "http"
+	netHttpImport := GetNamedImportInFile(f, netHttpImportName)
+	if netHttpImport.Alias != "" {
```
DavidSeptimus-Klotho, line 252: This might speak to a larger issue in how we handle Go imports (since `UpdateImportsInFile` operates on package names during removal), but shouldn't we be removing the specific import associated with the `net/http` code commented expose plugin (if it's otherwise aliased in the file, that's not relevant to our code transformations)? And if that's true, shouldn't we only be removing it if we're certain that the import isn't used elsewhere in the file aside from the location's we've commented out?

DavidSeptimus-Klotho: Got it. In that case, the only relevant issue is that of alternate aliases and the fact that `UpdateImportsInFile` will nuke them since it removes imports by package name rather than their local name.

https://github.com/KlothoPlatform/klotho/blob/4147135f475ace101c68efe2c8cb7b6b84c52d7a/pkg/lang/golang/imports.go#L79-L83

DavidSeptimus-Klotho: Agreed.

DavidSeptimus-Klotho (approved):

PR #364, opened 2023-03-15
use separate stacks for different regions

jhsinger-klotho:

because oss and pro use the same actions, this was causing the stack to have to be deleted and recreated each time. We will need to clean up the account, it seems that theres multiple stack resources created which we need to delete

### Standard checks

### DIFF
--- a/.github/workflows/run-integ-tests.yaml
+++ b/.github/workflows/run-integ-tests.yaml
@@ -131,7 +131,7 @@ jobs:
           set -x
           cd _test_resources
           npm i
-          pulumi -s integ-test-shared stack select --create
+          pulumi -s integ-test-shared-${{ env.AWS_REGION }} stack select --create
           pulumi stack tag set usage integ-test
           pulumi config refresh || true
           pulumi up -y --refresh


.github/workflows/run-integ-tests.yaml
```diff
@@ -131,7 +131,7 @@ jobs:
           set -x
           cd _test_resources
           npm i
-          pulumi -s integ-test-shared stack select --create
+          pulumi -s integ-test-shared-${{ env.AWS_REGION }} stack select --create
```
gordon-klotho, line 134: IMO we shouldn't use region as proxy for separating different actions and instead use some input that can be kept constant but separate between the two repos.

jhsinger-klotho: i could use the sample apps repo name instead if you would prefer that. 

The only benefit i see with region is pro and oss have the same exact resources in the shared stack, so if we did deploy them in the same region is less running services = lower cost. That may not be the case forever though if those tests change, but my hunch is they will always support the same shared resources which is why i opted for region

gordon-klotho: Ah, if we're intending pro and oss to share then that makes sense. 👍

gordon-klotho (approved):

PR #382, opened 2023-03-20
Implements annotation handling enhancements

DavidSeptimus-Klotho:
Resolves klothoplatform/klotho-history#321

- Uninterrupted sequential line comments are merged into a single block
- Multiline comments are never merged with other comments
- An annotation annotates the named sibling node immediately succeeding its parent comment unless that node is also a comment
- Sequential annotations spanning one or more comment blocks are now handled with only the last annotation in the sequence annotating the succeeding sibling node.

- ~~Added IsTransformed flag to annotations that is for indicating that the node annotated by an annotation has been transformed and is no longer suitable for further modification. This flag also results in an annotation being detached from the annotated node when a SourceFile is reparsed.~~

- Added `IsDetached` flag to annotations indicating that the node annotated by an annotation has been transformed in a way that makes it no longer suitable for further transformation (e.g. commented out). This flag also results in an annotation being detached from the annotated node when a SourceFile is reparsed.

### Standard checks

- **Docs**: Do we need to update any docs, internal or public? no, unless we want to add some initial documentation describing how annotations are handled


pkg/annotation/capability.go
```diff
@@ -13,40 +13,87 @@ type Capability struct {
...
 	}
+	return capabilities, nil
+}
 
-	return cap, nil
+// getExprEndIndex gets the index of end delimiter in a balanced expression of start and end delimiters.
+//
+// The escape argument is used for detecting escaped delimiters and should either be `\` or `\\`
+// depending on the format of the input string.
+func getExprEndIndex(input, escape string, start, end rune) int {
```
gordon-klotho, line 69: This should get unit tested

pkg/annotation/capability.go
```diff
@@ -13,40 +13,87 @@ type Capability struct {
...
-				return cap, fmt.Errorf("'id' must be less than 25 characters in length. 'id' was %s", id)
+				return nil, fmt.Errorf("'id' must be less than 25 characters in length. 'id' was %s", id)
 			}
 			match, err := regexp.MatchString(`^[\w-_.:/]+$`, id)
 			if err != nil {
-				return cap, errors.Wrap(err, "could not parse 'id' directive")
+				return nil, errors.Wrap(err, "could not parse 'id' directive")
 			} else if !match {
-				return cap, fmt.Errorf("'id' can only contain alphanumeric, -, _, ., :, and /. 'id' was %s", id)
+				return nil, fmt.Errorf("'id' can only contain alphanumeric, -, _, ., :, and /. 'id' was %s", id)
```
gordon-klotho, line 55: nit: I generally prefer printing the regexp that it matches against in the error so it always matches (the compiled regexp has a `.String()` method).

pkg/annotation/capability.go
```diff
@@ -13,40 +13,87 @@ type Capability struct {
...
+	return capabilities, nil
+}
 
-	return cap, nil
+// getExprEndIndex gets the index of end delimiter in a balanced expression of start and end delimiters.
+//
+// The escape argument is used for detecting escaped delimiters and should either be `\` or `\\`
+// depending on the format of the input string.
+func getExprEndIndex(input, escape string, start, end rune) int {
+	escapedStartPattern := regexp.MustCompile(fmt.Sprintf(`^[^%c]*?((?:%s)*)\%c`, start, escape, start))
```
gordon-klotho, line 70: Considering we only ever call this with `{}`, we could prevent the costly creation of regexp (compiling a regexp >> matching a regexp) in the inner loop and instead only do it once.

jhsinger-klotho:
> * Added IsTransformed flag to annotations that is for indicating that the node annotated by an annotation has been transformed and is no longer suitable for further modification. This flag also results in an annotation being detached from the annotated node when a SourceFile is reparsed.

Im not sure that this assumption will always be true, but fine for now. Theres potentially times when we will transform something and another plugin will still pick it up and transform it again in a non invasive way

pkg/annotation/capability_test.go
```diff
@@ -115,19 +115,19 @@ func TestParseCapability(t *testing.T) {
...
 		b = 2
 	}`,
-			want: &Capability{
+			want: []*Capability{{
 				Name: "thing",
 				Directives: Directives{
 					"key1": map[string]interface{}{"a": int64(1), "b": int64(2)},
 				},
-			},
+			}},
```
gordon-klotho, line 123: I would expect additional test cases here to cover where multiple capabilities are returned

pkg/core/annotation.go
```diff
@@ -17,6 +17,9 @@ type (
 		Capability *annotation.Capability
 		// Node is the node that has been annotated; not the comment node representing the annotation itself.
 		Node *sitter.Node
+		// IsTransformed indicates whether the content of an annotation's original Node field been transformed.
+		// When reparsing a transformed annotation, its Node field will be nil to prevent plugins from treating the transformed code as input.
+		IsTransformed bool
```
gordon-klotho, line 22: What do we need this for? The node file gets updated when the code is transformed so it's always correct.

Unless you mean `IsDetached` - not all transformations would result in detached nodes.

pkg/lang/capabilities.go
```diff
@@ -52,64 +57,87 @@ func CompositePreprocessor(preprocessors ...CommentPreprocessor) CommentPreproce
...
 }
 
+func IsCLineCommentBlock(previous, current *sitter.Node) bool {
+	return previous != nil && current != nil &&
+		current.StartPoint().Row-previous.StartPoint().Row == 1 &&
+		strings.HasPrefix(current.Content(), "//") &&
+		strings.HasPrefix(previous.Content(), "//")
+}
+
+func IsNumberCommentBlock(previous, current *sitter.Node) bool {
```
gordon-klotho, line 67: I was confused for a bit until I remembered `#` is called "number sign" sometimes. Considering the other function is `IsDoubleSlashCommentBlock` then perhaps we should name this one `IsPythonCommentBlock`? At least I'd use "hash" instead since I think that's the more common name for that character.

Alternatively, add a common implementation for the two like
```go
func CommentBlockByPrefix(prefix string) MergeCommentsPredicate {}
```
(function returning a function)

DavidSeptimus-Klotho: > Considering the other function is IsDoubleSlashCommentBlock

It's called `IsCLineCommentBlock`.

> perhaps we should name this one IsPythonCommentBlock? At least I'd use "hash" instead since I think that's the more common name for that character.

"Number" is just carried over from how VS Code refers to them. Hash is fine too.

gordon-klotho: > It's called IsCLineCommentBlock.

Oops, I meant "is not" 😄

pkg/core/source_file.go
```diff
@@ -41,7 +41,11 @@ func (lid LanguageId) CastFile(f File) (*SourceFile, bool) {
 	return nil, false
 }
 
-func (f *SourceFile) Reparse(newProgram []byte) (err error) {
+func (f *SourceFile) Reparse(newProgram []byte, transformedAnnotations ...*Annotation) (err error) {
+	for _, a := range transformedAnnotations {
+		a.IsTransformed = true // prevents reattachment of the annotation to a potentially transformed succeeding sibling node
+	}
```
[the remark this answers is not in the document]
DavidSeptimus-Klotho: I started down that path initially. The primary reason for adding it to `Reparse` is to remind developers to consider whether their file transformations require marking any annotations.

pkg/annotation/capability.go
DavidSeptimus-Klotho: Been meaning to convert this to a function that returns a function. to avoid recompiling the regex.

pkg/annotation/capability.go
```diff
@@ -13,40 +13,87 @@ type Capability struct {
...
+var capabilityStartRegex = regexp.MustCompile(`\s*@klotho::(\w+)\s*(\{)?`)
 
-func ParseCapability(s string) (*Capability, error) {
-	matches := capabilityRE.FindStringSubmatch(s)
-	if len(matches) < 2 {
-		return nil, nil
-	}
+func ParseCapabilities(s string) ([]*Capability, error) {
+	var capabilities []*Capability
+	matches := capabilityStartRegex.FindAllStringSubmatchIndex(s, -1)
```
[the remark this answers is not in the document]
DavidSeptimus-Klotho: sure, I can rename the related variables.

pkg/core/annotation.go
```diff
@@ -42,7 +45,7 @@ func (a *Annotation) MarshalJSON() ([]byte, error) {
 	return json.Marshal(m)
 }
 
-func (a Annotation) Format(s fmt.State, verb rune) {
+func (a *Annotation) Format(s fmt.State, verb rune) {
```
gordon-klotho, line 48: Why change these to pointer? The compiler will automatically reference/dereference as needed so this shouldn't be necessary. It is a read-only function so does not need a pointer receiver.

pkg/core/annotation.go
```diff
@@ -58,15 +61,28 @@ func (a Annotation) Format(s fmt.State, verb rune) {
...
+}
+
 func (m AnnotationMap) Update(other AnnotationMap) {
 	for k, v := range other {
 		if ex, ok := m[k]; ok {
 			// Update the contents not the pointer so existing annotation pointers are still valid
-			*ex = *v
+			newValue := *v
+			if ex.isDetached {
+				newValue.isDetached = true
```
gordon-klotho, line 83: Should this also clear out the `Node` since it's invalid when detached (ie, call `newValue.Detach()`)

pkg/lang/capabilities.go
```diff
@@ -2,33 +2,38 @@ package lang
...
-	"io"
-	"regexp"
-
 	"github.com/klothoplatform/klotho/pkg/annotation"
 	"github.com/klothoplatform/klotho/pkg/core"
 	"github.com/klothoplatform/klotho/pkg/multierr"
 	"github.com/klothoplatform/klotho/pkg/query"
 	"github.com/pkg/errors"
 	sitter "github.com/smacker/go-tree-sitter"
+	"io"
```
gordon-klotho, line 11: nit: Is this an IDE change? I generally prefer the imports to be std lib > empty line > external libs > empty line > internal packages (which I believe VS Code auto-formats this via the gopls language server).

pkg/lang/capabilities.go
```diff
@@ -52,64 +57,87 @@ func CompositePreprocessor(preprocessors ...CommentPreprocessor) CommentPreproce
...
-		cap, err := annotation.ParseCapability(block.comment)
-		if cap == nil {
-			continue
-		}
-		annotation := &core.Annotation{Capability: cap, Node: block.node}
-		if err != nil {
-			merr.Append(core.NewCompilerError(f, annotation, errors.Wrap(err, "error parsing annotation")))
-			continue
+	for _, block := range c.findAllCommentBlocks(f) {
+		caps, err := annotation.ParseCapabilities(block.comment)
```
gordon-klotho, line 94: If this returns no caps, then the err is unchecked. Getting 1 error per cap when it's only generated once is odd.

pkg/lang/capabilities.go
```diff
@@ -52,64 +57,87 @@ func CompositePreprocessor(preprocessors ...CommentPreprocessor) CommentPreproce
...
 		match, found := nextMatch()
 		if !found || match == nil {
 			break
 		}
 		capture := match[fullCaptureName]
-		comment := capture.Content()
-		comment = c.preprocessor(comment)
+		comment := c.preprocessor(capture.Content())
+		annotatedNode := capture.NextNamedSibling()
+		if annotatedNode != nil && strings.Contains(annotatedNode.Type(), "comment") {
```
gordon-klotho, line 125: Checking the node type is language specific. Maybe check the `c.sitterQuery` against it?

pkg/parseutils/parse_utils_test.go
```diff
@@ -0,0 +1,95 @@
+package parseutils
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestExpressionExtractor(t *testing.T) {
```
gordon-klotho, line 8: 👍

pkg/lang/capabilities.go
DavidSeptimus-Klotho: yeah, it looks like `gofmt` (intellij default) vs `goimports` -- which I can change on my end.

pkg/core/annotation.go
DavidSeptimus-Klotho: It's just an idiomatic go thing. Mixing pointer and value receivers is bad practice and can mess with a struct's ability to implement interfaces.

pkg/lang/capabilities.go
DavidSeptimus-Klotho: I guess I can execute the full query and check if the next match == the annotated node

pkg/lang/capabilities.go
DavidSeptimus-Klotho: Yeah, the error handling here needs to be updated to account for multiple annotations in a block.

pkg/core/annotation.go
gordon-klotho: > can mess with a struct's ability to implement interfaces.

In this case, it was intentionally not a pointer so the following would work properly:
```go
fmt.Sprintf("%+v", Annotation{...})
fmt.Sprintf("%+v", &Annotation{...}) // this works as well since the compiler will automatically dereference
```

pkg/core/errors.go
```diff
@@ -91,11 +91,13 @@ func (err *CompileError) Format(s fmt.State, verb rune) {
 			fmt.Fprint(s, "\n")
 			err.Annotation.Format(s, verb)
 			fmt.Fprintf(s, "\nin %s\n", err.File.Path())
-			fnode := &NodeContent{
-				Endpoints: err.Annotation.Node,
-				Content:   err.Annotation.Node.Content(),
+			if err.Annotation.Node != nil {
```
gordon-klotho, line 94: The trailing newline that this condition would cause (from the Fprintf in the previous line) would be awkward wouldn't it?

pkg/lang/javascript/nestjs.go
```diff
@@ -122,6 +122,7 @@ func (p *NestJsHandler) handleFile(f *core.SourceFile, unit *core.ExecutionUnit)
 
 		actedOn, newfileContent := p.actOnAnnotation(f, &listen, fileContent, appName, p.Config.GetResourceType(unit), cap.ID)
 		if actedOn {
+			annot.Detach()
```
gordon-klotho, line 125: Shouldn't this only detach if the code was commented out? That doesn't always happen if actedOn I don't think (eg, EKS case)

gordon-klotho (commented):
Mostly good - just a few minor comments

pkg/lang/javascript/nestjs.go
DavidSeptimus-Klotho: You're right. I'll move this to the lambda section in `actOnAnnotation` for both JS expose plugins.

pkg/annotation/capability_test.go
```diff
@@ -75,63 +76,134 @@ func TestParseCapability(t *testing.T) {
...
+		},
+		{
+			name: "escaped glob pattern is not parseable",
+			text: `
+			@klotho::thing1 {
+			  id = "#$%#$%sdfDSFdgdfgdf_(dfgdfgdfggdfgdfg"
+			}
+			@klotho::thing {
+			  included = "**\/*.js"
+			}`,
```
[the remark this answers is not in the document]
DavidSeptimus-Klotho: Replacing this with multiple test cases:
- fail if ID is too long
- fail if ID has invalid characters
- fail if a value includes an invalid TOML escape sequence (i.e. `\/` in `**\/*.js`)

pkg/core/errors.go
DavidSeptimus-Klotho: 🤷‍♂️  I guess I can just move that newline inside the if block.

gordon-klotho (approved):

PR #386, opened 2023-03-21
Input output files

### LINKED ISSUE #374: Output files
Core should write Output files in addition to the WriteTo to allow for migration

Out of scope - update existing outputs to use this output files.

jhsinger-klotho:
• Does it relate to or fix any issue? closes #374 

Creating the compilation document and outputTo method for it. 

This change will break all existing functionality until we switch over plugins to using the documents input files rather than looking for them within the compilation result. The next step is to start removing CloudResource and replace it with what is currently named ProviderResource and start using the graphs

### Standard checks


yuval-klotho (approved):
lgtm pending our discussion earlier

PR #388, opened 2023-03-21
Cli refactor

jhsinger-klotho:
• Does it relate to or fix any issue? follows https://github.com/klothoplatform/klotho/pull/387

updates to the cli and other minor places to start working in the new constructs

Adds some of the old helper functions which were being used to get resources based on type, etc.

Adds some basic functions to the graph class to get a single vertex or all vertices

- Removes Validation plugin from workflow which will need to be added back 

### Standard checks

### DIFF
--- a/pkg/graph/graph.go
+++ b/pkg/graph/graph.go
@@ -85,6 +85,33 @@ func (d *Directed[V]) AddEdge(source V, dest V) {
 	}
 }
 
+func (d *Directed[V]) GetVertex(source string) V {
+	v, err := d.underlying.Vertex(source)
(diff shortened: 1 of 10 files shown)


pkg/graph/graph.go
```diff
@@ -85,6 +85,33 @@ func (d *Directed[V]) AddEdge(source V, dest V) {
 	}
 }
 
+func (d *Directed[V]) GetVertex(source string) V {
+	v, err := d.underlying.Vertex(source)
+	if err != nil && !errors.Is(err, graph.ErrEdgeAlreadyExists) {
```
yuval-klotho, line 90: `ErrEdgeAlreadyExists` is a copypasta mistake, I assume?

PR #394, opened 2023-03-23
v1 of the iac plugin

yuval-klotho:
This chunk of code is still somewhat WIP, but I think it's in good enough shape to get it into the feature branch so that we can crash on it to connect it with the providers and make it real.

The "main event" here is `templatesCompiler`. It takes in a graph (expected to be a DAG) of cloud resources, and translates each one into a `const varname = ....` statement. For now, the idea is they'd all go into a single `index.ts` -- but we can think about splitting those up later. You can see this in effect at [`pkg/infra/iac2/templates_compiler_test.go`](#diff-64c964dcad5352d51084d8ccf37174d598e25bf0e943181d90af8f118aae4462).

To do this, we create a `ResourceCreationTemplate` per struct type in the graph. This template contains info parsed from `templates/${struct name in snake case}/factory.ts`, which has to be in a specific format: it has to contain an `interface Args`, and a `function create(args: Args): <type> { /* ... */}`.

We then use this template to create the specific resources we need, in topological order, and assign a variable to each.

Some outstanding work (probably for a follow-on PR):

- [ ] templatesCompiler's traversal is currently hacky and assumes a tree; I need to hook it up to `graph.Directed`, but I wanted to wait for any changes that Jordan's previous PR had for that module
- [ ] make sure the traversal is stable across runs (it currently isn't); useful both for unit tests and for customers who want to diff output across runs
- [ ] inline Args into the `create(...)` function; I realize now that we don't actually need the separate `Args`
- [ ] the current implementation assumes all of the resource fields are structs, not pointers; needs a bit of work to make this all work well
- [ ] the `resource` struct is a bit hacky, in part because I wanted to wait for the structs from the provider before I got too far along that. It's basically meant as a placeholder to cache/store info on each struct, but we may not need it in the end
- [ ] we can add unit tests that make sure that each struct lines up with its corresponding template (right args, etc). That'll come later.
- [ ] the actual IAC templates are stubbed out for now

### Standard checks

- **Unit tests**: n/a
- **Docs**: n/a


pkg/infra/iac2/templates/docker_lambda/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "iac-fragment",
+  "dependencies": {
+    "@pulumi/aws": "^5.16.0",
+    "@pulumi/pulumi": "^3.40.2"
```
jhsinger-klotho, line 5: what do we do if these conflict between 2 of the templates?

pkg/infra/iac2/templates_compiler.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+	tc.resources[v] = res
+	tc.resourceGraph.AddVertex(res)
+	for _, child := range getStructValues(v) {
+		if reflect.TypeOf(child).Kind() == reflect.Struct {
+			tc.AddResource(child)
+			childRes := tc.getResource(child)
+			tc.resourceGraph.AddEdge(res, childRes)
+		}
+	}
```
jhsinger-klotho, line 74: why are the struct values dependencies? Maybe im a little confused on what `getStructValues` is returning

pkg/infra/iac2/iac_templates.go
```diff
@@ -0,0 +1,121 @@
...
+	findArgsQuery string
+
+	//go:embed find_create_func.scm
+	findCreateFuncQuery string
+
+	//go:embed find_imports.scm
+	findImportsQuery string
+)
+
+func ParseResourceCreationTemplate(contents []byte) ResourceCreationTemplate {
```
gordon-klotho, line 47: I'm curious why this approach instead of using Go templates like the previous iteration? I don't too much care for the "magic" of requiring `interface Args` and such over explicit template variables. Are there considerations I'm missing?

pkg/infra/iac2/templates/cloudwatch_log_group/factory.js
yuval-klotho, line 1: will rm this file (and add a gitignore for it)

pkg/infra/iac2/templates/docker_lambda/factory.js
yuval-klotho, line 1: will rm this file (and add a gitignore for it)

pkg/infra/iac2/templates_compiler.go
```diff
@@ -0,0 +1,189 @@
...
+		if _, err := out.Write([]byte(statement)); err != nil {
+			return err
+		}
+		if _, err := out.Write([]byte("\n")); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
```
jhsinger-klotho, line 124: will this prevent dedupes? or is there any js logic we can reuse? I assume we want them at the top of the file, but im not quite sure if this will handle that

pkg/infra/iac2/iac_templates.go
jhsinger-klotho: ill let yuval explain more, but the reasoning was because we can then unit test and ensure that the ts files compile correctly

pkg/infra/iac2/templates_compiler.go
yuval-klotho: If you have:

```
type Foo {
  hello string
  bar Bar
}

foo := Foo {
  hello: "world",
  bar: Bar{},
}
```

then `getStructValues(foo)` would return:

```
map[string]any {
  "hello": "world",
  "bar": Bar{},
}
```

This then means that `foo` depends on that `Bar{}`.

jhsinger-klotho: by doing this secondary graph i think were doing extra work that we dont need to do. I see the usefulness in tracking the variable names, files, etc that the resource gets placed in, but if we recreate we could have a discrepancy between iac graph and provider resource graph

yuval-klotho: Oh, I see what you mean. Yeah, this part can probably go away once I hook this up to the actual provider structs.

Basically, I needed a few changes to the `graph.Directed` to make the rendering work, but I didn't want to do that while your other PR was still up, in case we'd conflict (since I thought you were making some changes in that). So this is just a bit of scaffolding that will go away once we hook it up to the real provider structs.

pkg/infra/iac2/iac_templates.go
gordon-klotho: Both approaches can be unit tested (but must be compiled first) so I don't see that as an advantage. Type-checking at build time is near use-less because they are pre-transformation so that has no bearing over whether the template is valid.

My main concern is that each transformation in this approach takes a lot of effort (multi-line queries, multi-line glue code in Go) compared to just `{{.Foo}}`.

yuval-klotho: > Type-checking at build time is near useless

We have had _many_ bugs (granted, mostly in our runtime) due to typos and similar in non-go code that we embed. So I do think even just "aws.lambda.Function(...) requires certain args, do we have the right ones?" is useful to check.

For things like "do we have the right dependencies", were you envisioning something like a piece of go code that takes a template, compiles it with some dummy data (which may be non-trivial, for required args) to produce a valid `.ts` file, and then execs out to `npm install && npx tsc` to check that it's valid?

Another part of the testing that I forgot to mention is how the various resources fit together. For example, let's say you have a `Foo` resource whose template creates a `pulumi.Output<Foo>`. Then you have a `Bar` resource that takes in a `Foo` arg. It'd be nice if we could tell at build/test time that these things don't match up:

```go
type (
  Foo struct {}  // template produces a pulumi.Output<Foo>
  Bar struct {
    Foo *Foo // template requires a Foo, not pulumi.Output<Foo>
  }
)
```

If the templates only use `{{.Foo}}`, I think figuring that out is pretty hard. Whereas, if they're wrapped in this function, then it's pretty easy in a test to figure out:

* `Foo`'s template will generate a `"plumi.Output<Foo>"`
* `Bar.Foo` is a `Foo`, but its template requires a `"Foo"`
* `"pulumi.Output<Foo>" != "Foo"`, so test fails

gordon-klotho: > We have had many bugs (granted, mostly in our runtime) due to typos and similar in non-go code that we embed.

No matter which approach, we can't make any guarantees that post-transformation will compile (or work) based on pre-transformation evaluations. Thus, we must make tests (unit or otherwise) for these in post-transformation. Given that, I don't see much incremental value in pre-transformation validation or testing.

> For things like "do we have the right dependencies", were you envisioning something like a piece of go code that takes a template, compiles it with some dummy data (which may be non-trivial, for required args) to produce a valid .ts file, and then execs out to npm install && npx tsc to check that it's valid?

Something like that yeah. Could even be non-trivial and run `jest` or `pulumi policy` on it (in addition to `tsc`). Could even be per-resource (baring dependencies). This kind of test is roughly what I had in mind for a post-transformation test.

> For example, let's say you have a Foo resource whose template creates a pulumi.Output<Foo>. Then you have a Bar resource that takes in a Foo arg.

`pulumi.Input` was designed for this purpose, so I'd propose we leverage that. Both `pulumi.Ouput<Foo>` and `Foo` satisfy `pulumi.Input<Foo>`. 

In the current approach, how would it resolve this? If one function returns `string` (say the createImage), and another uses `pulumi.Output<string>`, would it happen somewhere in the caller side to convert?

> If the templates only use {{.Foo}}, I think figuring that out is pretty hard. Whereas, if they're wrapped in this function, then it's pretty easy in a test to figure out.

Figuring that out from template side might be hard, but wouldn't the current approach have the same problems? You'd have to look at both templates to figure out that their types mismatch (except with the current approach, you'd also have to check the transformation code to make sure it didn't change anything either - unless we conventionally only transform a narrow portion of a template). I think both approaches are equally easy to test post-generation - both would fail a `tsc`.

jhsinger-klotho (approved):

pkg/infra/iac2/find_imports.scm
```diff
@@ -0,0 +1,3 @@
+(program
```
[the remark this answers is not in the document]
gordon-klotho: Technically these as TS files and not JS so it likely doesn't apply. Also, since these only check our code, we can be more rigid about the types of import statements we use.

pkg/infra/iac2/iac_templates.go
gordon-klotho: We talked offline. Summary: I didn't understand that these were "templates for templates" that are used to help write the templates in a way that is type checked and can carry the imports necessary. This approach is 💯 , just need some thought on template-level control flow (can be deferred until we need it).
Apologies on the churn ❤️ I should have read the whole thing before commenting 😄

pkg/infra/iac2/iac_templates.go
```diff
@@ -0,0 +1,121 @@
...
+	return result
+}
+
+// doQuery is a thin wrapper around `query.Exec` to use typescript as the Language.
+func doQuery(c *sitter.Node, q string) query.NextMatchFunc {
+	return query.Exec(tsLanguage, c, q)
+}
+
+// parameterizeArgs turns "args.foo" into {{.Foo}}. It's very simplistic and just works off regex
+func parameterizeArgs(contents string) string {
```
gordon-klotho, line 101: As discussed in our chat, would be nice to use treesitter instead of regexp for this.
Here's what I threw together in the playground during the call in case it's useful:
```
(member_expression
	object: (identifier) @obj
	(#eq @obj "args")
	property: (property_identifier) @prop
 )
```

pkg/infra/iac2/cloud_resources_stub.go
```diff
@@ -0,0 +1,20 @@
...
+		// fields...
+	}
+)
+
+func (cwg CloudwatchLogGroup) VariableName() string {
+	return "cloudwatchLogGroup"
+}
+
+func (dl DockerLambda) VariableName() string {
+	return "lambda_" + nonIdentifierChars.ReplaceAllString(dl.ExecUnitName, "_")
```
gordon-klotho, line 19: Hm, if there's any characters that aren't `_`, this could create name conflicts, right? One alternative to consider, and this is a little clever, is to use the names as-is in a `Map` (or object properties) instead of as standalone variables. We could restrict the IDs for now to just be `[a-zA-Z_]+` for now and not need the replacement too.

pkg/infra/iac2/iac_templates.go
```diff
@@ -0,0 +1,121 @@
...
+	returnTypeFunc := query.Select(doQuery(node, findCreateFuncQuery), query.ContentOf(query.ParamNamed("return_type")))
+	returnType, err := returnTypeFunc()
+	if !err {
+		// unexpected, since all inputs are from resources in the klotho binary
+		panic("couldn't find valid create() function")
+	}
+	result.outputType = returnType
+
+	// return expression
+	returnBodyFunc := query.Select(doQuery(node, findCreateFuncQuery), query.ContentOf(query.ParamNamed("return_body")))
```
gordon-klotho, line 74: nit: can the `doQuery` results be extracted into its own variable so it's not queried twice (first time on L65).

gordon-klotho (approved):
Overall 👍 nice

pkg/infra/iac2/iac_templates.go
yuval-klotho: I'll do that early next week (just so we can focus on gluing Jordan's and my work tonight/tomorrow). I've got a todo list locally, so I won't forget!

pkg/infra/iac2/cloud_resources_stub.go
yuval-klotho: Yeah, I should have added a comment — this is very much just stubbed out for "good enough for a demo on Friday". Sorry I forgot to document that!

I think vars are a bit nicer for the user than a map, because they can use their IDE to jump around between usage and declaration. But what I'd like eventually is for (a) the name to be automatically generated from the resource type and id, and (b) if there are any conflicts (due to `_`-type smushing or anything else), have the `templatesCompiler` realize that and just assign each a unique number suffix:

```
id="foo/bar"    => const lambdaFooBar_0
id="foo-bar"    => const lambdaFooBar_1
```

Thoughts on that?

`lambdaFoo_Bar_1`

pkg/infra/iac2/iac_templates.go
yuval-klotho: I tried it (after your comment), and it failed — the extracted var maintains state, do the second `Select` fails.

Since you gave the approval, I'll go ahead and merge as-is for now. But, open question: should I just remove these `query.Select`s altogether (if so, it'll probably be next week). I personally like them, but if I'm the only one (and I think I may be) then it's probably not worth the overall cognitive load.

PR #395, opened 2023-03-23
base provider spike - iam role, ecr, lambda

jhsinger-klotho:

This PR Provides the initial code for the spike on the provider (Still a little rough/first pass so feedback = good(

This pr enables taking a ConstructGraph from our compiler and transforming it into the necessary ResourceGraph

What is enabled:
  - ECR Repo
  - ECR Image
  - IAM Role
  - Lambda Function

The internal representation of each aws resource is specified in the `provider/aws/resources/{{service}}` directory. Each of these resources implements the necessary functions to make it compliant with Resource interface.

We also try to use Constructors where we have sanitization so that theres a centralized method for creation.

Sanitization is still very basic and can only do max length + regex replacements. We will want to expand on this to get parity with the deploy lib version.

### Standard checks

### DIFF
--- a/pkg/config/execution_unit.go
+++ b/pkg/config/execution_unit.go
@@ -1,5 +1,11 @@
 package config
 
+import (
+	"encoding/json"
+
+	"go.uber.org/zap"
+)
+
 type (
 	// ExecutionUnit is how execution unit Klotho constructs are represented in the klotho configuration
 	ExecutionUnit struct {
@@ -87,4 +93,21 @@ func (a Application) GetExecutionUnit(id string) ExecutionUnit {
 		cfg.InfraParams = defaultParams
 	}
 	return cfg
+}
+
+func (cfg ExecutionUnit) GetExecutionUnitParamsAsKubernetes() KubernetesTypeParams {
+
+	infraParams := cfg.InfraParams
+	jsonString, err := json.Marshal(infraParams)
+	if err != nil {
+		zap.S().Debug(err)
+	}
+
+	params := KubernetesTypeParams{}
+	if err := json.Unmarshal(jsonString, &params); err != nil {
+		zap.S().Debug(err)
+		return params
+	} else {
+		return params
+	}
 }
--- a/pkg/core/exec_unit.go
+++ b/pkg/core/exec_unit.go
@@ -1,6 +1,7 @@
 package core
 
 import (
+	"fmt"
 	"io"
 	"os"
 	"path/filepath"
@@ -154,6 +155,7 @@ func (unit *ExecutionUnit) AddSourceFile(f File) {
 
 func (unit *ExecutionUnit) AddStaticAsset(f File) {
 	if f != nil {
+		fmt.Println(unit)
 		unit.files.Set(f.Path(), f)
 		unit.Executable.StaticAssets[f.Path()] = struct{}{}
 	}
--- a/pkg/provider/aws/execution_unit.go
+++ b/pkg/provider/aws/execution_unit.go
@@ -1,0 +1,73 @@
+package aws
+
+import (
+	"fmt"
+
+	"github.com/klothoplatform/klotho/pkg/config"
+	"github.com/klothoplatform/klotho/pkg/core"
+	"github.com/klothoplatform/klotho/pkg/provider/aws/resources"
+	"github.com/klothoplatform/klotho/pkg/provider/aws/resources/ecr"
+	"github.com/klothoplatform/klotho/pkg/provider/aws/resources/iam"
+	"github.com/klothoplatform/klotho/pkg/provider/aws/resources/lambda"
+	"go.uber.org/zap"
+)
+
+func (a *AWS) GenerateExecUnitResources(unit *core.ExecutionUnit, dag *core.ResourceGraph) error {
+	log := zap.S()
+
+	execUnitCfg := a.Config.GetExecutionUnit(unit.ID)
+
+	// See if we have already created an ecr repository for the app and if not create one, otherwise add a ref to this exec unit
+	repo := dag.GetResource(ecr.GenerateRepoId(a.Config.AppName))
+	if repo == nil {
+		repo = ecr.NewEcrRepository(a.Config.AppName, unit.Provenance())
+		dag.AddResource(repo)
+	} else {
+		repo, ok := repo.(*ecr.EcrRepository)
+		if !ok {
+			return fmt.Errorf("expected resource with id, %s, to be ecr repository", repo.Id())
+		}
+		repo.ConstructsRef = append(repo.ConstructsRef, unit.Provenance())
+	}
+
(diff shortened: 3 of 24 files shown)


pkg/provider/aws/execution_unit.go
```diff
@@ -0,0 +1,73 @@
...
+
+	switch execUnitCfg.Type {
+	case Lambda:
+		lambda := lambda.NewLambdaFunction(unit, a.Config.AppName, role)
+		dag.AddResource(lambda)
+		dag.AddDependency(role, lambda)
+		dag.AddDependency(image, lambda)
+		return nil
+	case Eks:
+	case Ecs:
```
yuval-klotho, line 51: should we comment these two out for now, so that they'll result in the Error message? (go cases don't fall through)

pkg/provider/aws/execution_unit_test.go
```diff
@@ -0,0 +1,182 @@
...
+func Test_GetAssumeRolePolicyForType(t *testing.T) {
+	cases := []struct {
+		name  string
+		cfg   config.ExecutionUnit
+		wants string
+	}{
+		{
+			name:  `lambda`,
+			cfg:   config.ExecutionUnit{Type: Lambda},
+			wants: "{\n\tVersion: '2012-10-17',\n\tStatement: [\n\t\t{\n\t\t\tAction: 'sts:AssumeRole',\n\t\t\tPrincipal: {\n\t\t\t\tService: 'lambda.amazonaws.com',\n\t\t\t},\n\t\t\tEffect: 'Allow',\n\t\t\tSid: '',\n\t\t},\n\t],\n}",
```
yuval-klotho, line 155: a backticked-string (so it can be multiline, etc) might make these more readable

pkg/config/execution_unit.go
```diff
@@ -88,3 +94,20 @@ func (a Application) GetExecutionUnit(id string) ExecutionUnit {
...
+
+	infraParams := cfg.InfraParams
+	jsonString, err := json.Marshal(infraParams)
+	if err != nil {
+		zap.S().Debug(err)
+	}
+
+	params := KubernetesTypeParams{}
+	if err := json.Unmarshal(jsonString, &params); err != nil {
+		zap.S().Debug(err)
```
gordon-klotho, line 108: I'd think this should be an `Error` log since if this happens something went wrong, right?

pkg/core/exec_unit.go
```diff
@@ -154,6 +155,7 @@ func (unit *ExecutionUnit) AddSourceFile(f File) {
 
 func (unit *ExecutionUnit) AddStaticAsset(f File) {
 	if f != nil {
+		fmt.Println(unit)
```
gordon-klotho, line 158: debugging left in here :)

pkg/provider/aws/infra_template.go
```diff
@@ -1,9 +1,44 @@
...
 import (
+	"fmt"
+
 	"github.com/klothoplatform/klotho/pkg/core"
+	"go.uber.org/zap"
 )
 
 func (a *AWS) Translate(result *core.ConstructGraph, dag *core.ResourceGraph) (Links []core.CloudResourceLink, err error) {
+	fmt.Println(dag.ListConstructs())
+	fmt.Println(dag.ListDependencies())
```
gordon-klotho, line 12: more debugging :)

yuval-klotho (approved):
a couple small things, nothing blocker. You're getting these out so fast!

pkg/provider/aws/infra_template.go
```diff
@@ -1,9 +1,44 @@
...
+
+	log := zap.S()
+
+	rootConstructs := result.GetRoots()
+	for _, construct := range rootConstructs {
```
gordon-klotho, line 17: How come only the roots?

pkg/provider/aws/infra_template.go
```diff
@@ -1,9 +1,44 @@
...
+
+	}
+	constructs := dag.ListConstructs()
+	for _, c := range constructs {
+		fmt.Println(c)
+	}
+	deps := dag.ListDependencies()
+	for _, c := range deps {
+		fmt.Printf("Source: %s, Dest: %s\n", c.Source.Id(), c.Destination.Id())
+	}
```
gordon-klotho, line 42: Would be nice if you could just `fmt.Println(dag)`

pkg/sanitization/aws/s3/s3.go
```diff
@@ -1,4 +1,4 @@
-package resources
+package s3
```
gordon-klotho, line 1: How come `s3` is its own sanitization package and `iam`/`lambda`/etc are together at top-level?

pkg/sanitization/sanitizer.go
```diff
@@ -19,10 +22,13 @@ func (s *Sanitizer) Apply(input string) string {
 	for _, rule := range s.rules {
 		output = rule.Pattern.ReplaceAllString(output, rule.Replacement)
 	}
+	if s.maxLength != 0 && s.maxLength < len(output) {
```
gordon-klotho, line 25: Couldn't maxLength be achieved in regexp?
```go
Rule{
    Pattern: regexp.MustCompile(`(.{1,32}).*`),
    Replacement: "$1",
}
```

pkg/provider/aws/execution_unit.go
jhsinger-klotho: Yeah i need to likely go clean this all up, some of this and the other switch statements were just brain dumped that i left over

pkg/config/execution_unit.go
jhsinger-klotho: yeah likely, since you only call this method with the expectation they can be marshalled correctly. Ill make that change

pkg/provider/aws/infra_template.go
jhsinger-klotho: Right now the graph is purely just a single lambda and generating from there. I need to change this so it will support more, but focused more on just getting the generation done for lambda so we could try to wire it all together. Will make the changes to add more after this pr

pkg/provider/aws/infra_template.go
jhsinger-klotho: I removed this i think, but i guess its a UX question on do we want to print anything out in the aws resource level? Ill bring that up tomorrow

pkg/sanitization/aws/s3/s3.go
jhsinger-klotho: need to refactor this, i didnt mean to touch it let me change it back

pkg/sanitization/sanitizer.go
jhsinger-klotho: it could, i think we need to actually put some effort into the sanitizer so i decided to just do this as a placholder

PR #396, opened 2023-03-24
Read cors middleware to setup OPTIONS routes for go

gordon-klotho:
Tested locally that the headers get set and the routes get added.
image_360

Did not have a repro case that would actually send the preflight OPTIONS requests.

### Standard checks

### DIFF
--- a/pkg/lang/golang/plugin_expose.go
+++ b/pkg/lang/golang/plugin_expose.go
@@ -45,6 +45,7 @@ type (
 	}
 
 	chiRouterDefResult struct {
+		Name        string
 		Declaration *sitter.Node
 		Identifier  *sitter.Node
 		RootPath    string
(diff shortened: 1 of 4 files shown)


jhsinger-klotho:
Do we need to do this for other languages as well?

pkg/lang/golang/plugin_expose.go
```diff
@@ -469,6 +492,25 @@ func (h *restAPIHandler) findChiRouterMountPackage(f *core.SourceFile, mount *ro
...
+			break
+		}
+		args := match["args"]
+		for i := 0; i < int(args.ChildCount()); i++ {
+			mw = append(mw, args.Child(i))
+		}
+	}
+	return mw
+}
+
```
jhsinger-klotho, line 513: are we able to add unit tests for some of these functions?

jhsinger-klotho (approved):

gordon-klotho:
> Do we need to do this for other languages as well?

I didn't want to preempt someone needing it, but we should eventually yeah

PR #399, opened 2023-03-24
some cleanup and adding s3 + accountID to graph

jhsinger-klotho:

Adding S3 and accountID (GetCallerIdentity in pulumi) Objects and templates.

Some small cleanup and additional unit tests as well

### Standard checks


yuval-klotho (approved):

pkg/provider/aws/plugin.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+func (a *AWS) Translate(result *core.ConstructGraph, dag *core.ResourceGraph) (Links []core.CloudResourceLink, err error) {
+	log := zap.S()
+
+	constructIds, err := result.TopologicalSort()
+	if err != nil {
+		return
+	}
+	// We want to reverse the list so that we start at the leaf nodes. This allows us to check downstream dependencies each time and process them.
```
gordon-klotho, line 17: Not sure I understand what this means / why we need to do it in reverse.

gordon-klotho (approved):

PR #400, opened 2023-03-24
various fixes to get the iac2 plugin working (‼️)

yuval-klotho:
- CloudResources -> Resources
- replace old Pulumi plugin with the new iac2.Plugin
- add the IacPlugins loop to compiler.go
- in ResourceCreationTemplate:
  - escape `{` when it appears in the ts, so that `{args.Foo}` doesn't become `{{{args.Foo}}}` (which is invalid go-template)
  - give templates a name (just the struct they came from), for errors
- in templatesCompiler:
  - when rendering types, iterate over the type's template Args, not the type's fields; that way, we silently ignore any fields that the template doesn't use (which we do want to silently ignore)
  - handle all (?) of the types that the struct field could be. The original impetus was slices, but once I did those I figured I may as well do maps, and once I did those I figured I may as well do everything.
  - some typing stuff around the graph, which needs to be further refined
  - formatting stuff for the output (adding newlines)

### Standard checks

- **Unit tests**: **I owe unit tests for a lot of this**. I'm posting it as-is so that Jordan can sync on it, but I will be writing those tests.
- **Docs**: I also owe internal docs for this; ditto above.
- **Backwards compatibility**: "move fast and break things"


pkg/core/resource_graph.go
```diff
@@ -6,32 +6,32 @@ import (
 
 type (
 	ResourceGraph struct {
-		underlying *graph.Directed[Resource]
+		Underlying *graph.Directed[Resource]
```
jhsinger-klotho, line 9: Did we want to move this back to being hidden?

pkg/infra/iac2/templates_compiler.go
```diff
@@ -103,42 +111,102 @@ func (tc templatesCompiler) RenderImports(out io.Writer) error {
 }
 
 func (tc templatesCompiler) renderResource(out io.Writer, resource graph.Identifiable) error {
```
jhsinger-klotho, line 113: isnt resource of type core.Resource?

jhsinger-klotho (approved):
2 minor comments, but not blocking

pkg/core/resource_graph.go
yuval-klotho: Eventually/soon, but as the code stands now, it wouldn't compile. It's not a hard thing to fix, but if you wanted to pull my changes quickly, this is probably the fastest path. (I have a TODO to roll this back on `templatesCompiler`)

pkg/compiler/compiler.go
```diff
@@ -14,21 +14,18 @@ type (
 	}
 
 	AnalysisAndTransformationPlugin interface {
-		Name() string
-
+		Plugin
```
gordon-klotho, line 17: 👍

pkg/compiler/compiler.go
```diff
@@ -72,15 +69,26 @@ func (c *Compiler) Compile() error {
...
 		if err != nil {
 			return core.NewPluginError(p.Name(), err)
 		}
 		c.Document.Configuration.AddLinks(links)
 		log.Debug("completed")
 	}
 
-	return nil
+	for _, p := range c.IaCPlugins {
+		if isPluginNil(p) {
```
gordon-klotho, line 81: I know we did this before, so this is just a side commentary- Do we really need to be defensive about this? You'd have to go fairly out of your way to add a typed nil to the plugins.

pkg/infra/iac2/iac_templates.go
```diff
@@ -96,7 +99,12 @@ func doQuery(c *sitter.Node, q string) query.NextMatchFunc {
 
 // parameterizeArgs turns "args.foo" into {{.Foo}}. It's very simplistic and just works off regex
 func parameterizeArgs(contents string) string {
-	return parameterizeArgsRegex.ReplaceAllString(contents, `{{$1}}`)
+	// If the source has "{args.Foo}", then just turning "args.Foo" -> "{{.Foo}}" would result in "{{{.Foo}}}", which is
+	// invalid go-template. So, we first turn "{args." into "{{`{`}}args.", which will eventually result in
+	// "{{`{`}}{{.Foo}}" — which, while ugly, will result in the correct template execution.
+	contents = curlyEscapes.ReplaceAllString(contents, "{{`$1`}}args")
```
gordon-klotho, line 105: Just to make sure, this would be valid right? `{{'{'}}{{.Foo}}}` (the ending triple `}}}`)

pkg/infra/iac2/plugin.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+)
+
+func (p Plugin) Name() string {
+	return "pulumi2"
+}
+
+func (p Plugin) Translate(cloudGraph *core.ResourceGraph) ([]core.File, error) {
+
+	// TODO We'll eventually want to split the output into different files, but we don't know exactly what that looks
```
gordon-klotho, line 19: 👍 good comment

pkg/infra/iac2/templates_compiler.go
```diff
@@ -102,43 +110,103 @@ func (tc templatesCompiler) RenderImports(out io.Writer) error {
...
+	_, err = out.Write([]byte(";"))
 	if err != nil {
 		return err
 	}
 
 	return errs.ErrOrNil()
 }
 
+// resolveStructInput translates a value to a form suitable to inject into the typescript as an input to a function.
+func (tc templatesCompiler) resolveStructInput(childVal reflect.Value) string {
```
gordon-klotho, line 159: Isn't this just writing to JSON? Kind of hard to tell, but seems like it. (JSON being the "JavaScript Object Notation" and all, it should be what we want)

pkg/infra/iac2/iac_templates.go
yuval-klotho: Yep, that's valid (and is in fact what we generate).

pkg/compiler/compiler.go
yuval-klotho: Jordan and I were confused by this, too! I'll make a note to nix it in the next round of cleanups.

pkg/infra/iac2/templates_compiler.go
yuval-klotho: Very close, but not quite. The two big differences are:

- for structs (and pointers to structs), including ones deep in arrays/maps, we render them as the variable name that the struct had been rendered into (`tc.getVarName(typedChild)`)
- for strings, we use backtick strings instead of json strings, so that multiline messes like policy documents render nicely; otherwise, instead of
  ```
  assumeRolePolicy: `{
      Version: '2012-10-17',
      Statement: [
        {
          Action: 'sts:AssumeRole',
          Principal: {
            Service: 'lambda.amazonaws.com',
          },
          Effect: 'Allow',
          Sid: '',
        },
      ],
    }`,
  ```
  ... you get:
  ```
  assumeRolePolicy: "{\n\tVersion: '2012-10-17',\n\tStatement: [\n\t\t{\n\t\t\tAction: 'sts:AssumeRole',\n\t\t\tPrincipal: {\n\t\t\t\tService: 'lambda.amazonaws.com',\n\t\t\t},\n\t\t\tEffect:
'Allow',\n\t\t\tSid: '',\n\t\t},\n\t],\n}"
  ```

gordon-klotho: Makes sense 👍 (could that description go in the godocs)

For the multiline stuff, we'll likely need an indentation-level for true pretty printing, but good enough for now definitely.

yuval-klotho: Will definitely add it!

PR #401, opened 2023-03-25
Render package json for iac2

jhsinger-klotho:

Adds function to merge dependencies of package.json in the main javascript package.

Uses the method and loops over the templates package.jsons to combine them.

Adds the 2 pulumi files to render to output of the compiled dir and renames where the new iac2 compiles to so we can test deployments (The context/dockerfile are based on compiled root so deployments wouldnt succeed otherwise.

Fixed regex for string variables within the tempaltes.

### Standard checks


pkg/infra/iac2/templates_compiler.go
```diff
@@ -110,6 +124,24 @@ func (tc templatesCompiler) RenderImports(out io.Writer) error {
 	return nil
 }
 
+func (tc templatesCompiler) RenderPackageJSON() (javascript.NodePackageJson, error) {
+	errs := multierr.Error{}
+	mainPJson := javascript.NodePackageJson{}
+	for _, res := range tc.resourceGraph.GetAllVertices() {
+		pJson, err := tc.GetPackageJSON(res)
+		if err != nil {
+			fmt.Println(err)
```
gordon-klotho, line 133: nit: debugging

pkg/infra/iac2/templates_compiler.go
```diff
@@ -110,6 +124,24 @@ func (tc templatesCompiler) RenderImports(out io.Writer) error {
...
+			errs.Append(err)
+			continue
+		}
+		mainPJson.Merge(&pJson)
+	}
+	if err := errs.ErrOrNil(); err != nil {
+		return *mainPJson.Clone(), err
+	}
+	return *mainPJson.Clone(), nil
```
gordon-klotho, line 142: No need for clone here since `mainPJson` isn't used anywhere else.

gordon-klotho (approved):

pkg/infra/iac2/templates_compiler.go
jhsinger-klotho: yeah but because it contains a mux i believe go static check yells at you if you return it directly

pkg/lang/javascript/nodejs.go
```diff
@@ -37,6 +38,22 @@ func (f *PackageFile) Path() string {
 	return f.path
 }
 
+func (f *PackageFile) MergeDeps(other *PackageFile) {
+	for dep, version := range other.Content.Dependencies {
+		currentVersion, ok := f.Content.Dependencies[dep]
+		if ok {
+			if currentVersion != version {
+				zap.S().Warnf(`Found conflicting dependencies in package.json.
```
yuval-klotho, line 46: Nice, I like this approach and the logging!

Only thought is, do we today have any other log messages that are multiline within the message itself? An alternate way to do this would be to have the message just be `Found conflicting dependencies in package.json`, and model the original/current/used as a zap field.

pkg/infra/iac2/templates_compiler.go
```diff
@@ -1,26 +1,32 @@
...
+	"github.com/pkg/errors"
+	"go.uber.org/zap"
 )
 
 type (
 	templatesCompiler struct {
 		// templates is the fs.FS where we read all of our `<struct>/factory.ts` files
 		templates fs.FS
+		// packageFiles is the fs.FS where we read all of our `<struct>/package.json` files
+		packageFiles fs.FS
```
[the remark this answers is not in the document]
jhsinger-klotho: it wouldnt be different i wouldnt think, but i wasnt sure so i split it out. We can definitely keep them as one if we want, but my other thought is for standardizing the templates vs the project files maybe it would be useful. Not sure though

pkg/lang/javascript/nodejs.go
jhsinger-klotho: yeah let me see how that looks compared to this

pkg/infra/iac2/templates_compiler.go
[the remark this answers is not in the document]
yuval-klotho: I'd prefer to keep them as one for now:

1. Within Go code it keeps things simpler and with less chance of divergence (where we set one field but forget to set the other)
2. Within the file structure itself, I think we always want each `factory.ts` to sit next to its `package.json` so that we can do an `npm install && npx tsc && npx depcheck` on it.

PR #402, opened 2023-03-25
initial IR of networking resources

jhsinger-klotho:

Creates IR for all networking related resources we need to spin up our current representation of vpc. 

Creates some common methods for creating these resources and adding their dependencies

### Standard checks


pkg/provider/aws/resources/vpc/vpc.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func CreateNetwork(appName string, dag *core.ResourceGraph) {
+	vpc := NewVpc("test-app")
+	region := resources.NewRegion()
+
+	dag.AddResource(region)
+	dag.AddResource(vpc)
+
+	CreateGatewayVpcEndpoint("s3", vpc, region, dag)
```
gordon-klotho, line 56:
```suggestion
	// TODO add logic around VpcEndpoints to only include those which are needed
	CreateGatewayVpcEndpoint("s3", vpc, region, dag)
```

gordon-klotho (approved):

PR #404, opened 2023-03-27
Fixes issue where qualified module name is truncated for aliased nested submodule imports

DavidSeptimus-Klotho:
This PR fixes an issue where a qualified module name is truncated for aliased nested submodule imports by changing a check for a non-nil module node (not populated for these imports to a non-empty `moduleName` string, which is set regardless of the import type by looking at multiple potential sources for a module's name.

### Standard checks

### DIFF
--- a/pkg/lang/python/imports.go
+++ b/pkg/lang/python/imports.go
@@ -119,12 +119,10 @@ func FindImports(file *core.SourceFile) Imports {
 		qualifiedModuleName := parent
 		if qualifiedModuleName == "" {
 			qualifiedModuleName = moduleName
-		} else if module != nil {
-			if qualifiedModuleName != "" && qualifiedModuleName[len(qualifiedModuleName)-1] == '.' {
-				qualifiedModuleName += moduleName
-			} else {
-				qualifiedModuleName += "." + moduleName
-			}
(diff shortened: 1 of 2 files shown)


pkg/lang/python/imports.go
```diff
@@ -119,12 +119,10 @@ func FindImports(file *core.SourceFile) Imports {
 		qualifiedModuleName := parent
 		if qualifiedModuleName == "" {
 			qualifiedModuleName = moduleName
-		} else if module != nil {
-			if qualifiedModuleName != "" && qualifiedModuleName[len(qualifiedModuleName)-1] == '.' {
-				qualifiedModuleName += moduleName
-			} else {
-				qualifiedModuleName += "." + moduleName
-			}
+		} else if qualifiedModuleName[len(qualifiedModuleName)-1] == '.' {
```
gordon-klotho, line 122: nit: use `strings.HasSuffix`

gordon-klotho (approved):

pkg/lang/python/imports.go
DavidSeptimus-Klotho: yeah, old code here, but changed.

PR #406, opened 2023-03-27
Log group template

jhsinger-klotho:

render dependencies fully.

This makes the dependsOn a keyword in our Args within each IaC Tempalte. If detected we use that to grab all varnames to resources upstream of the resource we are rendering

### Standard checks


pkg/infra/iac2/templates_compiler.go
```diff
@@ -26,7 +26,7 @@ type (
 		// templates is the fs.FS where we read all of our `<struct>/factory.ts` files
 		templates fs.FS
 		// resourceGraph is the graph of resources to render
-		resourceGraph *graph.Directed[core.Resource] // TODO make this be a core.ResourceGraph, and un-expose that struct's Underlying
+		resourceGraph *core.ResourceGraph // TODO make this be a core.ResourceGraph, and un-expose that struct's Underlying
```
yuval-klotho, line 29: should rm the TODO comment now:

```suggestion
		resourceGraph *core.ResourceGraph
```

pkg/infra/iac2/templates_compiler.go
```diff
@@ -147,6 +147,11 @@ func (tc templatesCompiler) renderResource(out io.Writer, resource core.Resource
 	inputArgs := make(map[string]string)
 	var zeroValue reflect.Value
 	for fieldName := range tmpl.inputTypes {
+		// dependsOn will be a reserved field for us to use to map dependencies. If specified as an Arg we will automatically call resolveDependencies
+		if fieldName == "dependsOn" {
```
yuval-klotho, line 151: I'd moderately prefer if we did it based on type, not fieldName — just feels a bit easier to track through the code. That is, create in go a `type Dependencies struct{}`, and then check here:

```go
if _, ok := childVal.Interface().(core.Dependencies); ok {
```

Not a blocker.

pkg/infra/iac2/templates_compiler_test.go
```diff
@@ -21,13 +20,13 @@ func TestOutputBody(t *testing.T) {
...
-	graph.AddEdge(fizz.Id(), parent.Id())
-	graph.AddEdge(buzz.Id(), parent.Id())
-	graph.AddEdge(buzz.Id(), fizz.Id())
+	graph := core.NewResourceGraph()
+	graph.AddResource(fizz)
+	graph.AddResource(buzz)
+	graph.AddResource(parent)
+	graph.AddDependency(fizz, parent)
+	graph.AddDependency(buzz, parent)
+	graph.AddDependency(buzz, fizz)
```
yuval-klotho, line 29: Shoot, I forgot to add a comment on this line. Since you're touching it anyway, mind adding this?

```suggestion
	graph.AddDependency(buzz, fizz) // provides a stable topological ordering
```

yuval-klotho (commented):
lgtm, only real issue is to rm the comment for the TODO you TODID. I also left a comment on the dependsOn, but that's not a blocker — dealer's choice.

pkg/infra/iac2/templates_compiler.go
jhsinger-klotho: but we wont have that struct used within the actual IR structs so we would never match to it. Its odd to have a struct created and appended to every single IR module if we need dependencies. These are also represented in the graph which makes it weird. Open to another way of doing this though, but it seems to me like the logic should reside completely in the IaC plugin since we already map dependencies in the graph

yuval-klotho: Oh, yeah good point! Alright, ignore me. :)

pkg/infra/iac2/templates/lambda_function/factory.ts
```diff
@@ -9,6 +9,7 @@ interface Args {
     Role: aws.iam.Role
     // envVars: Record<string, pulumi.Output<string>>,
     // dependsOn: []
+    dependsOn: any[]
```
gordon-klotho, line 12: Technically, this is `dependsOn?: pulumi.Input<pulumi.Input<pulumi.Resource>[]> | pulumi.Input<pulumi.Resource>`

pkg/infra/iac2/templates_compiler.go
```diff
@@ -176,6 +181,23 @@ func (tc templatesCompiler) renderResource(out io.Writer, resource core.Resource
...
+// resolveDependencies creates a string which models an array containing all the variable names, which the resource depends on.
+func (tc templatesCompiler) resolveDependencies(resource core.Resource) string {
+	buf := strings.Builder{}
+	buf.WriteRune('[')
+	upstreamResources := tc.resourceGraph.GetUpstreamResources(resource)
+	numDeps := len(upstreamResources)
+	for i := 0; i < numDeps; i++ {
+		res := upstreamResources[i]
+		buf.WriteString(tc.getVarName(res))
+		if i < (numDeps - 1) {
```
gordon-klotho, line 193: Pretty sure TS is okay with trailing commas

gordon-klotho (approved):

pkg/infra/iac2/templates/lambda_function/factory.ts
jhsinger-klotho: yeah let me go update this

PR #407, opened 2023-03-27
documentation and cleanup

yuval-klotho:
Most of this is just adding godocs. As part of that, expose some fields
(so that godoc picks them up) and move iac_templates.go -> resource_creation_template.go.

### Standard checks

- **Unit tests**: n/a
- **Docs**: adding 'em


pkg/infra/iac2/cloud_resources_stub.go
yuval-klotho, line 1: This file was just a stub (as the name suggests), and we don't need it anymore.

PR #433, opened 2023-03-29
Add node module for templates to make development easier

gordon-klotho:
```sh
npm --prefix pkg/infra/iac2/templates install
```
Then working with individual template folders can utilize the shared node modules.

### Standard checks


pkg/infra/iac2/templates/package-lock.json
yuval-klotho, line 1: need to rm this file (I forget if Jordan's PR to add it to gitignore is already in)

yuval-klotho (changes requested):

yuval-klotho (approved):

PR #434, opened 2023-03-29
Iam policy gen and render document tag

jhsinger-klotho:

This PR includes
- Policy Generator: A Centralized utility to track unit's roles and corresponding policies
- Currently it attaches all policies to roles inline, but policy support is added so we can change that and do managed policies
- the `render: "document"` flag on structs. This is used if your struct contains a field which is not a core.IaCValue or a core.Resource and will then output the value as a native js object for the key: value pairs. (Will add to our internal docs for this)

### Standard checks


pkg/infra/iac2/templates_compiler.go
```diff
@@ -188,78 +196,133 @@ func (tc TemplatesCompiler) resolveDependencies(resource core.Resource) string {
...
 		} else if typedChild, ok := childVal.Interface().(core.IaCValue); ok {
-			return tc.handleIaCValue(typedChild)
+			output, err := tc.handleIaCValue(typedChild)
+			if err != nil {
+				return output, err
+			}
+			return output, nil
 		} else {
-			return ""
+			if iacTag == "document" {
```
yuval-klotho, line 225: I think this should be the first check in the if-else chain, not near the last.

If my struct tags a field as `resource:"document"`, then I'd expect that field to get rendered as a document no matter what, even if it also happens to implement `core.Resource` or `core.IaCValue`.

pkg/provider/aws/persist.go
```diff
@@ -0,0 +1,25 @@
...
+
+func (a *AWS) GenerateFsResources(construct *core.Fs, result *core.ConstructGraph, dag *core.ResourceGraph) error {
+	accountId := resources.NewAccountId()
+	bucket := s3.NewS3Bucket(construct, a.Config.AppName, accountId)
+	upstreamResources := result.GetUpstreamConstructs(construct)
+	for _, res := range upstreamResources {
+		unit, ok := res.(*core.ExecutionUnit)
+		if ok {
+			a.PolicyGenerator.AddAllowPolicyToUnit(unit.Id(), []string{"s3:*"},
+				[]core.IaCValue{{Resource: bucket, Property: core.ARN_IAC_VALUE}, {Resource: bucket, Property: core.ALL_BUCKET_DIRECTORY_IAC_VALUE}})
```
yuval-klotho, line 17: minor: this formatting is a bit hard to read, imo

pkg/provider/aws/resources/iam/policy_generator.go
```diff
@@ -0,0 +1,78 @@
...
+			Effect:   "Allow",
+			Action:   actions,
+			Resource: resources,
+		})
+	}
+}
+
+func (p *PolicyGenerator) AddUnitRole(unitId string, role *IamRole) error {
+	_, found := p.unitToRole[unitId]
+	if found {
```
yuval-klotho, line 63: minor:
```suggestion
	if _, found := p.unitToRole[unitId]; found {
```

yuval-klotho (changes requested):

pkg/infra/iac2/templates_compiler.go
jhsinger-klotho: that doesnt really make a ton of sense to me because if it is a Resource or an IaCValue we would completely break the iac by not returning what they do today, since in our minds they arent a nested document and instead they are something we know how to handle.

I see your point on that we would expect it to get handled, but we may need to think about that here.

Also then we have to look at each child value to see if it has the tag, where as now we only require having a top level tag because we know were rendering everything below as a child resource

pkg/infra/iac2/templates_compiler.go
```diff
@@ -188,78 +197,133 @@ func (tc TemplatesCompiler) resolveDependencies(resource core.Resource) string {
...
+				output := "{"
+				correspondingStruct := childVal
+				if childVal.Kind() == reflect.Pointer {
```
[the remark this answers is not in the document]
jhsinger-klotho: will change this

pkg/infra/iac2/templates_compiler.go
```diff
@@ -188,78 +197,133 @@ func (tc TemplatesCompiler) resolveDependencies(resource core.Resource) string {
...
+					correspondingStruct = childVal.Elem()
+				}
+				for i := 0; i < correspondingStruct.NumField(); i++ {
+
+					anotherChildVal := correspondingStruct.Field(i)
```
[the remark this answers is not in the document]
jhsinger-klotho: yeah completely forgot about this naming

pkg/infra/iac2/templates_compiler.go
```diff
@@ -188,78 +197,133 @@ func (tc TemplatesCompiler) resolveDependencies(resource core.Resource) string {
...
+					fieldName := correspondingStruct.Type().Field(i).Name
+					resolvedValue, err := tc.resolveStructInput(anotherChildVal, false, iacTag)
+
+					if err != nil {
+						return output, err
+					}
+					if resolvedValue == "" {
+						return output, errors.Errorf(`child struct of %v is not of a known type: %v`, childVal.Type().Name(), fieldName)
```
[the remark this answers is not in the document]
jhsinger-klotho: so the way to change this if you dont want to handle both is returning the error on structs we dont know how to parse. Otherwise we would never catch it for arrays and maps since they dont return `""` to our top level function and thats the only thing that signals an error today. I can go ahead and do that though

pkg/infra/iac2/templates_compiler_test.go
```diff
@@ -48,7 +49,11 @@ func TestOutputBody(t *testing.T) {
 			"",
 			"const bigMain = new DummyParent(",
 			"				fizzMyHello,",
-			"				{buzz: buzzShared});")
+			"				{",
+			"					buzz: buzzShared,",
+			"					nest: {",
+			"Fizz: fizzMyHello,}",
```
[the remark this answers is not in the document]
jhsinger-klotho: yeah we can work on the formatting later on i think, just want to get this working to unblock us

pkg/infra/iac2/templates_compiler.go
yuval-klotho: That all makes sense, and given that we don't actually need `core.Resource`s-as-documents today I'm fine with it. But then maybe just add a check above, and error out in case somebody does do it? Otherwise, I could see this being a pretty lengthy and frustrating debugging exercise in the future, if we do ever do have a `core.Resource`-as-document for some reason.

jhsinger-klotho: Fair, ill go ahead and add the check for our supported resources then

yuval-klotho (approved):

PR #435, opened 2023-03-29
replace deprecated LoadSyntax

yuval-klotho:
Also make sub-tests for each reflectively-found resource type; it makes it easier to verify that we actually found types.

### Standard checks

- **Unit tests**: n/a
- **Docs**: n/a


yuval-klotho:
closing in favor of #437

PR #441, opened 2023-03-30
ifc2: secrets

yuval-klotho:
Putting up a draft PR. The `ConstructIdToResourceId` is TODO, but everything else I think should be good?

### Standard checks

- **Unit tests**: no special tests added
- **Docs**: n/a

### DIFF
--- a/pkg/infra/iac2/templates/secret_version/factory.ts
+++ b/pkg/infra/iac2/templates/secret_version/factory.ts
@@ -1,0 +1,25 @@
+import * as aws from '@pulumi/aws'
+import * as fs from 'fs'
+
+interface Args {
+    SecretName: string
+    Secret: aws.secretsmanager.Secret
+    Path: string
+}
+
+// noinspection JSUnusedLocalSymbols
+function create(args: Args): void {
+    return fs.readFile(args.Path, (err, data) => {
+        if (err != null) {
+            return
+        }
+        new aws.secretsmanager.SecretVersion(
+            args.SecretName,
+            {
+                secretId: args.Secret.id,
+                secretBinary: data.toString('base64'),
+            },
+            { protect: this.lib.protect }
+        )
+    })
+}
--- a/pkg/provider/aws/secret.go
+++ b/pkg/provider/aws/secret.go
@@ -1,0 +1,19 @@
+package aws
+
+import (
+	"github.com/klothoplatform/klotho/pkg/core"
+	"github.com/klothoplatform/klotho/pkg/provider/aws/resources"
+)
+
+func (a *AWS) GenerateSecretsResources(construct *core.Secrets, result *core.ConstructGraph, dag *core.ResourceGraph) error {
+	for _, single := range construct.Secrets {
+		secret := resources.NewSecret(construct.Provenance(), single, a.Config.AppName)
+		dag.AddResource(secret)
+		a.MapResourceDirectlyToConstruct(secret, construct)
+
+		secretVersion := resources.NewSecretVersion(secret, single)
+		dag.AddResource(secret)
+		dag.AddDependency2(secret, secretVersion)
+	}
+	return nil
+}
--- a/pkg/provider/aws/secret.go
+++ b/pkg/provider/aws/secret.go
@@ -12,7 +12,7 @@ func (a *AWS) GenerateSecretsResources(construct *core.Secrets, result *core.Con
 		a.MapResourceDirectlyToConstruct(secret, construct)
 
 		secretVersion := resources.NewSecretVersion(secret, single)
-		dag.AddResource(secret)
+		dag.AddResource(secretVersion)
 		dag.AddDependency2(secret, secretVersion)
 	}
 	return nil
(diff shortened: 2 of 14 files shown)


pkg/core/resource_graph.go
```diff
@@ -20,8 +20,8 @@ func (cg *ResourceGraph) AddResource(resource Resource) {
 	cg.underlying.AddVertex(resource)
 }
 
-func (cg *ResourceGraph) AddDependency(source Resource, dest Resource) {
-	cg.underlying.AddEdge(source.Id(), dest.Id())
+func (cg *ResourceGraph) AddDependency(thingThatIsDependedOn Resource, thingThatDepends Resource) {
```
jhsinger-klotho, line 23: while these are a bit clearer its still a bit hard on the eyes lol

pkg/infra/iac2/templates/secret_version/factory.ts
```diff
@@ -0,0 +1,25 @@
...
+        }
+        new aws.secretsmanager.SecretVersion(
+            args.SecretName,
+            {
+                secretId: args.Secret.id,
+                secretBinary: data.toString('base64'),
+            },
+            { protect: this.lib.protect }
+        )
+    })
```
jhsinger-klotho, line 24: does anything depend on this in deployment order? Because the callback will be invoked async and we wont have the actual secretversion to reference since it wont be assigned to a constant

jhsinger-klotho (commented):
yeah this looks like its on the right track, just need to map the dependency to the lambda like you mentioned in the todo

pkg/core/resource_graph.go
yuval-klotho: I'll revert this, since it conflicts anyway with #449

pkg/infra/iac2/templates/secret_version/factory.ts
yuval-klotho: no — this depends on Secret, but nothing depends on this. And if you look at the return signature, it's void anyway.

Which btw will cause a bit of a weird (but valid) ts right now, because we'll actually assign a variable to it. Maybe I should add an exception that if the return type is void, don't assign it a var.

yuval-klotho: > Maybe I should add an exception that if the return type is void, don't assign it a var.

Filed as #450. I put it in milestone 4, but it's easy, so I may just do it one afternoon.

pkg/provider/aws/secret.go
```diff
@@ -0,0 +1,19 @@
...
+	for _, single := range construct.Secrets {
+		secret := resources.NewSecret(construct.Provenance(), single, a.Config.AppName)
+		dag.AddResource(secret)
+		a.MapResourceDirectlyToConstruct(secret, construct)
+
+		secretVersion := resources.NewSecretVersion(secret, single)
+		dag.AddResource(secret)
+		dag.AddDependency2(secret, secretVersion)
+	}
+	return nil
```
jhsinger-klotho, line 18: can we add a happy path unit test for this just to make sure the dag contains the correct resources and dependencies?

jhsinger-klotho (commented):
one tiny comment

pkg/provider/aws/secret.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/klothoplatform/klotho/pkg/core"
+	"github.com/klothoplatform/klotho/pkg/provider/aws/resources"
+)
+
+func (a *AWS) GenerateSecretsResources(construct *core.Secrets, result *core.ConstructGraph, dag *core.ResourceGraph) error {
+	for _, single := range construct.Secrets {
+		secret := resources.NewSecret(construct.Provenance(), single, a.Config.AppName)
+		dag.AddResource(secret)
+		a.MapResourceDirectlyToConstruct(secret, construct)
+
```
jhsinger-klotho, line 13: also does the execution units resource need to depend on the secret version? Because its an env var it likely does

yuval-klotho: `MapResourceDirectlyToConstruct` will make the exec unit depend on the secret, right? And the secret depends on the secret-version, so the exec unit will transitively depend on the secret-version. Does it need to be a direct dependency for env vars?

jhsinger-klotho: i guess it indirectly does because it will map as a dependency to the functions role and the function is dependent on that, so that may be fine yeah

pkg/provider/aws/secret.go
```diff
@@ -0,0 +1,19 @@
...
+		secretVersion := resources.NewSecretVersion(secret, single)
```
jhsinger-klotho, line 14: Dont we need to create a policy to access the secret version as well?

yuval-klotho: ah yeah, thanks for the reminder. (When I first wrote this, the policy stuff wasn't in yet, I think)

jhsinger-klotho: 👍 yeah other than this, looks good to me

jhsinger-klotho (approved):

PR #443, opened 2023-03-31
Adds support for `render:"template"` tags on a resource's struct fields

DavidSeptimus-Klotho:
- Added parseTS resource template function and injected it into all args usages in templates
- Wrapped all resource template values in structs implementing templateValue
- Added support for rendering nested structs in resource as standard go templates when the field is annotated with `render:"template"`
- Raw, unparsed, template values are accessible by invoking .Raw on resource template args

`render:"template` template paths are expected to be in the lower-snake-case form of: `{resource_struct_type}/{nested_struct_type}.ts[.tmpl]`

### Standard checks

### DIFF
--- a/pkg/infra/iac2/iac2.go
+++ b/pkg/infra/iac2/iac2.go
@@ -44,6 +44,30 @@
 //     factory.ts as a "template for a template". and this field as the template derived from that file.
 //  4. Imports, a set of string representing the TypeScript imports. Again, these are taken literally from the source.
 //
+// Template input values in a factory.ts file are wrapped in structs implementing the templateValue interface.
+//
+// To access an input's raw values in the factory template, invoke its `Raw` function.
+//
+// For Example:
+//
+//	const tableName = {{ .Table.Raw.Name }};
+//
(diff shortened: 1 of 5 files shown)


jhsinger-klotho (approved):
LGTM

yuval-klotho:
The changes are good. But now that we're getting a bit more complicated in terms of the API, could you add docs in `iac2.go` (or wherever) to explain how the tags get used? Probably in the `# ResourceCreationTemplate` section.

yuval-klotho (approved):

jhsinger-klotho (approved):

pkg/infra/iac2/iac2.go
```diff
@@ -44,6 +44,30 @@
...
+//
+// For Example:
+//
+//	const table = {{ renderTS .Table }}
+//
+// # Nested Struct Rendering
+//
+// Within a Resource struct, the default TemplatesCompiler rendering behavior allows only fields of primitive types,
+// slices, maps, and structs implementing the Resource or IacValue interfaces.
+// This behavior can be modified with the following using the following tas on struct fields:
```
yuval-klotho, line 65: This is perfect, thanks!

PR #447, opened 2023-03-31
start of eks

### LINKED ISSUE #416: EKS Setup
ability to spin up a cluster (For now just create a cluster and one node group and one fargate profile for the default namespace, similar to the one in deploylib today (this should catch fargate exec units))

Since we are assuming people are coming with their own cluster, lets do the minimal setup for now.

the definition of done:

jhsinger-klotho:
closes #416 

Adds setup for eks:
- Eks clusters (they will be created based off the units ClusterId field in config. If not set well use the first alphabetical clusterId)
- Node Groups
- Fargate Profiles (with a single selector in the default namespace that matches our label)

small change to vpc.go. CreateNetwork will return a vpc and see if the static vpc is added to the graph it wont try to create more resources

Also adds back the code to output the config file into the compiled dir

### Standard checks


pkg/compiler/compiler.go
```diff
@@ -88,9 +94,47 @@ func (c *Compiler) Compile() error {
...
+		return errors.Wrap(err, "Unable to output Klotho configuration file")
+	}
 	return c.Document.OutputTo(c.Document.Configuration.OutDir)
 }
 
+func (c *Compiler) createConfigOutputFile() error {
+	c.Document.Configuration.UpdateForResources(c.Document.Constructs.ListConstructs())
+	buf := new(bytes.Buffer)
+	var err error
+	var ext string
```
yuval-klotho, line 108: Minor: we don't really need this var, since it's always just `c.Document.Configuration.Format` for any case where we don't return an error.

pkg/infra/iac2/templates/eks_node_group/factory.ts
```diff
@@ -0,0 +1,37 @@
...
+    InstanceTypes: string[]
+    Labels: Record<string, string>
+}
+
+// noinspection JSUnusedLocalSymbols
+function create(args: Args): aws.eks.NodeGroup {
+    return new aws.eks.NodeGroup(args.Name, {
+        clusterName: args.Cluster.name,
+        nodeRoleArn: args.NodeRole.arn,
+        amiType: args.AmiType != '' ? args.AmiType : undefined,
```
yuval-klotho, line 23: It's too bad this go-ism (`""` being the null string) leaks into the output ts. I forget, do we have the ability to embed go-template rules directly into these ts files? If so, this could be a good place to add `/* TMPL if {{.AmiType}} */`. Otherwise, would you mind filing a ticket and adding it to the fit-and-finish milestone?

pkg/infra/iac2/templates_compiler.go
```diff
@@ -75,9 +75,6 @@ func (s stringTemplateValue) Raw() interface{} {
 
 func (v nestedTemplateValue) Parse() (string, error) {
 	childVal := v.childValue
-	if childVal.Kind() == reflect.Pointer && !childVal.IsNil() {
```
yuval-klotho, line 78: Totally out of curiosity, what problem was this causing?

pkg/provider/aws/execution_unit.go
```diff
@@ -50,19 +50,25 @@ func (a *AWS) GenerateExecUnitResources(unit *core.ExecutionUnit, result *core.C
...
-// convertExecUnitParams transforms the execution units environment variables to a map of key names and their corresponding core.IaCValue struct
+// convertExecUnitParams transforms the execution units environment variables to a map of key names and their corresponding core.IaCValue struct.
 //
-// If an environment variable does not pertain to a construct and is just a key, value string, the resource of the IaCValue will be left null
+// If an environment variable does not pertain to a construct and is just a key, value string, the resource of the IaCValue will be left null.
 func (a *AWS) convertExecUnitParams(result *core.ConstructGraph, dag *core.ResourceGraph) error {
 	execUnits := core.GetResourcesOfType[*core.ExecutionUnit](result)
 	for _, unit := range execUnits {
+
+		if a.Config.GetExecutionUnit(unit.ID).Type == Kubernetes {
```
yuval-klotho, line 69: Why is k8s being singled out here? (And could you add a comment about it?)

pkg/provider/aws/plugin.go
```diff
@@ -55,6 +59,49 @@ func (a *AWS) Translate(result *core.ConstructGraph, dag *core.ResourceGraph) (L
 	return
 }
 
+// createEksCluster determines whether any execution units have a type of EKS to determine whether a cluster needs to be created.
+//
+// If any units do have a type of EKS, the function will look at their configuration to determine how the cluster needs to be configured.
+// The clusterId field within an execution units configuration, will determine which cluster the unit will belong to, helping klotho understands how many clusters to create.
+// If the clusterId field is unassigned, the execution unit will be assigned to the first clusterId alphanumerically.
+// If there are no clusterIds defined by any units, one cluster will be created for all units.
```
yuval-klotho, line 67: Good comment!

pkg/provider/aws/plugin.go
```diff
@@ -55,6 +59,49 @@ func (a *AWS) Translate(result *core.ConstructGraph, dag *core.ResourceGraph) (L
...
+				unassignedUnits = append(unassignedUnits, unit)
+				continue
+			}
+			clusterIdToUnit[params.ClusterId] = append(clusterIdToUnit[params.ClusterId], unit)
+
+		}
+	}
+
+	//Assign unassigned units to the first key after sorted
+	keys := []string{}
```
yuval-klotho, line 85: I'm not sure how I feel about assigning it to the first one — that feels really arbitrary. If there's only one key, assigning the units to it makes sense. But otherwise, it feels like we should just error out.

pkg/provider/aws/plugin.go
```diff
@@ -55,6 +59,49 @@ func (a *AWS) Translate(result *core.ConstructGraph, dag *core.ResourceGraph) (L
...
+	for k := range clusterIdToUnit {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+
+	if len(keys) == 0 && len(unassignedUnits) != 0 {
+		keys = append(keys, "eks-cluster")
```
yuval-klotho, line 92: This is an awkward way of accomplishing it — it took me a bit to realize what you were doing.

Suggestion: two separate `if`s: one to append to `keys` if it's empty (regardless of `unassignedUnits`), and one to assign the `unassignedUnits` (in which you can assume `keys[0]` exists).

pkg/provider/aws/plugin_test.go
```diff
@@ -1 +1,142 @@
...
+
+func Test_createEksClusters(t *testing.T) {
+	cases := []struct {
+		name   string
+		units  []*core.ExecutionUnit
+		config *config.Application
+		want   []*resources.EksCluster
+	}{
+		{
+			name: `on exec unit, no cluster id`,
```
yuval-klotho, line 22:
```suggestion
			name: `one exec unit, no cluster id`,
```

pkg/provider/aws/resources/eks.go
```diff
@@ -0,0 +1,212 @@
...
+	}}
+	return fargateRole
+}
+
+func createNodeRole(appName string, roleName string, refs []core.AnnotationKey) *IamRole {
+	nodeRole := NewIamRole(appName, roleName, refs, EC2_ASSUMER_ROLE_POLICY)
+	nodeRole.AddAwsManagedPolicies([]string{"arn:aws:iam::aws:policy/AmazonEKSWorkerNodePolicy",
+		"arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryReadOnly",
+		"arn:aws:iam::aws:policy/AmazonEKS_CNI_Policy", "arn:aws:iam::aws:policy/AWSCloudMapFullAccess",
+		"arn:aws:iam::aws:policy/CloudWatchAgentServerPolicy"})
```
yuval-klotho, line 131: minor: I would add a newline after the opening `[]string{`, and then again after the last element (`CloudWatchAgentServerPolicy`, and also adding a comma after it as required by Go style). That way, if we add or remove items in the future, the diff will be cleaner.

pkg/provider/aws/resources/eks_test.go
```diff
@@ -0,0 +1,103 @@
...
+		deps  []stringDep
+	}
+	cases := []struct {
+		name string
+		want testResult
+	}{
+		{
+			name: "happy path",
+			want: testResult{
+				nodes: []string{"aws:eks_cluster:test-app-test-cluster",
```
yuval-klotho, line 31: similar comment to above (re the newline)

pkg/provider/aws/resources/eks_test.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	for _, tt := range cases {
+		t.Run(tt.name, func(t *testing.T) {
+			assert := assert.New(t)
+			dag := core.NewResourceGraph()
+			CreateEksCluster(appName, clusterName, []*Subnet{subnet}, nil, eus, dag)
+			for _, id := range tt.want.nodes {
+				resource := dag.GetResource(id)
+				assert.NotNil(resource, fmt.Sprintf("Resource %s, not found", id))
+				assert.Equal(resource.KlothoConstructRef(), []core.AnnotationKey{{ID: "test"}}, fmt.Sprintf("Resource %s, did not recieve klotho construct ref", resource.Id()))
```
yuval-klotho, line 56: minor: I don't think you need the `fmt.Sprintf` — you can just write the format string and `resource.Id()` directly. (You _may_ need to turn this into `Equalf` for that to work, though as far as I can tell they seem to be effectively the same?)

pkg/infra/iac2/templates/eks_node_group/factory.ts
jhsinger-klotho: yeah we do, i can add the templating here. We also want to port over the ami types so we support gpus which we dont have a ticket for so i will create that

pkg/provider/aws/execution_unit.go
jhsinger-klotho: because we actually put the environment variables in the chart as values already so all we need to do there is inject them with real values. At this point we arent actually creating any resource for eks exec units until we get the helm charts up and running so this will likely change, we just dont have it figured out yet.

If you still want me to comment i can but i figured that this will end up being different soon

pkg/provider/aws/plugin.go
jhsinger-klotho: hmmm yeah maybe thats the right method

pkg/provider/aws/resources/eks_test.go
jhsinger-klotho: cool didnt know equalF was a thing

jhsinger-klotho (commented):
should have addressed all of your comments, let me know

yuval-klotho (approved):

PR #449, opened 2023-03-31
Synchronize dependency order with visualizer and IaC references

gordon-klotho:
For visualizer, we want `unit -> role` directions
For IaC, unit references role, so also `unit -> role`
If we represent the graph internally like this, then the IaC should create variable for resources in the graph in inverse order (role first, then unit).

### Standard checks


jhsinger-klotho:
For IaC, unit references role, so also unit -> role

I believe this is opposite in terms of deployment order. we dont actually care about the reference in terms of the graph today in our iac plugin but this could change

jhsinger-klotho:
also if we are reversing the order of our resourceGraph wouldnt we want to do the same for our ConstructGraph?

jhsinger-klotho (approved):

yuval-klotho (changes requested):
This is going to break any outstanding PRs (like mine, selfishly) in ways that won't be caught at compile time.

Can we rename `ResourceGraph.AddDependency` to something else at least in the near term, so that anyone who has outstanding code will get a nice compiler error?

pkg/provider/aws/resources/vpc.go
```diff
@@ -76,7 +76,7 @@ func CreateNetwork(appName string, dag *core.ResourceGraph) {
 	dag.AddResource(region)
 	dag.AddResource(vpc)
 	dag.AddResource(igw)
-	dag.AddDependency(vpc, igw)
+	dag.AddDependency(igw, vpc)
```
jhsinger-klotho, line 79: how are the unit tests for createNetwork still passing also? They check for deps

gordon-klotho: The test was actually ineffectual since `GetDependency` would never return nil. This is updated (to return nil) and the test changed in the recent commit.

yuval-klotho (approved):

PR #452, opened 2023-03-31
VPC is now downstream of Subnets

gordon-klotho:
Fixes bug introduced in https://github.com/klothoplatform/klotho/pull/449

### Standard checks


pkg/provider/aws/resources/vpc.go
```diff
@@ -133,7 +135,7 @@ func CreateInterfaceVpcEndpoint(service string, vpc *Vpc, region *Region, dag *c
 
 func (vpc *Vpc) GetVpcSubnets(dag *core.ResourceGraph) []*Subnet {
 	subnets := []*Subnet{}
-	downstreamDeps := dag.GetDownstreamResources(vpc)
+	downstreamDeps := dag.GetUpstreamResources(vpc)
```
gordon-klotho, line 138: This is the bug fix, the rest is updating the unit tests to have caught this bug.

jhsinger-klotho (approved):

PR #453, opened 2023-03-31
Add elasticache and security group support

gordon-klotho:
Running ts-redis sample app:
```ts
const awsElasticacheSubnetgroupPersistRedis = new aws.elasticache.SubnetGroup(`persist:redis`, {
        subnetIds: [awsVpcSubnetTestAppPublic1,awsVpcSubnetTestAppPublic2,awsVpcSubnetTestAppPrivate2,awsVpcSubnetTestAppPrivate1].map((sg) => sg.id),
    });

const awsLogGroupGgDevAwselasticacheggDevPersistredisPersistRedis = new aws.cloudwatch.LogGroup(`gg_dev_awselasticachegg_dev_persistredis_persist_redis`, {
        name: `/aws/elasticache/gg-dev-persist:redis-persist-redis`,
        retentionInDays: 0,
    });

const awsSecurityGroupGgDev = new aws.ec2.SecurityGroup(`gg-dev`, {
        name: `gg-dev`,
        vpcId: awsVpcTestApp.id,
        egress: [
            {
                cidrBlocks: ['0.0.0.0/0'],
                description: 'Allows all outbound IPv4 traffic.',
                fromPort: 0,
                protocol: '-1',
                toPort: 0,
            },
        ],
        ingress: [
            {
                description:
                    'Allows inbound traffic from network interfaces and instances that are assigned to the same security group.',
                fromPort: 0,
                protocol: '-1',
                self: true,
                toPort: 0,
            },
            {
                description: 'For EKS control plane',
                cidrBlocks: ['0.0.0.0/0'],
                fromPort: 9443,
                protocol: 'TCP',
                toPort: 9443,
            },
            {
                description: 'For private subnets internally',
                cidrBlocks: [awsVpcTestApp.cidrBlock],
                fromPort: 0,
                protocol: '-1',
                toPort: 0,
            },
        ],
    });

const awsElasticachePersistRedis = new aws.elasticache.Cluster(`persist:redis`, {
        engine: 'redis',
        clusterId: `redis`,
        logDeliveryConfigurations: [
            {
                destination: awsLogGroupGgDevAwselasticacheggDevPersistredisPersistRedis.name,
                destinationType: 'cloudwatch-logs',
                logFormat: 'text',
                logType: 'slow-log',
            },
            {
                destination: awsLogGroupGgDevAwselasticacheggDevPersistredisPersistRedis.name,
                destinationType: 'cloudwatch-logs',
                logFormat: 'json',
                logType: 'engine-log',
            },
        ],
        subnetGroupName: awsElasticacheSubnetgroupPersistRedis.name,
        securityGroupIds: [awsSecurityGroupGgDev].map((sg) => sg.id),
    });
```

### Standard checks

### DIFF
--- a/pkg/infra/iac2/templates/security_group/factory.ts
+++ b/pkg/infra/iac2/templates/security_group/factory.ts
@@ -1,0 +1,46 @@
+import * as aws from '@pulumi/aws'
+
+interface Args {
+    Name: string
+    Vpc: aws.ec2.Vpc
+}
+
+function create(args: Args): aws.ec2.SecurityGroup {
+    return new aws.ec2.SecurityGroup(args.Name, {
+        name: args.Name,
+        vpcId: args.Vpc.id,
+        egress: [
+            {
+                cidrBlocks: ['0.0.0.0/0'],
+                description: 'Allows all outbound IPv4 traffic.',
+                fromPort: 0,
+                protocol: '-1',
+                toPort: 0,
+            },
+        ],
+        ingress: [
+            {
+                description:
+                    'Allows inbound traffic from network interfaces and instances that are assigned to the same security group.',
+                fromPort: 0,
+                protocol: '-1',
+                self: true,
+                toPort: 0,
+            },
+            {
+                description: 'For EKS control plane',
+                cidrBlocks: ['0.0.0.0/0'],
+                fromPort: 9443,
+                protocol: 'TCP',
+                toPort: 9443,
+            },
+            {
+                description: 'For private subnets internally',
+                cidrBlocks: [args.Vpc.cidrBlock],
+                fromPort: 0,
+                protocol: '-1',
+                toPort: 0,
+            },
+        ],
+    })
+}
--- a/pkg/infra/iac2/templates_compiler.go
+++ b/pkg/infra/iac2/templates_compiler.go
@@ -417,9 +417,19 @@ func (tc TemplatesCompiler) handleIaCValue(v core.IaCValue) (string, error) {
 	case string(core.ALL_RESOURCES_IAC_VALUE):
 		return "*", nil
 	case string(core.HOST):
-		return fmt.Sprintf("pulumi.interpolate`%s.cacheNodes[0].address`", tc.getVarName(v.Resource)), nil
+		switch v.Resource.(type) {
+		case *resources.Elasticache:
+			return fmt.Sprintf("pulumi.interpolate`%s.cacheNodes[0].address`", tc.getVarName(v.Resource)), nil
+		default:
+			return "", errors.Errorf("unsupported resource type %T for '%s'", v.Resource, v.Property)
+		}
 	case string(core.PORT):
-		return fmt.Sprintf("pulumi.interpolate`%s.cacheNodes[0].port`", tc.getVarName(v.Resource)), nil
+		switch v.Resource.(type) {
+		case *resources.Elasticache:
+			return fmt.Sprintf("pulumi.interpolate`%s.cacheNodes[0].port`", tc.getVarName(v.Resource)), nil
+		default:
+			return "", errors.Errorf("unsupported resource type %T for '%s'", v.Resource, v.Property)
+		}
 	}
 	return "", errors.Errorf("unsupported IaC Value Property, %s", v.Property)
 }
--- a/pkg/infra/iac2/templates_compiler.go
+++ b/pkg/infra/iac2/templates_compiler.go
@@ -416,6 +416,10 @@ func (tc TemplatesCompiler) handleIaCValue(v core.IaCValue) (string, error) {
 		return fmt.Sprintf("%s.invokeArn", tc.getVarName(v.Resource)), nil
 	case string(core.ALL_RESOURCES_IAC_VALUE):
 		return "*", nil
+	case string(core.HOST):
+		return fmt.Sprintf("pulumi.interpolate`%s.cacheNodes[0].address`", tc.getVarName(v.Resource)), nil
+	case string(core.PORT):
+		return fmt.Sprintf("pulumi.interpolate`%s.cacheNodes[0].port`", tc.getVarName(v.Resource)), nil
 	}
 	return "", errors.Errorf("unsupported IaC Value Property, %s", v.Property)
 }
--- a/pkg/provider/aws/persist.go
+++ b/pkg/provider/aws/persist.go
@@ -25,14 +25,6 @@ func (a *AWS) GenerateFsResources(construct *core.Fs, result *core.ConstructGrap
 
 func (a *AWS) GenerateRedisResources(construct core.Construct, result *core.ConstructGraph, dag *core.ResourceGraph) error {
 	ec := resources.CreateElasticache(a.Config, dag, construct)
-	upstreamResources := result.GetUpstreamConstructs(construct)
-	for _, res := range upstreamResources {
-		unit, ok := res.(*core.ExecutionUnit)
-		if ok {
-			unit.EnvironmentVariables.Add(core.GenerateRedisHostEnvVar(construct))
-			unit.EnvironmentVariables.Add(core.GenerateRedisPortEnvVar(construct))
-		}
-	}
 	a.MapResourceDirectlyToConstruct(ec, construct)
 	return nil
 }
--- a/pkg/provider/aws/persist.go
+++ b/pkg/provider/aws/persist.go
@@ -22,3 +22,17 @@ func (a *AWS) GenerateFsResources(construct *core.Fs, result *core.ConstructGrap
 	dag.AddResource(bucket)
 	return nil
 }
+
+func (a *AWS) GenerateRedisResources(construct core.Construct, result *core.ConstructGraph, dag *core.ResourceGraph) error {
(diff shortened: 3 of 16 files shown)


pkg/provider/aws/resources/vpc.go
```diff
@@ -78,6 +79,11 @@ func CreateNetwork(appName string, dag *core.ResourceGraph) *Vpc {
 	dag.AddResource(igw)
 	dag.AddDependency2(igw, vpc)
 
+	CreatePrivateSubnet(appName, "private1", vpc, "10.0.0.0/18", dag)
```
gordon-klotho, line 82: Since the endpoints use the subnets, these need to go first now.

pkg/infra/iac2/templates/security_group/factory.ts
```diff
@@ -0,0 +1,46 @@
...
+        ingress: [
+            {
+                description:
+                    'Allows inbound traffic from network interfaces and instances that are assigned to the same security group.',
+                fromPort: 0,
+                protocol: '-1',
+                self: true,
+                toPort: 0,
+            },
+            {
```
jhsinger-klotho, line 30: These should be defined in the provider and configurable. We dont need to add the eks control plane rule unless we have eks in the architecture

pkg/infra/iac2/templates_compiler.go
```diff
@@ -407,6 +407,10 @@ func (tc TemplatesCompiler) handleIaCValue(v core.IaCValue) (string, error) {
 		return fmt.Sprintf("%s.arn", tc.getVarName(v.Resource)), nil
 	case string(core.ALL_BUCKET_DIRECTORY_IAC_VALUE):
 		return fmt.Sprintf("pulumi.interpolate`${%s.arn}/*`", tc.getVarName(v.Resource)), nil
+	case string(core.HOST):
+		return fmt.Sprintf("pulumi.interpolate`%s.cacheNodes[0].address`", tc.getVarName(v.Resource)), nil
+	case string(core.PORT):
+		return fmt.Sprintf("pulumi.interpolate`%s.cacheNodes[0].port`", tc.getVarName(v.Resource)), nil
```
jhsinger-klotho, line 413: This needs to make sure that the resource is redis specific, because grabbing the host and port are going to be different pulumi code. Im thinking using the sub templating weve built may be better than these iac values now. Not a blocker for this pr but i think were going to want to rethink this

pkg/provider/aws/persist.go
```diff
@@ -22,3 +22,17 @@ func (a *AWS) GenerateFsResources(construct *core.Fs, result *core.ConstructGrap
...
 }
+
+func (a *AWS) GenerateRedisResources(construct core.Construct, result *core.ConstructGraph, dag *core.ResourceGraph) error {
+	ec := resources.CreateElasticache(a.Config, dag, construct)
+	upstreamResources := result.GetUpstreamConstructs(construct)
+	for _, res := range upstreamResources {
+		unit, ok := res.(*core.ExecutionUnit)
+		if ok {
+			unit.EnvironmentVariables.Add(core.GenerateRedisHostEnvVar(construct))
+			unit.EnvironmentVariables.Add(core.GenerateRedisPortEnvVar(construct))
```
jhsinger-klotho, line 33: why are we adding env vars to a construct here? The provider shouldnt need to do that, it should already be done within the plugin itself

pkg/provider/aws/resources/elasticache.go
```diff
@@ -0,0 +1,91 @@
...
+func (ecsn *ElasticacheSubnetgroup) KlothoConstructRef() []core.AnnotationKey {
+	return ecsn.ConstructsRef
+}
+
+// ID returns the id of the cloud resource
+func (ecsn *ElasticacheSubnetgroup) Id() string {
+	return fmt.Sprintf("%s:%s:%s", ecsn.Provider(), ECSN_TYPE, ecsn.Name)
+}
+
+func CreateElasticache(cfg *config.Application, dag *core.ResourceGraph, source core.Construct) *Elasticache {
```
jhsinger-klotho, line 62: i would take in a list of constructs because multiple could depend on this and then we can set the reference to all of them. Something i started to realize with creating new resources

pkg/provider/aws/resources/securitygroup.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"github.com/klothoplatform/klotho/pkg/config"
+	"github.com/klothoplatform/klotho/pkg/core"
+)
+
+type SecurityGroup struct {
+	Name          string
+	Vpc           *Vpc
+	ConstructsRef []core.AnnotationKey
+}
```
jhsinger-klotho, line 14: Should have Ingress and Egress rules here so we can add to them and modify dynamically

pkg/provider/aws/resources/vpc.go
jhsinger-klotho: 👍 i had this in my pr as well, but i can wait for you to get this in and update my eks one

pkg/provider/aws/resources/vpc.go
```diff
@@ -86,15 +92,25 @@ func CreateNetwork(appName string, dag *core.ResourceGraph) *Vpc {
...
+			return vpc
+		}
+	}
+	return CreateNetwork(cfg.AppName, dag)
+}
+
+func GetSubnets(cfg *config.Application, dag *core.ResourceGraph) (sns []*Subnet) {
+	vpc := GetVpc(cfg, dag)
+	return vpc.GetVpcSubnets(dag)
+}
```
jhsinger-klotho, line 110: i believe we have a GetVPCSbnets already which i think is more applicable because if we ever support multiple VPCs this wont work.

pkg/provider/aws/resources/securitygroup.go
```diff
@@ -0,0 +1,46 @@
...
+	for _, r := range dag.ListResources() {
+		if sg, ok := r.(*SecurityGroup); ok {
+			return sg
+		}
+	}
+	sg := &SecurityGroup{
+		Name: cfg.AppName,
+		Vpc:  GetVpc(cfg, dag),
+	}
+	dag.AddResource(sg)
```
jhsinger-klotho, line 29: this has to have a dependency on the vpc and subnets likely because were going to create this all at one rather than doing the add rules and since we use the subnet cidr blocks we need the iac to come after those

pkg/provider/aws/resources/securitygroup.go
gordon-klotho: Would you be fine with leaving that for a future change when we want to make them dynamic?

jhsinger-klotho: Sure although i think we do it as a fast follow because as we do the eks work it being dynamic will be beneficial

pkg/provider/aws/resources/elasticache.go
```diff
@@ -0,0 +1,91 @@
...
+		sg.ConstructsRef = append(sg.ConstructsRef, source.Provenance())
+		dag.AddDependency2(ec, sg)
+	}
+	for _, sn := range ec.SubnetGroup.Subnets {
+		sn.ConstructsRef = append(sn.ConstructsRef, source.Provenance())
+		dag.AddDependency2(ec.SubnetGroup, sn)
+	}
+
+	return ec
+}
```
jhsinger-klotho, line 91: can we unit test this to make sure it creates the correct resources and dependencies since it directly interacts with the dag?

pkg/provider/aws/resources/elasticache.go
```diff
@@ -0,0 +1,91 @@
...
+	ec := &Elasticache{
+		Name:            source.Id(),
+		Engine:          "redis", // TODO determine this from the type of `source`
+		CloudwatchGroup: NewLogGroup(cfg.AppName, fmt.Sprintf("/aws/elasticache/%s-%s-persist-redis", cfg.AppName, source.Id()), source.Provenance(), 0),
+		SubnetGroup: &ElasticacheSubnetgroup{
+			Name:          source.Id(),
+			Subnets:       GetSubnets(cfg, dag),
+			ConstructsRef: []core.AnnotationKey{source.Provenance()},
+		},
+		SecurityGroups: []*SecurityGroup{GetSecurityGroup(cfg, dag)},
```
jhsinger-klotho, line 72: I still dont love the GetSubnets and GetSecurityGroup because its not specific to the network and boxes us in to only being able to support a single vpc.

I wont consider this blocking for now but i do think we should rewrite this logic

pkg/provider/aws/resources/securitygroup.go
jhsinger-klotho: Can you create a task for this? Becuase the eks cluster will definitely need to add its own rule to the SG

jhsinger-klotho (approved):
just some tasks to create or TODOs to mention

pkg/provider/aws/resources/elasticache.go
gordon-klotho: We don't know how multiple networks will work, so it's hard to say how this would look like (does it take in the VPC? just the specific subnets?). I added a TODO to mark that it will need to change at some point.

PR #454, opened 2023-04-01
Adds DynamodbTable resource and templates

DavidSeptimus-Klotho:
Implements #415 

This PR adds the provider resource and IaC templates for generating DynamoDB tables. It also includes the logic for generating a `Kv` construct's DynamoDB Table resource and IAM related IAM policy statements.

### DIFF
--- a/pkg/provider/aws/kv.go
+++ b/pkg/provider/aws/kv.go
@@ -1,0 +1,43 @@
+package aws
+
+import (
+	"fmt"
+
+	"github.com/klothoplatform/klotho/pkg/core"
+	"github.com/klothoplatform/klotho/pkg/provider/aws/resources"
+)
+
+func (a *AWS) GenerateKvResources(kv *core.Kv, result *core.ConstructGraph, dag *core.ResourceGraph) error {
+	table := resources.NewDynamodbTable(
+		kv,
+		fmt.Sprintf("KV_%s", a.Config.AppName),
+		[]resources.DynamodbTableAttribute{
+			{Name: "pk", Type: "S"},
+			{Name: "sk", Type: "S"},
+		},
+	)
+	table.HashKey = "pk"
+	table.RangeKey = "sk"
+	if err := table.Validate(); err != nil {
+		return err
+	}
+
+	a.MapResourceDirectlyToConstruct(table, kv)
+	dag.AddResource(table)
+
+	upstreamResources := result.GetUpstreamConstructs(kv)
+	for _, res := range upstreamResources {
+		unit, ok := res.(*core.ExecutionUnit)
+		if ok {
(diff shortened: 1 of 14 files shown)


pkg/provider/aws/kv.go
```diff
@@ -0,0 +1,43 @@
+package aws
+
+import (
+	"fmt"
+
+	"github.com/klothoplatform/klotho/pkg/core"
+	"github.com/klothoplatform/klotho/pkg/provider/aws/resources"
+)
+
+func (a *AWS) GenerateKvResources(kv *core.Kv, result *core.ConstructGraph, dag *core.ResourceGraph) error {
```
jhsinger-klotho, line 10: so how are we going to handle this with multiple annotations with different IDs for kv, because this would essentially make us loop over those and create a table per, which i believe is the correct method, but we likely need to handle that in some regard

pkg/provider/aws/kv.go
```diff
@@ -0,0 +1,43 @@
...
+	table.RangeKey = "sk"
+	if err := table.Validate(); err != nil {
+		return err
+	}
+
+	a.MapResourceDirectlyToConstruct(table, kv)
+	dag.AddResource(table)
+
+	upstreamResources := result.GetUpstreamConstructs(kv)
+	for _, res := range upstreamResources {
```
jhsinger-klotho, line 29: nit: upstream constructs, just a bit clearer

pkg/provider/aws/plugin.go
```diff
@@ -33,6 +33,11 @@ func (a *AWS) Translate(result *core.ConstructGraph, dag *core.ResourceGraph) (L
 			if err != nil {
 				return
 			}
+		case *core.Kv:
+			err = a.GenerateKvResources(construct, result, dag)
+			if err != nil {
+				return
+			}
```
jhsinger-klotho, line 40: this needs to handle the dependency between exec unit and ddb table, or that needs to go into the execution unit file since its created second. Not sure if it actually matters for KV since we dont use env vars but my assumption is we will changge that

jhsinger-klotho (commented):
i think we need to discuss kv in depth a little (not this pr but our annotation)

pkg/provider/aws/plugin.go
jhsinger-klotho: nevermind this should be handled by default in generateExecUnit i believe, but this will create a table per id so that discussion i mentioned before needs to happen

pkg/provider/aws/kv.go
jhsinger-klotho: yeah youll just need to add the new reference tothe atable i think because it will only ever have the first kv as a ref and then nothing else

jhsinger-klotho (approved):

PR #456, opened 2023-04-03
adding expose w lambda

### LINKED ISSUE #412: Expose with Lambda
Dependencies: Lambda (Already Completed)

the definition of done:

jhsinger-klotho:
• Does it relate to or fix any issue? closes #412 

adds all the api gateway resources as well as lambda permissions.

The logic to create expose is ported over from deploylib to the best of the ability and should mirror the actions for creating a restApi

- We only set the source arn on the lambda permission to the api gateway arn. This is because previously we constructed the entire arn with account id, region, path, gateway id, etc. In order to do this, we would need to make a getCallerIdentity call and have the region in the code base. 

### Standard checks

- **Unit tests**: Any special considerations? added for some basic expose cases to ensure we create the right resources and dependencies in the graph


pkg/provider/aws/expose.go
```diff
@@ -0,0 +1,123 @@
...
+
+	"github.com/klothoplatform/klotho/pkg/annotation"
+	"github.com/klothoplatform/klotho/pkg/core"
+	"github.com/klothoplatform/klotho/pkg/multierr"
+	"github.com/klothoplatform/klotho/pkg/provider/aws/resources"
+	"github.com/pkg/errors"
+)
+
+func (a *AWS) GenerateExposeResources(gateway *core.Gateway, result *core.ConstructGraph, dag *core.ResourceGraph) error {
+	err := a.CreateRestApi(gateway, result, dag)
```
yuval-klotho, line 17: Is this just a "self-documenting code" thing, or is there a reason for this indirection? Ie, why not just have the `CreateRestApi` body be this body"

pkg/provider/aws/expose.go
```diff
@@ -0,0 +1,123 @@
...
+		execUnit, ok := construct.(*core.ExecutionUnit)
+		if !ok {
+			errs = append(errs, errors.Errorf("Expected construct with id, %s, to be an execution unit", construct.Id()))
+			continue
+		}
+		api_references = append(api_references, execUnit.Provenance())
+		routeTrigger := fmt.Sprintf("%s:%s:%s", route.ExecUnitName, route.Path, route.Verb)
+		triggers[routeTrigger] = routeTrigger
+
+		segments := strings.Split(route.Path, "/")
```
yuval-klotho, line 44: A block comment here about this, and the `for` loop below, would help a lot. I find myself forgetting what the `routh.Path` really is, and then when I see you stripping `:`s and `*`s below, etc, I'm not really sure why.

Basically, I'm not quite sure at a high level what the input is and what processing we need to do with it.

pkg/provider/aws/expose.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func (a *AWS) CreateRestApi(gateway *core.Gateway, result *core.ConstructGraph, dag *core.ResourceGraph) error {
+	var errs multierr.Error
+	api := resources.NewRestApi(a.Config.AppName, gateway)
+	dag.AddResource(api)
+	api_references := []core.AnnotationKey{gateway.AnnotationKey}
+	triggers := map[string]string{}
+
+	resourceMap := map[string]*resources.ApiResource{}
```
yuval-klotho, line 28: minor-ish: with all the things going on below, it's hard to remember exactly what this map is. Maybe rename to `apisBySegment` or similar?

pkg/provider/aws/expose.go
```diff
@@ -0,0 +1,123 @@
...
+		if len(constructResources) != 1 {
+			return nil, errors.Errorf("Expected one resource to be tied to a lambda execution unit, %s, but found %s", unit.ID, strconv.Itoa(len(constructResources)))
+		}
+		function, ok := constructResources[0].(*resources.LambdaFunction)
+		if !ok {
+			return nil, errors.Errorf("Expected resource to be of type, lambda function, for execution unit, %s", unit.ID)
+		}
+		lambdaPermission := resources.NewLambdaPermission(function, core.IaCValue{Resource: method.RestApi, Property: core.ARN_IAC_VALUE}, "apigateway.amazonaws.com", "lambda:InvokeFunction", refs)
+		dag.AddResource(lambdaPermission)
+		dag.AddDependency2(lambdaPermission, function)
```
yuval-klotho, line 107: Huh, am I double-confused? I thought this would be in the reverse order. In the visualization, it would look something like this, I think (right?):

```
GET /foo ➤ APIG ➤ Lambda ➤ Permission
```

Is the `AddDependency2` order the reverse of the arrows? I thought it used to be reverse, but now it's the same.

pkg/provider/aws/expose.go
```diff
@@ -0,0 +1,123 @@
...
+		dag.AddResource(integration)
+		dag.AddDependency2(integration, method)
+		dag.AddDependency2(integration, function)
+		return integration, nil
+	default:
+		return nil, errors.Errorf("Unrecognized integration type, %s, for api gateway", cfg.Type)
+	}
+}
+
+func convertPath(path string) string {
```
yuval-klotho, line 118: convert from what to what? is this leaking a particular library's route syntax (e.g. for var substitution)?

yuval-klotho (commented):
generally looks good. Some readability things I think could be helped with a comment or two, and then I had a couple slightly higher-level questions.

pkg/provider/aws/expose.go
jhsinger-klotho: i mainly just split it out because when we add other resources that can get created from expose, it would be an ugly function

pkg/provider/aws/expose.go
jhsinger-klotho: to be honest, im not entirely sure why all the time either, so i can definitely add comments. Looking through deploylib very much confused me so likely a good pattern to comment about why we have specific logic

pkg/provider/aws/expose.go
jhsinger-klotho: yeah i can rename

pkg/provider/aws/expose.go
jhsinger-klotho: yeah i was trying to say this, we go against the order of the visualizer now. This is the whole talk we had late last week

pkg/provider/aws/expose.go
jhsinger-klotho: i believe this is specific for api gw but let me comment whats going on. Ill just add a bunch of comments

pkg/provider/aws/expose.go
yuval-klotho: Lol, okay I was so lost apparently. I thought we previously went against the order of the visualizer, and with `AddDependency2` we're in sync with it. Why are graphs so hard, haha.

pkg/provider/aws/expose.go
yuval-klotho: Ah cool. If it's specific for API GW that's totally fine in `provider/aws`. I just wanted to make sure it's not specific to express, or something like that.

pkg/provider/aws/expose.go
jhsinger-klotho: nope, opposite, but at this point i dont think we switch back now but im confused looking at our graphs now

PR #457, opened 2023-04-03
Adds ifc2 initial implementation for helm chart creation.

### LINKED ISSUE #418: Helm chart
The ability to deploy a helm chart in pulumi

This should also include switching over the kubernetes plugin to always generate helm charts for kubernetes exec units. We now will need to use the kubernetes type in config to make that determination and should remove the `helm_install: true` configuration since it will always be true for this type.

the definition of done:

DavidSeptimus-Klotho:
resolves #418 

This PR adds IAC for helm charts. 

It includes the following changes:
- Added a relationship between helm charts and their owning exec unit constructs
- Added a resource and IAC factory for pulumi's k8s provider as a temporary mock solution for supplying the provider arg to our helm_chart IAC factory.
- Removed config for manually enabling helm charts
- Added support for writing output associated with the resource dag to the compiled directory (e.g. helm charts)
- Added default helm chart for k8s exec units
- AWS Provider makes sure necessary values are added onto helm chart

I removed the following which david has added. Wanted to make sure i noted it
_- Added nested template functions, `parseVal` and `parseField` to enable contextual parsing of structs inside a nested template (the main reason for this is enabling the use of `IaCValue`s inside nested templates)__
  - Plain HelmChart resources are _not_ rendered in index.ts.
- Converts HelmChart resources to AwsHelmCharts which are specifically configured for EKS



```
Adding resource internal:InternalKlothoPayloads
Adding resource execution_unit:main
Found 2 route(s) on server 'app'                                                                                                                                                                                                                                        unit: main
Found 2 route(s) for middleware 'router'                                                                                                                                                                                                                                unit: main
Adding resource expose:app
Adding resource persist:users
Adding Deployment manifest for exec unit                                                                                                                                                                                                                                unit: main
Adding ServiceAccount manifest for exec unit                                                                                                                                                                                                                            unit: main
Adding Service manifest for exec unit                                                                                                                                                                                                                                   unit: main
Adding TargetGroupBinding manifest for exec unit
```

sample manifest
```
apiVersion: v1
automountServiceAccountToken: true
kind: ServiceAccount
metadata:
  annotations:
    eks.amazonaws.com/role-arn: '{{ .Values.mainRoleArn }}'
  creationTimestamp: null
  labels:
    execUnit: main
  name: main
  namespace: default
```

sample helm chart iac
```

const kubernetesHelmChartMain = new pulumi_k8s.helm.v3.Chart(
        `main`,
        {
            path: `./charts/${``}`,
            values: {"mainImage":awsEcrImagePleaseworkMain.imageName,"mainRoleArn":awsIamRolePleaseworkMainExecutionRole.arn},
        },
        {
            provider: awsEksProviderUNIMPLEMENTEDEksProvider,
        }
    );
```


pkg/config/execution_unit.go
```diff
@@ -12,14 +12,13 @@ type (
 		Type                 string            `json:"type" yaml:"type" toml:"type"`
 		NetworkPlacement     string            `json:"network_placement,omitempty" yaml:"network_placement,omitempty" toml:"network_placement,omitempty"`
 		EnvironmentVariables map[string]string `json:"environment_variables,omitempty" yaml:"environment_variables,omitempty" toml:"environment_variables,omitempty"`
-		HelmChartOptions     *HelmChartOptions `json:"helm_chart_options,omitempty" yaml:"helm_chart_options,omitempty" toml:"helm_chart_options,omitempty"`
+		HelmChartOptions     HelmChartOptions  `json:"helm_chart_options,omitempty" yaml:"helm_chart_options,omitempty" toml:"helm_chart_options,omitempty"`
```
jhsinger-klotho, line 15: By removing the pointer, if we modify this somehwere wouldnt it not persist? like we set values here when we do analysis on certain things which we will need to stay in here

pkg/infra/iac2/templates/helm_chart/factory.ts
```diff
@@ -1,13 +1,47 @@
...
+            //TMPL {{- if .Namespace.Raw }}
+            namespace: args.Namespace,
+            //TMPL {{- end }}
+            //TMPL {{- if .Version.Raw }}
+            version: args.Version,
+            //TMPL {{- end }}
+            //TMPL {{- if .Values.Raw }}
+            values: args.Values,
+            //TMPL {{- end }}
+        },
```
jhsinger-klotho, line 39: lol so ugly to look at, but also really nice that we can do this behind the scenes and it will look clean when its output

pkg/infra/kubernetes/helm_chart.go
```diff
@@ -14,23 +14,49 @@ import (
...
-	Files          []core.File
-	Values         []Value
-	AnnotationKeys []core.AnnotationKey
+	Name               string
+	Chart              string
+	ValuesFiles        []string
+	ExecutionUnits     []*HelmExecUnit
+	Directory          string
+	Files              []core.File
+	Values             []Value `render:"document"`
```
jhsinger-klotho, line 28: im not sure that this will really be rendered correctly because values is a map. This i think is the trickiest part that we have to deal with because our Values today were internally mapped so they placed nicely with how we were doing things in pulumi. But think of an external chart, how do we propose setting values here?

pkg/infra/kubernetes/plugin.go
```diff
@@ -42,13 +43,27 @@ func (p Kubernetes) Translate(constructGraph *core.ConstructGraph, dag *core.Res
...
+				chartContent = &chart.Chart{
+					Metadata: &chart.Metadata{
+						Name:        p.Config.AppName,
+						APIVersion:  "v2",
+						AppVersion:  "0.0.1",
+						Version:     "0.0.1",
+						KubeVersion: ">= 1.19.0-0",
+						Type:        "application",
+					},
+				}
```
jhsinger-klotho, line 59: 👍 i wouldnt make it the appname though because someone could direct us to have multiple charts if they have multiple exec units. I would make it a combination of the execution units in the klotho chart or if you have other ideas im open to them

pkg/infra/kubernetes/plugin.go
```diff
@@ -150,14 +161,16 @@ func (p *Kubernetes) getKlothoCharts(constructGraph *core.ConstructGraph) (map[s
 			}
 		}
 
-		if cfg.HelmChartOptions.Install {
+		if cfg.Type == "kubernetes" {
```
jhsinger-klotho, line 164: could you put a type somewhere like in config for this? We do it in aws, but my feeling is that aws may need to depend on our kubernetes library so maybe rather than the Kubernetes type being in the aws provider config, it should be in here

pkg/lang/javascript/aws_runtime/aws.go
```diff
@@ -194,7 +194,7 @@ func (r *AwsRuntime) AddProxyRuntimeFiles(unit *core.ExecutionUnit, proxyType st
 	var proxyFile []byte
 	unitType := r.Config.GetResourceType(unit)
 	switch proxyType {
-	case "eks":
+	case "kubernetes":
```
jhsinger-klotho, line 197: same comment about the type if it does not cause circular dependencies

pkg/provider/aws/execution_unit.go
```diff
@@ -51,7 +51,17 @@ func (a *AWS) GenerateExecUnitResources(unit *core.ExecutionUnit, result *core.C
...
 	case Kubernetes:
-		return nil
+		// transform kubernetes resources for EKS
+		for _, res := range dag.ListResources() {
+			if res.Provider() != "kubernetes" {
+				continue
+			}
+			for _, ref := range res.KlothoConstructRef() {
+				if ref.ToId() == unit.ToId() {
+					a.MapResourceDirectlyToConstruct(res, unit)
```
jhsinger-klotho, line 61: this would put a reference on every helm chart for every exec unit? Im not sure thats what we want because we store which helm charts are correlated to the exec units

pkg/infra/iac2/templates/helm_chart/factory.ts
DavidSeptimus-Klotho: Yeah. There's potential to build some higher order utility into the resource templates to automatically remove fields with a value of `null` or something like that.

pkg/infra/iac2/templates/kubernetes_provider/factory.ts
```diff
@@ -0,0 +1,15 @@
...
+interface Args {
+    Name: string
+    KubeConfig: string
+}
+
+// noinspection JSUnusedLocalSymbols
+function create(args: Args): pulumi_k8s.Provider {
+    return new pulumi_k8s.Provider(args.Name, {
+        kubeconfig: args.KubeConfig,
+    })
```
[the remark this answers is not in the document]
DavidSeptimus-Klotho: Right. I didn't even bother with the contents of KubeConfig since those rely on the cluster resource and AWS region and it's not clear what patterns we should be using here.

pkg/infra/kubernetes/helm_chart.go
DavidSeptimus-Klotho: I think that's the next piece for me to look into. I haven't gotten to that piece yet.

pkg/infra/kubernetes/plugin.go
DavidSeptimus-Klotho: Your suggestion sounds good enough to me right now.

pkg/provider/aws/execution_unit.go
DavidSeptimus-Klotho: it maps a kubernetes resource (I'll narrow this down to just charts in an update) directly to any execution unit whose `Id` matches any of the resource's `ConstructRefs` -- not every execution unit.

pkg/infra/kubernetes/plugin.go
DavidSeptimus-Klotho: I just added exec unit types as consts in config. Feel free to adjust as desired.

pkg/config/execution_unit.go
```diff
@@ -111,3 +113,10 @@ func (cfg ExecutionUnit) GetExecutionUnitParamsAsKubernetes() KubernetesTypePara
 		return params
 	}
 }
+
+const (
+	Lambda     = "lambda"
+	Kubernetes = "kubernetes"
+	Ecs        = "ecs"
+	AppRunner  = "apprunner"
+)
```
jhsinger-klotho, line 122: these types being in here make our config provider specific. What was our reasoning for moving them in here?

pkg/infra/iac2/templates/aws_helm_chart/factory.ts
```diff
@@ -0,0 +1,47 @@
...
+    FetchOpts?: pulumi.Input<pulumi_k8s.helm.v3.FetchOpts>
+    Values?: pulumi.Inputs
+    Version?: string
+    Namespace?: string
+    KubernetesProvider: pulumi_k8s.Provider
+    dependsOn: pulumi.Input<pulumi.Input<pulumi.Resource>[]> | pulumi.Input<pulumi.Resource>
+}
+
+// noinspection JSUnusedLocalSymbols
+function create(args: Args): pulumi_k8s.helm.v3.Chart {
```
jhsinger-klotho, line 17: the name of aws_helm_chart is a bit weird just because we can deploy non aws helm charts to an eks cluster, but definitely not a blocker. Im just wondering if we can have something more generic, but because eks is our only kubernetes provider to date im ok with this and try to learn from it

pkg/infra/iac2/templates/helm_chart/factory.ts
```diff
@@ -1,13 +1,33 @@
...
-    Directory: string
+    Directory?: string
+    Chart?: string
+    dependsOn: pulumi.Input<pulumi.Input<pulumi.Resource>[]> | pulumi.Input<pulumi.Resource>
 }
 
+// This template is not actually rendered in index.ts right now, but is required for unit tests.
+
 // noinspection JSUnusedLocalSymbols
 function create(args: Args): pulumi_k8s.helm.v3.Chart {
```
jhsinger-klotho, line 14: oh it looks like you solved my above comment lol

pkg/infra/iac2/templates_compiler.go
```diff
@@ -180,6 +181,11 @@ func (tc TemplatesCompiler) RenderPackageJSON() (*javascript.NodePackageJson, er
 
 func (tc TemplatesCompiler) renderResource(out io.Writer, resource core.Resource) error {
 
+	switch resource.(type) {
+	case *kubernetes.HelmChart:
+		return nil // Raw helm charts
+	}
```
jhsinger-klotho, line 187: if we do this, then whats the point of having the helm_chart factory?

pkg/infra/kubernetes/helm_chart.go
```diff
@@ -16,21 +16,22 @@ import (
...
+	ConstructRefs  []core.AnnotationKey
 }
 
 // Provider returns name of the provider the resource is correlated to
 func (chart *HelmChart) Provider() string { return "kubernetes" }
 
-// KlothoResource returns AnnotationKey of the klotho resource the cloud resource is correlated to
-func (chart *HelmChart) KlothoConstructRef() []core.AnnotationKey { return chart.AnnotationKeys }
+// KlothoConstructRef returns a slice containing the ids of any Klotho constructs is correlated to
+func (chart *HelmChart) KlothoConstructRef() []core.AnnotationKey { return chart.ConstructRefs }
```
jhsinger-klotho, line 33: does helm chart need to be a core.Resource? I see value in it being one, but it doesnt look like it maps to the factory and wouldnt deploy. Im wondering if we just omit it and all other providers require the transformation. 

This would make the kubernetes plugin an analysis and transformation rather than a providerplugin.

pkg/infra/kubernetes/plugin.go
```diff
@@ -42,13 +43,31 @@ func (p Kubernetes) Translate(constructGraph *core.ConstructGraph, dag *core.Res
...
+						AppVersion:  "0.0.1",
+						Version:     "0.0.1",
+						KubeVersion: ">= 1.19.0-0",
+						Type:        "application",
+					},
+				}
+			} else {
+				errs.Append(err)
+				continue
+			}
```
jhsinger-klotho, line 68: nice, this is a good improvement for our ux

pkg/provider/aws/resources/helm.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func NewAwsHelmChart(khchart *kubernetes.HelmChart) *AwsHelmChart {
+	chart := &AwsHelmChart{
+		Name:          khchart.Name,
+		Chart:         khchart.Chart,
+		Directory:     khchart.Directory,
+		Files:         khchart.Files,
+		ConstructRefs: khchart.ConstructRefs,
+		EnvVarKeys:    map[string]string{},
```
jhsinger-klotho, line 57: whats the difference of EnvVarKeys and values? it seems like the env vars should also only be values

pkg/config/execution_unit.go
gordon-klotho: You can still modify it: https://go.dev/play/p/yon-_o93FT8

pkg/infra/iac2/templates/helm_chart/factory.ts
gordon-klotho: Could also do it as a function, something like:
```ts
//TMPL {{ argument "fetchOpts" .FetchOpts }}
```

pkg/infra/iac2/templates/helm_chart/factory.ts
```diff
@@ -1,13 +1,33 @@
...
 
 interface Args {
     Name: string
-    Directory: string
+    Directory?: string
+    Chart?: string
+    dependsOn: pulumi.Input<pulumi.Input<pulumi.Resource>[]> | pulumi.Input<pulumi.Resource>
 }
 
+// This template is not actually rendered in index.ts right now, but is required for unit tests.
```
yuval-klotho, line 11: Will it be, and if so, when? If not... we shouldn't create a `factory.ts` just to appease the tests, we should figure out why the test's metaphorical "mental model" broke and fix it.

(If this is just a future task, that's fine — it's not clear to me from the comment, though.)

pkg/infra/iac2/templates_compiler.go
```diff
@@ -329,7 +335,22 @@ func (tc TemplatesCompiler) resolveStructInput(resourceVal *reflect.Value, child
 				tmpl, err := tc.getNestedTemplate(path.Join(
 					camelToSnake(resourceVal.Type().Name()),
 					camelToSnake(correspondingStruct.Type().Name()),
-				))
+				), template.FuncMap{
+					// parseField parses a single field from the supplied value
+					"parseField": func(val reflect.Value, fieldName string) (string, error) {
```
yuval-klotho, line 340: Unused?

pkg/infra/kubernetes/helm_chart.go
yuval-klotho: +1 — I was having the exact same thought as I was scrolling through the above.

The rule we'd discussed earlier is "it's a resource if it has an ARN". I don't think Helm charts do, right? They're more of a document that affects the k8s NodeGroup (?), similar to how a policy document affects an IAM Role.

jhsinger-klotho: well i do think helm chart is a resource because its a different provider than aws technically. For the kubernetes world it would be anything that has a manifest is a resource and a helm chart contains many manifests.

My point here is more so, we have the AWSHelmChart which is the chart after weve done all conversions for aws resources and thats what we deploy. So having the base helm chart be a resource sounds a bit far fetched, the question becomes how do we support base helm charts then

pkg/infra/iac2/templates_compiler.go
jhsinger-klotho: yeah i wasnt entirely sure where this was used

pkg/infra/kubernetes/helm_chart.go
yuval-klotho: Right now, our provider is for AWS. We have plans to write a different, k8s-only provider — but that's future work. So while we're in the mode of writing the AWS provider, imo we should really try to stick to "resource ⟺ arn-haver".

jhsinger-klotho: yeah but then we wouldnt deploy any eks exec units or its corresponding resources

jhsinger-klotho: so helm chart has to be a resource to some degree, we just have the aws helm chart which is the resource and i dont think the generic one needs to be one

yuval-klotho: We talked about this offline, I'm aligned 👍

pkg/infra/kubernetes/plugin.go
```diff
@@ -42,13 +43,31 @@ func (p Kubernetes) Translate(constructGraph *core.ConstructGraph, dag *core.Res
 	// For exec units that specify their own chart, we want to render and replace
 	for dir, khChart := range klothoCharts {
 		dirToLoad := filepath.Join(p.Config.Path, dir)
-		chart, err := p.helmHelper.LoadChart(dirToLoad)
+		chartContent, err := p.helmHelper.LoadChart(dirToLoad)
+
 		if err != nil {
-			errs.Append(err)
-			continue
+			if err.Error() == "Chart.yaml file is missing" {
```
gordon-klotho, line 49: Not a huge fan of checking error messages. Is there not a type you can `errors.Is` with?

gordon-klotho (approved):

pkg/infra/kubernetes/plugin.go
jhsinger-klotho: not sure, this was from before i took over, so i can change it

PR #463, opened 2023-04-04
use managed policies instead of inline and other small changes

jhsinger-klotho:

These changes relate to getting the go-cloudfs app working. In here we:]
- Use managed policies instead of inline due to having pulumi output in the resources section
- Template the s3 website because it fails if exists
- Modify the syntax of our async export so pulumi can pick it up (includes adding tsconfig)
- Make our assume roly policies valid json

### Standard checks

- **Unit tests**: Any special considerations? updated all unit tests which have policy changes


pkg/infra/iac2/templates/s3_bucket/factory.ts
```diff
@@ -20,9 +20,11 @@ function create(args: Args): aws.s3.Bucket {
                     bucketKeyEnabled: true,
                 },
             },
+            //TMPL {{ if .IndexDocument.Raw }}
```
yuval-klotho, line 23: https://www.youtube.com/watch?v=q5Dj5G1kaqI

pkg/infra/iac2/tsconfig.json
yuval-klotho, line 1: Nice! Thanks for adding this.

yuval-klotho (approved):

PR #464, opened 2023-04-04
Add dependencies via reflection

gordon-klotho:
This should make it much easier to add dependencies in many (most?) cases.

### Standard checks

### DIFF
--- a/pkg/core/resource_graph.go
+++ b/pkg/core/resource_graph.go
@@ -91,41 +91,40 @@ func (rg *ResourceGraph) TopologicalSort() ([]string, error) {
 // - `DependencyMap  map[string]Resource`
 // - `SpecificDepMap map[string]*T`
 func (rg *ResourceGraph) AddDependenciesReflect(source Resource) {
-	rv := reflect.ValueOf(source)
-	rt := rv.Type()
-	if rt.Kind() == reflect.Pointer {
-		rv = rv.Elem()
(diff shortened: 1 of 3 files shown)


pkg/core/resource_graph.go
```diff
@@ -10,58 +12,120 @@ type (
...
+//
+// Supported field types (`*T` is a struct that implements Resource)
+// - `SingleDependency   Resource`
+// - `SpecificDependency *T`
+// - `DependencyArray  []Resource`
+// - `SpecificDepArray []*T`
+// - `DependencyMap  map[string]Resource`
+// - `SpecificDepMap map[string]*T`
+func (rg *ResourceGraph) AddDependenciesReflect(source Resource) {
+	rv := reflect.ValueOf(source)
```
yuval-klotho, line 91: I find these hypershort names hard to follow through the method. I'd prefer `resVal` etc.

yuval-klotho:
this LGTM, but I'd like to get input from Jordan and David as well to make sure we all have consensus around it

jhsinger-klotho:
yeah im fine going this route, theres still scenarios in which we would need to invoke AddDependency though i believe, but well discover those

DavidSeptimus-Klotho (approved):
It seems neat to me. I'm curious to see how the magical aspect of this plays out over time (vs explicit dependencies).

PR #466, opened 2023-04-04
Use reflect dependencies

gordon-klotho:
Mostly updating the tests to include what was previously relied on transitive dependencies to be explicit ones.

Main things to review:
- Support for IaCValue in reflect dependencies
- the `coretesting` package and how it's used in tests

### Standard checks


pkg/core/coretesting/package.go
```diff
@@ -0,0 +1,3 @@
+// coretesting contains utilities to make testing things involving the `core` types and functions easier.
+// This is a separate package from core so it can be depended on only by tests and therefore pruned in non-test builds.
```
gordon-klotho, line 2: note: this is similar to `net/http/httptest` in stdlib

pkg/core/coretesting/resources.go
```diff
@@ -0,0 +1,49 @@
...
+type (
+	StringDep = graph.Edge[string]
+
+	ResourcesExpectation struct {
+		Nodes []string
+		Deps  []StringDep
+
+		// OnlyContains only check whether the dag contains the `.Nodes` and `.Deps`. If false,
+		// checks full equality.
+		OnlyContains bool
```
yuval-klotho, line 20: "Only" is an ambiguous word, and confused me in this usage. To me, this reads as "check that the dag Only Contains these elements", which is pretty much the opposite of what I think you meant ("Only check that the dag Contains these elements").

Maybe `PartialMatchAccepted` or such?

pkg/provider/aws/aws_test.go
```diff
@@ -0,0 +1,5 @@
+package aws
+
+import "github.com/klothoplatform/klotho/pkg/core/coretesting"
+
+type DagExpectation = coretesting.ResourcesExpectation
```
yuval-klotho, line 5: What's the benefit of this type alias? (and similarly in `resources_test.go`)

pkg/core/coretesting/resources.go
```diff
@@ -0,0 +1,49 @@
...
+
+func (expect ResourcesExpectation) Assert(t *testing.T, dag *core.ResourceGraph) {
+	var res []string
+	for _, r := range dag.ListResources() {
+		res = append(res, r.Id())
+	}
+	if expect.OnlyContains {
+		for _, r := range expect.Nodes {
+			assert.Contains(t, res, r)
+		}
```
yuval-klotho, line 32: Would this work?

```suggestion
		assert.Subset(t, res, expect.Nodes)
```

Ditto for the Deps below. 

If so, you might even consider something like:

```
matchFunc := assert.ElementsMatch
if expect.OnlyContains {
   matchFunc = assert.Subset
}
```

...and then you don't need the if-elses in the rest of this function.

pkg/provider/aws/aws_test.go
gordon-klotho: Just makes it shorter. In the context of the provider, "DAG" is the resource DAG, but without that context (ie, in coretesting) it needs to be specified.
Easy to change if we don't like it.

pkg/core/coretesting/resources.go
gordon-klotho: Ah, thanks - I was looking for this assert function but didn't think to check subset (looking for like a "ContainsAll" or "In").

pkg/core/coretesting/resources.go
```diff
@@ -0,0 +1,45 @@
...
+		AssertSubset bool
+	}
+)
+
+func (expect ResourcesExpectation) Assert(t *testing.T, dag *core.ResourceGraph) {
+	var res []string
+	for _, r := range dag.ListResources() {
+		res = append(res, r.Id())
+	}
+	if expect.AssertSubset {
```
yuval-klotho, line 29: Aw, you didn't like the `matchFunc` idea? imo it makes it a bit more readable because it keeps branches out of the meat of the func, but definitely not a blocker.

pkg/provider/aws/aws_test.go
yuval-klotho: It adds a bit of indirection for little gain, so imo it's not worth it — but definitely not a blocker.

yuval-klotho (approved):

pkg/core/coretesting/resources.go
gordon-klotho: Mostly because I wanted the `ElementsMatch` ordering to match all the other asserts (ie, `assert(expected, actual)`) - even though it oddly uses `listA`, `listB` instead for some reason.

yuval-klotho: Why is `listA, listB` incompatible with `expected, actual`?

gordon-klotho: Mostly because the order is dictated by the `Subset(outer, inner)` where we pass in `Subset(dag, expected)` which is in the opposite order as `ElementsMatch(expected, dag)`.

yuval-klotho (approved):

PR #469, opened 2023-04-05
adding service account roles and ability to run pulumi apply

### LINKED ISSUE #445: Create OpenIdConnectProvider for eks cluster
Today we generate create an OpenIdConnectProvider so that pods can use IRSA , https://github.com/klothoplatform/klotho/blob/ifc2-foundations/pkg/infra/pulumi_aws/iac/eks.ts#L887. We need to replicate this in the new format.

the definition of done for each task is:

### LINKED ISSUE #446: Create necessary Pod Resources for eks pod
To have a running pod setup in the way we need, we will need to create a [Service Account Role](https://github.com/klothoplatform/klotho/blob/ifc2-foundations/pkg/infra/pulumi_aws/iac/eks.ts#L905) and render the necessary values to the pods helm chart.

Definition of Done:
- Create the necessary resources within the provider ( There are likely no templates needing creation here)
- Ensure the providers resources can be wired into the helm charts values
- Ensure we can have a successful pod deployment

jhsinger-klotho:
• Does it relate to or fix any issue? closes #445 and closes #446

This solves our oidc provider and creation of service account roles by:
 - Switching to aws-native for eks clusters where oidc provider is baked in. This prevents us from having to write a lot of custom logic and figuring out how to wire that into the iac plugin
 - Creating a service account assumer role policy for kubernetes exec units iam role

Modifications to Iam IR:
- Had to add principal and condition to the policy document
- Make AssumeRoleDocument a policy document instead of string, so we can use resources within it
- These get rendered as null or '' sometimes, which im not sure if that will cause issues. If it does we may have to make an adjustment to the way it is rendered

To do this we also had to add the functionality to run pulumi.all.apply by:
- Adding an appliedoutputs field to a resourceCreationTemplate
- On dependent Core.IaCValues we add to the templates list of applied outputs
- If the list in non 0 in length we render the pulumi.all.apply when rendering the templates return statement.
example
```
const awsIamRoleTsReMainExecutionRole = pulumi.all([awsEksClusterTsReMyCluster1.openIdConnectIssuerUrl, awsEksClusterTsReMyCluster1.arn]).apply(([cluster_oidc_url, cluster_arn]) => { return new aws.iam.Role(`ts-re-main-ExecutionRole`, {
        assumeRolePolicy: {Version: `2012-10-17`,
Statement: [{Effect: `Allow`,
Resource: [],
Principal: {Service: ``,
Federated: `arn:aws:iam::${cluster_arn.split(':')[4]}:oidc-provider/${cluster_oidc_url}`,
},
Condition: {StringEquals: {[`${cluster_oidc_url}:sub`]:`system:serviceaccount:default:main`},
},
}],
},
```

### Standard checks

- **Unit tests**: Any special considerations? added unit tests to the k8s route and the functions added to the resource creation template


pkg/infra/iac2/templates_compiler.go
```diff
@@ -75,7 +75,7 @@ func (s stringTemplateValue) Raw() interface{} {
 
 func (v nestedTemplateValue) Parse() (string, error) {
 	childVal := v.childValue
-	return v.tc.resolveStructInput(&v.parentValue, childVal, v.useDoubleQuotedStrings, v.iacTag)
```
yuval-klotho, line 78: I don't really understand what's going on here. You pass in the `*ResourceCreationTemplate`, and then it looks like you modify it but don't ever use it. But you never do anything with this original reference, so where does that information all go? What uses the template you build up?

(This is different from the unit test, where you first create an RCT, then pass it into this method, and then check that it got built up correctly.)

gordon-klotho:
Not sure if it's necessarily cleaner, but an alternative would be to use interpolate/apply on the field values instead. This would move the pulumi-type logic into the factory.

For example:
```ts
// before
const awsIamRoleTsReMainExecutionRole = pulumi
    .all([awsEksClusterTsReMyCluster1.openIdConnectIssuerUrl, awsEksClusterTsReMyCluster1.arn])
    .apply(([cluster_oidc_url, cluster_arn]) => {
        return new aws.iam.Role(`ts-re-main-ExecutionRole`, {
            assumeRolePolicy: {
                Version: `2012-10-17`,
                Statement: [
                    {
                        Effect: `Allow`,
                        Resource: [],
                        Principal: {
                            Service: ``,
                            Federated: `arn:aws:iam::${
                                cluster_arn.split(':')[4]
                            }:oidc-provider/${cluster_oidc_url}`,
                        },
                        Condition: {
                            StringEquals: {
                                [`${cluster_oidc_url}:sub`]: `system:serviceaccount:default:main`,
                            },
                        },
                    },
                ],
            },
        })
    })

// after
return new aws.iam.Role(`ts-re-main-ExecutionRole`, {
    assumeRolePolicy: {
        Version: `2012-10-17`,
        Statement: [
            {
                Effect: `Allow`,
                Resource: [],
                Principal: {
                    Service: ``,
                    Federated: pulumi.interpolate`arn:aws:iam::${
                        awsEksClusterTsReMyCluster1.arn.split(':')[4]
                    }:oidc-provider/${awsEksClusterTsReMyCluster1.openIdConnectIssuerUrl}`,
                },
                Condition: awsEksClusterTsReMyCluster1.openIdConnectIssuerUrl.apply((url) => {
                    return {
                        StringEquals: {
                            [`${url}:sub`]: `system:serviceaccount:default:main`,
                        },
                    }
                }),
            },
        ],
    },
})
```

Doing `all().apply(() => new)` type stuff makes it pretty hard to read, so if we can avoid that, I think it'd be nice.

yuval-klotho (commented):
I feel like I _almost_ understand this PR, but not quite. (I'm sure it's a me-problem.) I dropped one specific comment about not seeing how the tmpl you build actually gets populated, but I also don't see/remember how these dependent IaCValues actually get used in the template. Are they always just strings, and this auto-`pulumi.all` trick is what turns the promises/whatever into actual strings?

A bit of doc in the `iac2.go` block could help here too, I think.

jhsinger-klotho:
> Not sure if it's necessarily cleaner, but an alternative would be to use interpolate/apply on the field values instead. This would move the pulumi-type logic into the factory.
> 
> For example:
> 
> ```ts
> // before
> const awsIamRoleTsReMainExecutionRole = pulumi
>     .all([awsEksClusterTsReMyCluster1.openIdConnectIssuerUrl, awsEksClusterTsReMyCluster1.arn])
>     .apply(([cluster_oidc_url, cluster_arn]) => {
>         return new aws.iam.Role(`ts-re-main-ExecutionRole`, {
>             assumeRolePolicy: {
>                 Version: `2012-10-17`,
>                 Statement: [
>                     {
>                         Effect: `Allow`,
>                         Action: [`sts:AssumeRoleWithWebIdentity`],
>                         Resource: [],
>                         Principal: {
>                             Service: ``,
>                             Federated: `arn:aws:iam::${
>                                 cluster_arn.split(':')[4]
>                             }:oidc-provider/${cluster_oidc_url}`,
>                         },
>                         Condition: {
>                             StringEquals: {
>                                 [`${cluster_oidc_url}:sub`]: `system:serviceaccount:default:main`,
>                             },
>                         },
>                     },
>                 ],
>             },
>         })
>     })
> 
> // after
> const awsIamRoleTsReMainExecutionRole = new aws.iam.Role(`ts-re-main-ExecutionRole`, {
>     assumeRolePolicy: {
>         Version: `2012-10-17`,
>         Statement: [
>             {
>                 Effect: `Allow`,
>                 Action: [`sts:AssumeRoleWithWebIdentity`],
>                 Resource: [],
>                 Principal: {
>                     Service: ``,
>                     Federated: pulumi.interpolate`arn:aws:iam::${
>                         awsEksClusterTsReMyCluster1.arn.split(':')[4]
>                     }:oidc-provider/${awsEksClusterTsReMyCluster1.openIdConnectIssuerUrl}`,
>                 },
>                 Condition: awsEksClusterTsReMyCluster1.openIdConnectIssuerUrl.apply((url) => {
>                     return {
>                         StringEquals: {
>                             [`${url}:sub`]: `system:serviceaccount:default:main`,
>                         },
>                     }
>                 }),
>             },
>         ],
>     },
> })
> ```
> 
> Doing `all().apply(() => new)` type stuff makes it pretty hard to read, so if we can avoid that, I think it'd be nice. Plus, it makes `awsIamRoleTsReMainExecutionRole: aws.iam.Role` and not `: pulumi.Output<aws.iam.Role>`.

interpolate wouldnt work on the stringequals portion though because thats not a valid key

pkg/infra/iac2/templates_compiler.go
jhsinger-klotho: ill add some docs as well, but the template has an appliedOutputs list now and we can add to that list if we recognize fields within the returned resource need to be applied in order for it to work properly.

pkg/infra/iac2/templates_compiler.go
```diff
@@ -263,7 +279,7 @@ func (tc TemplatesCompiler) resolveDependencies(resource core.Resource) string {
 }
 
 // resolveStructInput translates a value to a form suitable to inject into the typescript as an input to a function.
-func (tc TemplatesCompiler) resolveStructInput(resourceVal *reflect.Value, childVal reflect.Value, useDoubleQuotedStrings bool, iacTag string) (string, error) {
+func (tc TemplatesCompiler) resolveStructInput(resourceVal *reflect.Value, childVal reflect.Value, useDoubleQuotedStrings bool, iacTag string, appliedOutputs *[]AppliedOutput) (string, error) {
```
gordon-klotho, line 282:
```suggestion
func (tc TemplatesCompiler) resolveStructInput(resourceVal *reflect.Value, childVal reflect.Value, useDoubleQuotedStrings bool, iacTag string, appliedOutputs []AppliedOutput) (string, error) {
```

Arrays/slices are already pointers.

pkg/provider/aws/resources/eks.go
```diff
@@ -100,6 +104,21 @@ func CreateEksCluster(appName string, clusterName string, subnets []*Subnet, sec
...
+//
+// If the dag does not contain the resource or the resource is not an EksCluster, it will return nil
+func GetEksCluster(appName string, clusterId string, dag *core.ResourceGraph) *EksCluster {
+	if clusterId == "" {
+		clusterId = DEFAULT_CLUSTER_NAME
+	}
+	cluster := NewEksCluster(appName, clusterId, nil, nil, nil)
+	resource := dag.GetResource(cluster.Id())
+	if existingCluster, ok := resource.(*EksCluster); ok {
+		return existingCluster
```
gordon-klotho, line 117: When would `existingCluster` not be `cluster`?

gordon-klotho (commented):
Can you post an updated example output? Mostly looks good.

pkg/infra/iac2/templates/iam_role/factory.ts
```diff
@@ -12,7 +12,7 @@ interface Args {
 // noinspection JSUnusedLocalSymbols
 function create(args: Args): aws.iam.Role {
     return new aws.iam.Role(args.Name, {
-        assumeRolePolicy: args.AssumeRolePolicyDoc,
+        assumeRolePolicy: JSON.stringify(args.AssumeRolePolicyDoc),
```
yuval-klotho, line 15: One thing to note here is that if you don't do the stringify, then this `factory.ts` will still compile, but the eventual `index.ts` will fail. We should consider whether this is something we can guard against (either by finding an alternative index.ts-rendering that allows both formats, or by finding Go+TS representations that let us catch this in unit tests).

I think the only actionable thing I'd kinda like to see is:
* a brief writeup of what we're doing, why we're doing it, its limitations, and alternatives we considered (along with why we didn't do them)
* a ticket to consider revisiting the issue in the fit-and-finish milestone, with a link to that writeup. We can always punt on that ticket once we get to that milestone (and the ticket should say that, so we don't forget).

Basically, this is subtle enough that I think there's a decent chance we'd want to revisit it, and when we do there's a really good chance we won't remember why we did it this way unless we write it down.

yuval-klotho (commented):
lgtm, just an internal-documentation note that I do think is decently important. It doesn't block this PR per se, but I do think we should do it soon (like, today or tomorrow, or else we'll start forgetting it).

pkg/provider/aws/resources/eks.go
jhsinger-klotho: cluster wouldnt have the references to the subnets, or other things that the existing one would have, The only reason i create cluster is to be able to use it for its id

pkg/infra/iac2/templates_compiler.go
jhsinger-klotho: will update

pkg/infra/iac2/templates/iam_role/factory.ts
jhsinger-klotho: yeah where do you want us to track all of this? do you want me to just write a task for it in that milestone with all of the notes? 

Also regarding the internal documentation, is that in regards to how to use pulumi.apply for us or why we need to do json.stringify?

pkg/infra/iac2/templates_compiler.go
yuval-klotho: > Arrays/slices are already pointers.

This approach affects the underlying var from the call site, which wouldn't happen without the `*`.

```go
func TestAppend(t *testing.T) {
	t.Run("pointer to array", func(t *testing.T) {
		var slice []string
		appendHello(&slice)
		assert.Equal(t, []string{"hello"}, slice)
	})

	t.Run("array itself", func(t *testing.T) {
		var slice []string
		appendWorld(slice)
		assert.Equal(t, []string{"world"}, slice)
	})
}

func appendHello(strs *[]string) {
	*strs = append(*strs, "hello")
}

func appendWorld(strs []string) {
	strs = append(strs, "world")
}

```

`pointer_to_array` succeeds, `array_itself` fails.

jhsinger-klotho: actually i tried that and it didnt persist the data to the one

pkg/infra/iac2/templates/iam_role/factory.ts
yuval-klotho: I think a task is  fine in the milestone with all the notes there. If that becomes annoying to discuss (when we get to it), we can always convert it to a gdoc or to a GH Discussion.

I was thinking basically both the things you mentioned. But more specifically, I think this is probably all the info we'd want to capture (though obviously feel free to add anything else you think is relevant):

* why we need the pulumi.apply (not so much "how to use it", but really "why did we need to develop all this in the first place")
* and then the fact that we wrap it in the `pulumi.apply` (so I guess, the "how" from bullet 1)
* and then why we need the stringify, what will go wrong if we don't include it, and specifically calling out that this means we lose the guarantee of "if factory.ts compiles and unit tests pass, then index.ts will also compile"
* any alternatives we considered (I think it was just inlining the pulumi.interpolate?), and why those didn't work

jhsinger-klotho: created https://github.com/orgs/klothoplatform/projects/1/views/18?pane=issue&itemId=24930378 with details

pkg/infra/iac2/templates_compiler.go
gordon-klotho: Oh, I see... I didn't trace it all the way down. It's a little weird, possibly worth a wrapper type but good for now 👍

pkg/provider/aws/resources/eks.go
gordon-klotho: Ah, right - I got confused between the `New` and `Create` functions.

gordon-klotho (approved):

Issue #467, opened 2023-04-04
Get ecr creds to use in image push


jhsinger-klotho:
when trying to push an image we currently see 
```
  docker:index:Image (ts-re-main):
    error: EOF
    : This error is most likely due to incorrect or mismatched registry credentials. Please double check you are using the correct credentials and registry name.
```
this is because we dont have creds. We will want to get those creds via https://github.com/pulumi/pulumi-awsx/blob/de12084d5d270d8c3313cd4e3742d7fcf22f1691/awsx/ecr/image.ts#L70 and use the username and password in our docker images registry.

This will likely need to be custom logic that resides in the iac plugin.

Definition of Done:
- When we generate an ecr repo, we generate corresponding creds
- We attach those creds to images which are intended to be pushed to that repo
- we can successfully push images to the repo

PR #472, opened 2023-04-05
create availability zone resource to be used in iac

### LINKED ISSUE #468: Region to AZ Mapping and assign subnets
When currently trying to deploy, we see 
```
* creating EC2 VPC Endpoint (com.amazonaws.eu-west-1.lambda): DuplicateSubnetsInSameZone: Found another VPC endpoint subnet in the availability zone of subnet-0a7bfb78cc4591ffa. VPC endpoint subnets should be in different availability zones supported by the VPC endpoint service.
```
pulumi sticks every subnet in the same AZ, so we have no ability to reconcile this without making api calls in the compiler. This right now effects interface vpc endpoints but im sure it will affect other things (we know we hit this in memorydb already). 

To move forward we will map azs to regions and support 3 azs within the regions we support. We will then need to modify the subnet IR to be able to assign azs so for resources which cant be in multiple subnets in the same az we can make the appropriate determinations.

Definition of done:
- Be able to assign subnets to an AZ
- Support 3 Azs in 10 larger regions (us regions, eu regions definitely)
- be able to create interface vpc endpoints

jhsinger-klotho:
• Does it relate to or fix any issue? closes #468  and #467

Create an availability zones resource so we can correctly assign subnets and create interface vpc endpoints (For reasoning look at the issue in the GH link)

- We create 1 private and public subnet in az[0]
- We create 1 private and public subnet in az[1]
- We assign our interface endpoints only to the private subnets

We also update the docker image (ecr_image) factory to grab the repos creds so it can push the image

iac code:
```

const awsAvailabilityZonesAvailabilityZones = pulumi.output(aws.getAvailabilityZones({
        state: "available",
    }));
const awsVpcSubnetTsRePrivate1 = new aws.ec2.Subnet(`ts_re_private1`, {
        vpcId: awsVpcTsRe.id,
        cidrBlock: `10.0.0.0/18`,
        availabilityZone: awsAvailabilityZonesAvailabilityZones.names[0]
    });

const awsVpcSubnetTsRePrivate2 = new aws.ec2.Subnet(`ts_re_private2`, {
        vpcId: awsVpcTsRe.id,
        cidrBlock: `10.0.64.0/18`,
        availabilityZone: awsAvailabilityZonesAvailabilityZones.names[1]
    });

const awsVpcSubnetTsRePublic1 = new aws.ec2.Subnet(`ts_re_public1`, {
        vpcId: awsVpcTsRe.id,
        cidrBlock: `10.0.128.0/18`,
        availabilityZone: awsAvailabilityZonesAvailabilityZones.names[0]
    });

const awsVpcSubnetTsRePublic2 = new aws.ec2.Subnet(`ts_re_public2`, {
        vpcId: awsVpcTsRe.id,
        cidrBlock: `10.0.192.0/18`,
        availabilityZone: awsAvailabilityZonesAvailabilityZones.names[1]
    });
```

successful deploy:
```
 +   ├─ aws:ec2:VpcEndpoint      ts_re_dynamodb             created (7s)
 +   ├─ aws:ec2:Subnet           ts_re_public1              created (1s)
 +   ├─ aws:ec2:VpcEndpoint      ts_re_s3                   created (7s)
 +   ├─ aws:ec2:Subnet           ts_re_private2             created (2s)
 +   ├─ aws:ec2:Subnet           ts_re_private1             created (2s)
 +   ├─ aws:ec2:Subnet           ts_re_public2              created (2s)
 +   ├─ aws:ec2:NatGateway       ts_re_private2             created (95s)
 +   ├─ aws:ec2:NatGateway       ts_re_private1             created (85s)
 +   ├─ aws:ec2:VpcEndpoint      ts_re_sns                  created (53s)
 +   ├─ aws:ec2:VpcEndpoint      ts_re_secretsmanager       created (53s)
 +   ├─ aws:ec2:VpcEndpoint      ts_re_sqs                  created (53s)
 +   └─ aws:ec2:VpcEndpoint      ts_re_lambda               created (50s)

```

### Standard checks


pkg/infra/iac2/templates/availability_zones/factory.ts
```diff
@@ -0,0 +1,13 @@
+import * as aws from '@pulumi/aws'
+import * as pulumi from '@pulumi/pulumi'
+
+interface Args {}
+
+// noinspection JSUnusedLocalSymbols
+function create(args: Args): pulumi.Output<pulumi.UnwrappedObject<aws.GetAvailabilityZonesArgs>> {
+    return pulumi.output(
+        aws.getAvailabilityZones({
+            state: 'available',
```
yuval-klotho, line 10: I should hope so! Otherwise it's just a Zone.

pkg/infra/iac2/templates/ecr_image/package.json
```diff
@@ -3,6 +3,7 @@
     "dependencies": {
         "@pulumi/aws": "^5.16.0",
         "@pulumi/pulumi": "^3.40.2",
-        "@pulumi/docker": "^v4.1.0"
+        "@pulumi/docker": "^v4.1.0",
+        "@types/node": "^18.15.11"
```
yuval-klotho, line 7: Is this used?

yuval-klotho (commented):
lgtm other than the extra dependency

yuval-klotho (approved):

pkg/infra/iac2/templates/ecr_image/package.json
jhsinger-klotho: yeah we need it for the Buffer type even though its not imported

jhsinger-klotho: i guess our dep check says we dont but let me see what happens if i remove it

pkg/infra/iac2/templates_compiler.go
```diff
@@ -398,7 +398,10 @@ func (tc TemplatesCompiler) handleIaCValue(v core.IaCValue) (string, error) {
 			return output, err
 		}
 		return output, nil
+	} else if reflect.TypeOf(v.Resource) == reflect.TypeOf((*resources.AvailabilityZones)(nil)) {
```
gordon-klotho, line 401:
```suggestion
	} else if _, ok := v.Resource.(*resources.AvailabilityZones); ok {
```
Use type assertion here, it's more idiomatic and performant.

pkg/infra/iac2/templates_compiler.go
```diff
@@ -430,6 +433,19 @@ func (tc TemplatesCompiler) handleIaCValue(v core.IaCValue) (string, error) {
...
+		method, ok := v.Resource.(*resources.ApiMethod)
+		if !ok {
+			return "", errors.Errorf("unsupported resource type %T for '%s'", v.Resource, v.Property)
+		}
+		verb := strings.ToUpper(method.HttpMethod)
+		if verb == "ANY" {
+			verb = "*"
+		}
+		accountId := resources.NewAccountId()
+		region := resources.NewRegion()
```
gordon-klotho, line 446: Is there a reason the AccountId and the Region don't go on the struct if it depends on them?
Can this be derived from the RestApi arn instead? Something like
```ts
pulumi.interpolate`${api.arn}/*/%s${method.path}`
```

Non-blocking, but hopefully we don't leave it in for too long

pkg/provider/aws/resources/vpc.go
```diff
@@ -68,26 +69,37 @@ type (
...
 	if dag.GetResource(vpc.Id()) != nil {
 		return vpc
 	}
 
 	region := NewRegion()
+	azs := NewAvailabilityZones()
 	igw := NewInternetGateway(appName, "igw1", vpc)
 
 	dag.AddResource(region)
+	dag.AddResource(azs)
```
gordon-klotho, line 85: nit: `AddDependency` will add the resources for you if they don't exist, so this isn't needed

gordon-klotho (approved):

pkg/infra/iac2/templates_compiler.go
jhsinger-klotho: the rest api arn doesnt have either of them and doesnt follow the same format, i had tried that originally, but then had to redeploy our old version to see the correct format.

The permissions isnt dependent on the region or the account id necessarily so it would be weird to pack those in since they dont really fit on that or a method. Because region and account id are global i dont think this is the worst way to go about it for now, but i do agree theres room for improvement.

I dont think well see too many things like this,
just for reference what we want is 
`  "AWS:SourceArn": "arn:aws:execute-api:eu-west-1:338991950301:es3nxti7se/*/GET/read-file/{path}"`


and the api gateway arn is `arn:aws:api-gateway:eu-west-1::/rest-api/es3nxti7se` or something like that

pkg/infra/iac2/templates_compiler.go
jhsinger-klotho: ill switch this over

pkg/infra/iac2/templates_compiler.go
gordon-klotho: > The permissions isnt dependent on the region or the account id necessarily

It doesn't in a general sense, but when we render the IaC it definitely depends on it in code (ie, it would never be valid to put the Region or AccountId variables after this). The DAG is supposed to encode dependencies like this.

jhsinger-klotho: yeah i mean we could put those in the core.IaCValue somehow so it has them or in the method so it would have them, but i feel like thats just muddying the water a little bit. The weird thing here is that were saying give me the arn to the method, but the iac doesnt have the ability to do that which is why we have to construct it via pulumi.

other iacs might have the ability to just grab the arn which is why to me the provider is doing what it needs to in saying the methods arn. Hard to know until we support more iacs though

PR #473, opened 2023-04-05
nicer formatting for yaml-based tests

yuval-klotho:
I found the current tests pretty hard to read because of this formatting, so I took a crack at cleaning them up. I find them a lot more readable now, but if people don't like the extra indirection/transform within the test-expect, I can retract this.

### Standard checks

- **Unit tests**: n/a
- **Docs**: n/a

### DIFF
--- a/pkg/config/execution_unit.go
+++ b/pkg/config/execution_unit.go
@@ -106,8 +106,6 @@ func (cfg ExecutionUnit) GetExecutionUnitParamsAsKubernetes() KubernetesTypePara
 	params := KubernetesTypeParams{}
 	if err := json.Unmarshal(jsonString, &params); err != nil {
 		zap.S().Error(err)
-		return params
-	} else {
-		return params
 	}
+	return params
 }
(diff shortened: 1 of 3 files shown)


gordon-klotho:
You could make a case for these tests to be actual `.yaml` files and then tested via `go:embed`. That way we could utilize IDE formatting too.

yuval-klotho:
Personally, I find that to be even harder to read — it's too much distance from the input and the expected. Given a choice between the code as it is (before this PR) and embed'ed, I would pick as-is.

gordon-klotho:
That's fair too. Personally when files start to get > 200-300 lines long my eyes start to glaze over.

gordon-klotho (approved):

yuval-klotho:
Yeah, mine definitely do too — I don't see a great solution to that aspect of it. :-\

Thanks for the checkmark. I'll keep this up a bit to see if the others have thoughts on this. (I need to head out in a few, anyway).

Issue #422, opened 2023-03-28
Cloudfront for s3


jhsinger-klotho:
Dependent on existing s3 support (Done already)

the definition of done:
- The Internal Representation within the provider (basic unit tests)
- The template within the IaC Plugin
- the proper logic in the provider plugin to create the resource for the corresponding Construct
- Any Configuration work we may need done (this does not include imports work)

PR #475, opened 2023-04-06
cloudfront for iac2

jhsinger-klotho:
• Does it relate to or fix any issue? #422 and #423 

This adds the following IRs and factories:
- Cloudfront Distribution
- Cloudfront Origin Access Identity
- S3 Bucket Policy
- 
Which gives us the ability to:
- Front an api gateway stage with cloudfront(expose)
- Front an s3 bucket with cloudfront(static unit)

Until we understand how to do path based routing on the identities we can only support a single origin behind cloudfront, which was our previous and now current state.

```
     pulumi:pulumi:Stack             ts-re-test
 +   └─ aws:cloudfront:Distribution  ts-re-cdn1  created (233s)

```

### Standard checks

- **Unit tests**: Any special considerations? added for newly added functions and remodeled the static unit one to use the new coretesting lib

### DIFF
--- a/pkg/infra/iac2/templates/origin_access_identity/factory.ts
+++ b/pkg/infra/iac2/templates/origin_access_identity/factory.ts
@@ -1,5 +1,4 @@
 import * as aws from '@pulumi/aws'
-import * as pulumi from '@pulumi/pulumi'
 
 interface Args {
     Name: string
--- a/pkg/infra/iac2/templates/origin_access_identity/factory.ts
+++ b/pkg/infra/iac2/templates/origin_access_identity/factory.ts
@@ -1,0 +1,14 @@
+import * as aws from '@pulumi/aws'
+import * as pulumi from '@pulumi/pulumi'
+
+interface Args {
+    Name: string
+    Comment: string
+}
+
+// noinspection JSUnusedLocalSymbols
+function create(args: Args): aws.cloudfront.OriginAccessIdentity {
+    return new aws.cloudfront.OriginAccessIdentity(args.Name, {
+        comment: args.Comment,
+    })
+}
--- a/pkg/infra/iac2/templates/s3_bucket_policy/factory.ts
+++ b/pkg/infra/iac2/templates/s3_bucket_policy/factory.ts
@@ -9,7 +9,7 @@ interface Args {
 // noinspection JSUnusedLocalSymbols
 function create(args: Args): aws.s3.BucketPolicy {
     return new aws.s3.BucketPolicy(args.Name, {
-        bucket: args.Bucket.id, // refer to the bucket created earlier
+        bucket: args.Bucket.id,
         policy: args.Policy,
     })
 }
--- a/pkg/infra/iac2/templates/s3_bucket_policy/factory.ts
+++ b/pkg/infra/iac2/templates/s3_bucket_policy/factory.ts
@@ -1,0 +1,15 @@
+import * as aws from '@pulumi/aws'
+
+interface Args {
+    Name: string
+    Bucket: aws.s3.Bucket
+    Policy: aws.iam.PolicyDocument
+}
+
+// noinspection JSUnusedLocalSymbols
+function create(args: Args): aws.s3.BucketPolicy {
+    return new aws.s3.BucketPolicy(args.Name, {
+        bucket: args.Bucket.id, // refer to the bucket created earlier
+        policy: args.Policy,
+    })
+}
--- a/pkg/provider/aws/resources/cloudfront_test.go
+++ b/pkg/provider/aws/resources/cloudfront_test.go
@@ -1,0 +1,94 @@
+package resources
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/klothoplatform/klotho/pkg/core"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_CreateS3Origin(t *testing.T) {
+
+	unit := &core.StaticUnit{AnnotationKey: core.AnnotationKey{ID: "test"}}
+	bucket := NewS3Bucket(&core.Fs{AnnotationKey: core.AnnotationKey{ID: "bucket"}}, "test")
+	distro := NewCloudfrontDistribution("test", "1")
+
+	assert := assert.New(t)
+	dag := core.NewResourceGraph()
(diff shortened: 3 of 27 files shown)


pkg/infra/iac2/templates/origin_access_identity/factory.ts
```diff
@@ -0,0 +1,13 @@
...
+
+interface Args {
+    Name: string
+    Comment: string
+}
+
+// noinspection JSUnusedLocalSymbols
+function create(args: Args): aws.cloudfront.OriginAccessIdentity {
+    return new aws.cloudfront.OriginAccessIdentity(args.Name, {
+        comment: args.Comment,
```
yuval-klotho, line 11: Should we add `// comment`?

(No actual need, I just wanted to add a comment to add a comment to the comment Comment.)

pkg/infra/iac2/templates/s3_bucket_policy/factory.ts
```diff
@@ -0,0 +1,15 @@
...
+interface Args {
+    Name: string
+    Bucket: aws.s3.Bucket
+    Policy: aws.iam.PolicyDocument
+}
+
+// noinspection JSUnusedLocalSymbols
+function create(args: Args): aws.s3.BucketPolicy {
+    return new aws.s3.BucketPolicy(args.Name, {
+        bucket: args.Bucket.id, // refer to the bucket created earlier
```
yuval-klotho, line 12: What is this comment trying to hint at, and who is it for —  us, or the user? The user will see this.

pkg/provider/aws/resources/cloudfront_test.go
```diff
@@ -0,0 +1,94 @@
...
+	if !assert.NotNil(oai) {
+		return
+	}
+	res := dag.GetResource(fmt.Sprintf("aws:s3_bucket_policy:%s-%s", bucket.Name, unit.ID))
+	if !assert.NotNil(res) {
+		return
+	}
+	bucketPolicy, ok := res.(*S3BucketPolicy)
+	if !ok {
+		assert.True(ok)
```
yuval-klotho, line 35: minor/cleanup:

```suggestion
	if !assert.True(ok) {
```

pkg/provider/aws/static_unit.go
```diff
@@ -16,6 +16,8 @@ func (a *AWS) GenerateStaticUnitResources(unit *core.StaticUnit, dag *core.Resou
 		object := resources.NewS3Object(bucket, filepath.Base(f.Path()), f.Path(), filepath.Join(unit.ID, f.Path()))
 		dag.AddResource(object)
 		dag.AddDependency2(object, bucket)
+		a.MapResourceDirectlyToConstruct(object, unit)
 	}
+	a.MapResourceDirectlyToConstruct(bucket, unit)
```
yuval-klotho, line 21: hm, _does_ the bucket come directly from the static unit, though? It feels like either the bucket or the objects should, but not both.

I'm not really sure about that, though. It's a bit strange: the annotation is file/object-based, but the yaml.config is bucket-based (ish).

To ground my question a bit more, I guess I'll ask: what do we need these for? Is it just general graph hygiene, or a more concrete issue it's addressing?

pkg/infra/iac2/templates/s3_bucket_policy/factory.ts
jhsinger-klotho: let me remove that, was probably in deploylib and i missed removing it

pkg/provider/aws/resources/cloudfront_test.go
jhsinger-klotho: smart

pkg/provider/aws/static_unit.go
jhsinger-klotho: yeah right now each static unit generates their own bucket and the objects from it. we need the bucket for wiring up cloudfront. 

Im a little confused on the questions here to be honest if you dont mind elaborating. Static unit is the bucket and the contents within that bucket

yuval-klotho: My question is about the "directly" aspect. I can think of three ways to think of how the bucket and the objects relate to the construct:

1. The construct directly implies the objects; the bucket is there as a supporting resource for those objects
2. The construct directly implies the bucket; the objects are there as an implementation detail of the bucket
3. The construct directly implies both the objects and the bucket.

Of those, this code is option 3. But options 1 and 2 make sense to me, too.

(Option 1 has the most UX impact: it means that if the user wrote a `@klotho::static_unit` that ended up including zero files, that we would then elide away the bucket, since it's no longer needed to support the non-existent objects.)

The DAG will already say that the objects depend on the bucket, so we'll be able to get from the construct to the objects regardless of the approach. My only question is the "directly" part, and I'm only really asking it because you mentioned that we should be really careful about which resources we map directly to constructs.

What I meant by this:

> To ground my question a bit more, I guess I'll ask: what do we need these for? Is it just general graph hygiene, or a more concrete issue it's addressing?

... is to ask basically what would happen if we didn't have these `MapResourceDirectlyToConstruct` calls, from a UX perspective. That might help us pick which of the three options we want.

Again, I'm only asking because you had warned against being too liberal with these calls. If it's not a problem to include them, I'm fine with it. Seems like a 2-way door either way, right?

jhsinger-klotho: yeah seems like a 2 way door, i see your point. In my opinion the annotation denotes the bucket and the objects are a description of the contents within the bucket. We really care mainly about the bucket in the case of things being mapped since thats what we need to retrieve for configuration on the static unit annotation itself. Im fine removing it, the reason i did both is because those objects can only ever belong to the single static unit annotation as well, so to me they are also a direct mapping to the annotation

yuval-klotho: >  those objects can only ever belong to the single static unit annotation as well

by that reasoning, a lambda's execution role would be directly mapped to the `@klotho::execution_unit` annotation, no?

yuval-klotho (approved):

pkg/provider/aws/static_unit.go
jhsinger-klotho: yeah i think thats reasonable

Issue #476, opened 2023-04-06
Panic in expose (both Go and JS)


gordon-klotho:
Stack trace:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x14a5e4c]

goroutine 1 [running]:
github.com/klothoplatform/klotho/pkg/provider/aws/resources.(*ApiResource).Id(0x0)
        /home/gordon/workspace/klotho/pkg/provider/aws/resources/api_gateway.go:132 +0x6c
github.com/klothoplatform/klotho/pkg/infra/iac2.TemplatesCompiler.getVarName({0xc000535340?, 0xc00021c0c0?, 0xc00044a480?, 0xc00044a4b0?}, {0x2f806f0, 0x0})
        /home/gordon/workspace/klotho/pkg/infra/iac2/templates_compiler.go:500 +0x69
github.com/klothoplatform/klotho/pkg/infra/iac2.TemplatesCompiler.handleIaCValue({0xc000535340?, 0xc00021c0c0?, 0xc00044a480?, 0xc00044a4b0?}, {{0x2f806c0?, 0xc0000f4240?}, {0x2c63f29?, 0x15?}}, 0xc000a2e410)
        /home/gordon/workspace/klotho/pkg/infra/iac2/templates_compiler.go:485 +0x10de
github.com/klothoplatform/klotho/pkg/infra/iac2.TemplatesCompiler.resolveStructInput({0xc000535340?, 0xc00021c0c0?, 0xc00044a480?, 0xc00044a4b0?}, 0xc000a2e3e0, {0x2a2ff80?, 0xc00056a0b0?, 0x1?}, 0x1?, {0x0, ...}, ...)
        /home/gordon/workspace/klotho/pkg/infra/iac2/templates_compiler.go:308 +0x545
github.com/klothoplatform/klotho/pkg/infra/iac2.TemplatesCompiler.renderResource({0xc000535340, 0xc00021c0c0, 0xc00044a480, 0xc00044a4b0}, {0x2f70e60, 0xc00044a3f0}, {0x2f809f0?, 0xc00056a070})
        /home/gordon/workspace/klotho/pkg/infra/iac2/templates_compiler.go:226 +0x807
[...]

PR #478, opened 2023-04-06
fix permission source arn

jhsinger-klotho:

addresses https://github.com/klothoplatform/klotho/issues/476, was missing the end of the logic in https://github.com/klothoplatform/klotho/blob/main/pkg/infra/pulumi_aws/iac/api_gateway.ts#L492 for if the parent was nil

### Standard checks

### DIFF
--- a/pkg/infra/iac2/templates_compiler.go
+++ b/pkg/infra/iac2/templates_compiler.go
@@ -481,8 +481,12 @@ func (tc TemplatesCompiler) handleIaCValue(v core.IaCValue, appliedOutputs *[]Ap
 		}
 		accountId := resources.NewAccountId()
 		region := resources.NewRegion()
-		return fmt.Sprintf("pulumi.interpolate`arn:aws:execute-api:${%s.name}:${%s.accountId}:${%s.id}/*/%s${%s.path}`", tc.getVarName(region),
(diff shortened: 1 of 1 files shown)


pkg/infra/iac2/templates_compiler.go
```diff
@@ -481,8 +481,12 @@ func (tc TemplatesCompiler) handleIaCValue(v core.IaCValue, appliedOutputs *[]Ap
 		}
 		accountId := resources.NewAccountId()
 		region := resources.NewRegion()
-		return fmt.Sprintf("pulumi.interpolate`arn:aws:execute-api:${%s.name}:${%s.accountId}:${%s.id}/*/%s${%s.path}`", tc.getVarName(region),
-			tc.getVarName(accountId), tc.getVarName(method.RestApi), verb, tc.getVarName(method.Resource)), nil
+		path := "/"
+		if method.Resource != nil {
+			path = fmt.Sprintf("{%s.path}", tc.getVarName(method.Resource))
```
gordon-klotho, line 486: Should this be `"${`?

jhsinger-klotho: yup good catch

gordon-klotho (approved):

Issue #419, opened 2023-03-28
Kubernetes plugin work 


jhsinger-klotho:
We want the plugin to offer the configurability we offer in KubernetesTypeParams.

This includes things like configuring scaling limits, replicas, etc (anything that is purely K8s specific (Not NodeConstraints))

Definition of Done:
- All KubernetesTypeParams fields are supported via klotho.yaml (except NodeConstraints) and will be output in their corresponding yaml files

PR #479, opened 2023-04-06
k8s configuration: for cpu and memory limits

yuval-klotho:
Write cpu limits to k8s the manifest, both for deployments and for pods.

This includes a bit of refactoring work, to combine the "make sure there's no more than 1 container, and then configure it" that both deploy and pod were doing. That approach lets us test that configuration separately from the spec that's being configured, which means we don't need to do the full combination of all possible inputs for both deploy and pod.

In service of #419.

### Standard checks

- **Unit tests**: added test; also, note the use of yamlpath I added, to help shorten the test spec and make it easier for humans to focus in on the specific thing we're trying ot test
- **Docs**: n/a

### DIFF
--- a/pkg/config/execution_unit.go
+++ b/pkg/config/execution_unit.go
@@ -47,10 +47,10 @@ type (
 
 	// KubernetesLimits represents the configurability of kubernetes limits for execution units which match the kubernetes compatibility
 	KubernetesLimits struct {
-		// Cpu specifies the limit per pod in millicores
-		Cpu int `json:"cpu,omitempty" yaml:"cpu,omitempty" toml:"cpu,omitempty"`
+		// Cpu specifies the limit per pod in millicores. It is "any" so that the user can specify it as either a string or a number
+		Cpu any `json:"cpu,omitempty" yaml:"cpu,omitempty" toml:"cpu,omitempty"`
 		// Memory specifies the limit per pod in MB
-		Memory int `json:"memory,omitempty" yaml:"memory,omitempty" toml:"memory,omitempty"`
+		Memory any `json:"memory,omitempty" yaml:"memory,omitempty" toml:"memory,omitempty"`
 	}
 
 	// KubernetesLimits represents the configurability of kubernetes limits for execution units which match the kubernetes compatibility
--- a/pkg/config/execution_unit.go
+++ b/pkg/config/execution_unit.go
@@ -55,7 +55,7 @@ type (
 
 	// KubernetesLimits represents the configurability of kubernetes limits for execution units which match the kubernetes compatibility
 	KubernetesHorizontalPodAutoScalingConfig struct {
-		// MemoryUtilization specifies he percentage of cpu a pod can utilize before the cluster will attempt to scale the pod
+		// MemoryUtilization specifies the percentage of cpu a pod can utilize before the cluster will attempt to scale the pod
 		CpuUtilization int `json:"cpu_utilization,omitempty" yaml:"cpu_utilization,omitempty" toml:"cpu_utilization,omitempty"`
 		// MemoryUtilization specifies the percentage of memory a pod can utilize before the cluster will attempt to scale the pod
 		MemoryUtilization int `json:"memory_utilization,omitempty" yaml:"memory_utilization,omitempty" toml:"memory_utilization,omitempty"`
--- a/pkg/infra/kubernetes/exec_unit.go
+++ b/pkg/infra/kubernetes/exec_unit.go
@@ -167,13 +167,6 @@ func (unit *HelmExecUnit) transformDeployment(cfg config.ExecutionUnit) ([]HelmC
 	return values, nil
 }
 
-func mapOrNew[K comparable, V any](input map[K]V) map[K]V {
-	if input == nil {
-		input = make(map[K]V)
-	}
-	return input
-}
-
 func (unit *HelmExecUnit) transformService() (values []HelmChartValue, err error) {
 	log := zap.L().Sugar().With(logging.FileField(unit.Service), zap.String("unit", unit.Name))
 	log.Debugf("Transforming file, %s, for exec unit, %s", unit.Service.Path(), unit.Name)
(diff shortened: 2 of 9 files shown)


pkg/infra/kubernetes/exec_unit.go
```diff
@@ -165,6 +169,13 @@ func (unit *HelmExecUnit) transformDeployment() ([]Value, error) {
 	deployment.Spec.Selector.MatchLabels["execUnit"] = unit.Name
 
 	deployment.Spec.Template.Spec.ServiceAccountName = unit.getServiceAccountName()
+
+	k8sCfg := cfg.GetExecutionUnitParamsAsKubernetes()
+	if k8sCfg.Limits.Cpu != 0 {
+		container.Resources.Limits = mapOrNew(container.Resources.Limits)
+		container.Resources.Limits[corev1.ResourceCPU] = *k8s_resource.NewDecimalQuantity(*inf.NewDec(int64(k8sCfg.Limits.Cpu), 0), k8s_resource.DecimalSI)
```
jhsinger-klotho, line 176: just a note, cpu limit is in millicores so we should make sure we align config with that

pkg/infra/kubernetes/exec_unit.go
```diff
@@ -122,7 +125,7 @@ func (unit *HelmExecUnit) transformPod() (values []Value, err error) {
 	return
 }
 
-func (unit *HelmExecUnit) transformDeployment() ([]Value, error) {
+func (unit *HelmExecUnit) transformDeployment(cfg config.ExecutionUnit) ([]Value, error) {
```
jhsinger-klotho, line 128: well need to make the same modification in transformPod because if they specify a pod manifest we wont add a deployment manifest

pkg/infra/kubernetes/exec_unit.go
yuval-klotho: er, I'm not sure what to do with that statement

jhsinger-klotho: im guessing your statement converts to millicores, just not sure exactly what that statement does. but just saying our config is an int so 1 isnt 1 cpu its .0001 cpu or whatever

yuval-klotho: I think it just takes whatever int was in the yaml, and writes it in decimal format — so it shouldn't change anything. But I'll see if there are other formats (like unit suffixes) that k8s allows, and add tests for them if so.

yuval-klotho: I ended up going a different route: I just create some yaml, and then k8s's yaml library to turn it into the ResourceRequirements.

pkg/config/execution_unit.go
```diff
@@ -46,15 +46,15 @@ type (
 
 	// KubernetesLimits represents the configurability of kubernetes limits for execution units which match the kubernetes compatibility
 	KubernetesLimits struct {
-		// Cpu specifies the limit per pod in millicores
-		Cpu int `json:"cpu,omitempty" yaml:"cpu,omitempty" toml:"cpu,omitempty"`
+		// Cpu specifies the limit per pod in millicores. It is "any" so that the user can specify it as either a string or a number
+		Cpu any `json:"cpu,omitempty" yaml:"cpu,omitempty" toml:"cpu,omitempty"`
 		// Memory specifies the limit per pod in MB
-		Memory int `json:"memory,omitempty" yaml:"memory,omitempty" toml:"memory,omitempty"`
+		Memory any `json:"memory,omitempty" yaml:"memory,omitempty" toml:"memory,omitempty"`
```
jhsinger-klotho, line 52: why any instead of int?

yuval-klotho: Whoops, forgot to copy the comments from a couple lines above. But it's so that they can do either an int or string. For example, `memory: 128M` is valid, and parses to a string. But you have to allow numbers too, because `memory: 128` is also valid and parses to int.

jhsinger-klotho: im wondering if we force either a string or an int though to make it easier for us, theyll still be able to understand the values

pkg/infra/kubernetes/exec_unit.go
```diff
@@ -443,3 +428,62 @@ func (unit *HelmExecUnit) addEnvVarToPod(envVars core.EnvironmentVariables) ([]H
...
+
+func (unit *HelmExecUnit) configureContainer(container *corev1.Container, cfg config.ExecutionUnit) error {
+	k8sCfg := cfg.GetExecutionUnitParamsAsKubernetes()
+
+	limits := make(map[corev1.ResourceName]any)
+	if k8sCfg.Limits.Cpu != nil {
+		limits[corev1.ResourceCPU] = k8sCfg.Limits.Cpu
+	}
+	if k8sCfg.Limits.Memory != nil {
+		limits[corev1.ResourceMemory] = k8sCfg.Limits.Memory
```
jhsinger-klotho, line 471: in order for limits to actually do anything, we need to set the requests, otherwise it will never scale. We dont surface requests in config because we assume limits are 100% of requests for now so we can try to fully utilize nodes

jhsinger-klotho (approved):

pkg/config/execution_unit.go
yuval-klotho: Imo, given that all we do here is pass the value through to k8s (which can handle it as both string and int), the overriding concern should be the user experience — which seems significantly better with this as an `any`. My understanding is that this is used to read their input, right? So it's not enough that the user can understand it, they also have to be able to input the values they want: whether that's `cpu: 1` (an int), `cpu: 0.5` (a float),  or `cpu: 100m` (a string).

Unless I keep these as `any`, various `Test_configureContainer` cases fail.

jhsinger-klotho: alright, as long as that doesnt break our casting to kubernetes type params from the generic params then thats fine

yuval-klotho: Ah, I see. Yeah it doesn't matter for the casting: our method is to json-ify the generic InfraParams and then de-json-ify that back as KubernetesTypesParams, so `any` works totally fine (and is in fact the type of the incoming InfraParams, which is an `InfraParams map[string]any`).

jhsinger-klotho (approved):

PR #480, opened 2023-04-07
rename AddDependency2 -> AddDependency

yuval-klotho:
Rename `AddDependency2` back to `AddDependency`, and rename its parameters to make it a bit clearer which goes where.

### Standard checks

- **Unit tests**: n/a
- **Docs**: added godocs


pkg/core/resource_graph.go
```diff
@@ -26,14 +26,19 @@ func (rg *ResourceGraph) AddResource(resource Resource) {
...
-func (rg *ResourceGraph) AddDependency2(source Resource, dest Resource) {
-	for _, res := range []Resource{source, dest} {
+// Adds a dependency such that `deployedFirst` has to be deployed before `deployedNext`. For example, if you have a
+// Lambda and its execution role, then:
+//
+//	╭───────────────╮   ╭────────────────╮
+//	│     first     ├──➤│      next      │
+//	│    IamRole    │   │ LambdaFunction │
+//	╰───────────────╯   ╰────────────────╯
+func (rg *ResourceGraph) AddDependency(deployedFirst Resource, deployedNext Resource) {
```
gordon-klotho, line 36: This is flipping the order again.
`AddDependency(lambda, role)` is how it's used and results in the arrow `lambda -> role`

gordon-klotho (changes requested):

pkg/core/resource_graph.go
yuval-klotho: Oh, I misread the `dag.AddDependency(role, resource)` in `execution_unit.go` (I had just opened it in the search preview, and didn't look to see what `resource` was — I assumed it was the lambda, but it's not.)

gordon-klotho: The main thing I use for my mental model is the reflection-based one. All the edges are `dag.AddDependency(theStruct, theField)`

yuval-klotho: I like that mental model. I just pushed a change to add it to the godocs (didn't use the reflection aspect, but the struct-and-field).

pkg/core/resource_graph.go
```diff
@@ -26,14 +26,30 @@ func (rg *ResourceGraph) AddResource(resource Resource) {
...
+// association consistent with our visualizer, and with the Go struct graph.
+//
+// For example, if you have a Lambda and its execution role, then:
+//
+//	╭────────────────╮   ╭────────────────╮
+//	│ LambdaFunction ├──➤│    IamRole     │
+//	│ deployedSecond │   │ deployedFirst  │
+//	╰────────────────╯   ╰────────────────╯
+//
+// And you would invoke it as:
```
gordon-klotho, line 39:
```suggestion
// And you would construct it as:
```

nit: "invoke" doesn't seem quite the right word to me

gordon-klotho (approved):

pkg/core/resource_graph.go
yuval-klotho: ah, good catch. ("invoke" was from when I just had the method call, without the struct). What do you think about "use" (since the method call is still in the snippet, so "construct" doesn't quite work either)?

Issue #484, opened 2023-04-07
Dont override outputted files if they already exist, for specific files


jhsinger-klotho:
We want to make sure things like secrets and pulumi config stay static and we dont replace them on each compile

Issue #482, opened 2023-04-07
write rds username and password to pulumi config and use from there


jhsinger-klotho:
Today we used the pulumi config, but it seems as if we are moving away from needing to set anything in that config. We can set randomly generated username and passwords in our provider since it will automatically be wired up, but we still need it to show up in pulumi and we likely dont want it hardcoded

if we set in pulumi config still and they use the pregenerated stack yaml it will work otherwise they would have to run pulumi config set again.

for now we just have randomly generated strings which are passed down into the index.ts file

PR #485, opened 2023-04-07
start of rds

jhsinger-klotho:

```
 +   ├─ aws:rds:SubnetGroup                py-orm-sqllchemy                               created (3s)
 +   ├─ aws:rds:Proxy                      py-orm-sqllchemy                               created (230s)
 +   ├─ aws:rds:Instance                   py-orm-sql-lchemy                              created (226s)
```

This contains all the RDS functionality:
- Create a instance, proxy, subnet group, proxyTarget IR and templates
- function to create all resources and dependencies (including secrets manager secret for the proxy)
- Small fixes to some of the api gw factories found from running py-orm

New functionality:
- Change the outputting of the compilation document. I saw the output files dont persist outside of thecompile function since the array is not a pointer so we can fix that if needed
- Instead of relying on pulumi config for secrets, we will generate a random username and password and write it to a file in the compiled dir. This file can be changed then by the user and then will be read by our pulumi code. Part of the reasoning for doing this is because our secrets manager factory.ts files only take file paths, so we had no other mechanism to create the secret. cut https://github.com/klothoplatform/klotho/issues/482 to further enhance this.

This also made me realize we always overwrite so cut https://github.com/klothoplatform/klotho/issues/484 to track that.

One note is aws-native either requires you to set config, or requires you to create your own provider. We will always see 
```
    error: missing required configuration key "aws-native:region":The region where AWS operations will take place. Examples are eu-east-1, eu-west-2, etc.
    	Set a value using the command `pulumi config set aws-native:region <value>`
```
So it may be time for us to think about bringing region forward so that we dont have different regions for aws native and aws.

### Standard checks

- **Unit tests**: Any special considerations? Need to add and will update the pr


pkg/compiler/compilation_document.go
```diff
@@ -59,3 +64,86 @@ func (doc *CompilationDocument) OutputTo(dest string) error {
...
+
+	return
+}
+
+func (document *CompilationDocument) OutputHelpers(outDir string) error {
+	var merr multierr.Error
+	for _, resource := range document.Resources.ListResources() {
+		if res, ok := resource.(core.HasOutputFiles); ok {
+			document.OutputFiles = append(document.OutputFiles, res.GetOutputFiles()...)
+		}
```
yuval-klotho, line 137: 👀 note to self (& other reviewers if interested): this is new in this copy of the code, relative to the `-` above)

```diff
--- 	for _, resource := range dag.ListResources() {
+++	for _, resource := range document.Resources.ListResources() {
+++		if res, ok := resource.(core.HasOutputFiles); ok {
+++			document.OutputFiles = append(document.OutputFiles, res.GetOutputFiles()...)
+++		}
```

pkg/compiler/compilation_document.go
```diff
@@ -59,3 +64,86 @@ func (doc *CompilationDocument) OutputTo(dest string) error {
...
+
+	merr.Append(document.OutputHelpers(outDir))
+
+	merr.Append(document.OutputTo(document.Configuration.OutDir))
+	err = merr.ErrOrNil()
+
+	return
+}
+
+func (document *CompilationDocument) OutputHelpers(outDir string) error {
```
yuval-klotho, line 132: 👀 note to self (& other reviewers if interested): this is new in this copy of the code, relative to the `-` above)

```diff
--- 	func OutputHelpers(dag *core.ResourceGraph, outDir string) error {
+++	func (document *CompilationDocument) OutputHelpers(outDir string) error {
```

pkg/compiler/compilation_document.go
```diff
@@ -59,3 +64,86 @@ func (doc *CompilationDocument) OutputTo(dest string) error {
...
+		enc.SetIndent("", "  ")
+		err = enc.Encode(resourcesOutput)
+		if err != nil {
+			merr.Append(errors.Wrap(err, "error writing resources"))
+		}
+	}
+
+	merr.Append(document.OutputHelpers(outDir))
+
+	merr.Append(document.OutputTo(document.Configuration.OutDir))
```
yuval-klotho, line 126: 👀 note to self (& other reviewers if interested): this is new in this copy of the code, relative to the `-` above)

```diff
--- 	merr.Append(OutputHelpers(document.Resources, outDir))
+++	merr.Append(document.OutputHelpers(outDir))
+++
+++	merr.Append(document.OutputTo(document.Configuration.OutDir))
```

pkg/infra/iac2/templates/api_method/factory.ts
```diff
@@ -13,7 +13,11 @@ interface Args {
 function create(args: Args): aws.apigateway.Method {
     return new aws.apigateway.Method(args.Name, {
         restApi: args.RestApi.id,
+        //TMPL {{- if .Resource.Raw }}
         resourceId: args.Resource.id,
+        //TMPL {{- else}}
+        //TMPL resourceId: args.RestApi.rootResourceId,
+        //TMPL {{- end }}
```
yuval-klotho, line 20: :-\ It's a shame we have to do this. It's the second or third time we've had to, right? I wonder if it means we should consider coming up with a mechanism to make it nicer.

For example, I could see something where we define a custom interface for "resourceId provider", or something-something. idk.

We definitely don't need to do anything about it now, but do you think it's worth thinking about? It may be worth thinking about for a bit even if we don't intend to implement it, even if only to write a technical blog.

pkg/infra/iac2/templates/rds_instance/factory.ts
```diff
@@ -0,0 +1,38 @@
...
+function create(args: Args): aws.rds.Instance {
+    return new aws.rds.Instance(
+        args.Name,
+        {
+            instanceClass: args.InstanceClass,
+            engine: args.Engine,
+            engineVersion: args.EngineVersion,
+            dbName: args.DatabaseName,
+            username: fs.readFileSync(args.CredentialsPath, 'utf-8').split('\n')[1].split('"')[3],
+            password: fs.readFileSync(args.CredentialsPath, 'utf-8').split('\n')[2].split('"')[3],
```
yuval-klotho, line 29: I wonder, will ts let us use `...` notation to turn this into an invocation of an anonymous function? something like:

```
/* ... */
    dbName: args.DatabaseName,
    ... () => { // literally the dots
      const fileContents = fs.readFileSync(args.CredentialsPath, 'utf-8');
      /* ... */
      return usernameContents, passwordContents;
    }(),
},
```

pkg/infra/iac2/templates_compiler.go
```diff
@@ -464,6 +464,28 @@ func (tc TemplatesCompiler) handleIaCValue(v core.IaCValue, appliedOutputs *[]Ap
...
+			for _, resource := range downResources {
+				if rdsProxyTargetGroup, ok := resource.Source.(*resources.RdsProxyTargetGroup); ok {
+					instance = rdsProxyTargetGroup.RdsInstance
+				}
+			}
+			if instance == nil {
+				return "", errors.Errorf("Rds Proxy, %s, must have an associated instance", v.Resource.Id())
+			}
+
+			fetchUsername := fmt.Sprintf(`fs.readFileSync('%s', 'utf-8').split("\n")[1].split('"')[3]`, instance.CredentialsPath)
```
yuval-klotho, line 481: It's kinda wonky, imo, that we have some of these `readFileSync`s within a factory.ts, and some righthere. And generally speaking, imo the less ts logic goes directly in go, the better.

I think this is starting to get at a "virtual resource" that I've mentioned before: a `core.Resource` that the IaC injects into the DAG, in a pass before the rendering. I know we've discussed it, but I can't remember if we're doing it yet.

In this case, it would be to identify whether we need an `RdsConnectionInfo` resource (which would be a struct within the iac package, which implements `core.Resource`). If we do, we create it — and it has a normal factory.ts that does the read, and returns a type of `{username: string, password: string}` (we could even do it as an async read, and return a `Promise<{username: string, password: string}>`).

pkg/provider/aws/persist.go
```diff
@@ -9,19 +9,19 @@ func (a *AWS) GenerateFsResources(construct *core.Fs, result *core.ConstructGrap
 	bucket := resources.NewS3Bucket(construct, a.Config.AppName)
 	dag.AddResource(bucket)
 	a.MapResourceDirectlyToConstruct(bucket, construct)
+	actions := []string{"s3:*"}
```
yuval-klotho, line 12: Did this move so that we don't create multiple of the same bucket, in multi-exec? And did you catch it via any tests, or just inspection?

pkg/provider/aws/resources/rds.go
```diff
@@ -0,0 +1,327 @@
...
+// generateUsername generates a random username for the rds instance.
+//
+// The first letter of an RDS username must be a letter
+func generateUsername() string {
+	rand.Seed(time.Now().UnixNano())
+	initialChars := []rune("ABCDEFGHIJKLMNOPQRSTUVWXYZ" +
+		"abcdefghijklmnopqrstuvwxyz")
+	chars := []rune("ABCDEFGHIJKLMNOPQRSTUVWXYZ" +
+		"abcdefghijklmnopqrstuvwxyz" +
+		"0123456789")
```
yuval-klotho, line 180: minor:

```suggestion
	initialChars := []rune("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz")
	chars := []rune("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789")
```

(83-char width is fine, imo)

pkg/provider/aws/resources/rds.go
```diff
@@ -0,0 +1,327 @@
...
+		if res, ok := dep.Source.(*IamPolicy); ok {
+			if len(res.Policy.Statement[0].Action) == 1 && res.Policy.Statement[0].Action[0] == "rds-db:connect" {
+				return res
+			}
+		}
+	}
+	return pol
+}
+
+// generateUsername generates a random username for the rds instance.
```
yuval-klotho, line 171: Do we want a random username? Password, sure — but username feels like it should be human-readable, or at least understandable if someone looks at it in an admin console. Something like `klotho-generated-${random}`

pkg/provider/aws/resources/rds.go
```diff
@@ -0,0 +1,327 @@
...
+	b.WriteRune(chars[rand.Intn(len(initialChars))])
+	for i := 0; i < length-1; i++ {
+		b.WriteRune(chars[rand.Intn(len(chars))])
+	}
+	return b.String()
+}
+
+// generatePassword generates a random password for the rds instance.
+func generatePassword() string {
+	rand.Seed(time.Now().UnixNano())
```
yuval-klotho, line 192: from [`rand` godocs](https://pkg.go.dev/math/rand):

> For random numbers suitable for security-sensitive work, see the crypto/rand package

pkg/provider/aws/resources/rds.go
```diff
@@ -0,0 +1,327 @@
...
+		SecurityGroups:                   securityGroups,
+		IamDatabaseAuthenticationEnabled: true,
+		DatabaseName:                     orm.ID,
+		Username:                         generateUsername(),
+		Password:                         generatePassword(),
+		Engine:                           "postgres",
+		EngineVersion:                    "13.7",
+		InstanceClass:                    "db.t4g.micro",
+		SkipFinalSnapshot:                true,
+		AllocatedStorage:                 20,
```
yuval-klotho, line 218: Some of these feel like they should be configurable (e.g. `InstanceClass`), and pretty all of them (from `DatabaseName` down) feel like they belong in the IaC, not provider.

I know that means the IaC would need to handle `RdsInstance` specially, but I think that makes sense to do.

It may be useful to create a method on the graph, `replaceItem(V old, V new)` which replaces the vertex while keeping all the edges (having them instead point to the new item). So the iac plugin would take an `RdsInstance`, transform it into a `RenderableRdsInstance`, and replace that in the graph.

Not an absolute blocker, but if we don't do it now then I'd like at least a ticket to remind us to revisit it later, time permitting.

pkg/infra/iac2/templates/rds_instance/factory.ts
jhsinger-klotho: so i had tried to do something like that, but couldnt get it to be happy in our formatting/unit tests. I can keep trying though

pkg/infra/iac2/templates_compiler.go
jhsinger-klotho: yeah i mean thats the other route we could go, is to create some resource to move this out, but we dont do that yet. Would you be ok with me cutting a ticket, specifically calling this one out and moving that to fit and finish? Since this works im wondering if it could be a checkpoint, but for now because this has to be in sync with the env var, and the secret i dont want to make this pr too much larger

pkg/provider/aws/persist.go
jhsinger-klotho: before we were creating the policy for every single exec unit, so i moved it out so we create the policy once and attach it to the exec units roles that need it

pkg/provider/aws/resources/rds.go
jhsinger-klotho: are you suggesting we use another package? im not sure what more we would get out of that when were just generating a 12 char string

pkg/provider/aws/resources/rds.go
jhsinger-klotho: yeah they should be, we just dont expose them yet. And why do you think they belong in the IaC? If they go in the klotho configuration then it would be handled by the provider and thats our layer of configuratbility.

We will need the replacement method at some point, i agree, but im not sure that i understand the reasoning for rds to follow a different flow than everything else

pkg/infra/iac2/templates/api_method/factory.ts
jhsinger-klotho: its just in the apigatewayv1 stuff where we have had to do this so far. I dont think its worth investing in quite yet IMO to make thsi better unless we really start running into a lot of scenarios

pkg/provider/aws/resources/rds.go
jhsinger-klotho: i dont know that i entirely agree. First we cant use - in the username. Also im just not sold that it has to be prefixed with something about us but we could do that

pkg/provider/aws/resources/rds.go
yuval-klotho: eh, yeah maybe it's not. idk, these just feel lower level. But maybe I just need to adjust my mental model a bit.

jhsinger-klotho: yeah i mean these could all affect things like cost, throughput, storage, etc which is information we will want to have within the provider to optimize later. Anything within the IaC is also static so if we want any configuratibility we cannot put these in the IaC. I realize they are static now but the intention is that they wont be forever

pkg/provider/aws/resources/rds.go
jhsinger-klotho: they would need to know when klotho was run, not when the instance was created, which seems nearly impossible at the moment. But if its more secure, still good to switch it out so i can do that

yuval-klotho: Run logs might not be hard to find, either. Think about at Amazon — anyone internally could tell you when pretty much any build job happened, company-wide.

I think the "fix" is pretty easy, too: we could just use crypto/rand to populate a []byte, and then base64-encode that.

pkg/provider/aws/resources/rds.go
yuval-klotho: Ah, that all makes sense — good call.

yuval-klotho (approved):

Issue #458, opened 2023-04-04
Support amitype and node constraints for eks node groups


jhsinger-klotho:
To get parity in our EKS setup we want to be able to support our node constraints feature in configuration. 

This includes (definition of done)
 - Determining the number of node groups we need to launch by looking at the instance type + disk size combinations
 - Being able to set the amitype of the node group based on instance type
 - Being able to modify the deployment/pods node affinity to take in the node group like we do today so we can guarantee they will be launched on those sizes

PR #487, opened 2023-04-10
Support NodeSelectors and set up EKS node groups

gordon-klotho:
Still need to run some test compiles to confirm it's working, but should just be tweaks/fixes from here.

https://github.com/klothoplatform/klotho/issues/458

### Standard checks


pkg/infra/iac2/templates_compiler.go
```diff
@@ -192,61 +200,79 @@ func (tc TemplatesCompiler) renderResource(out io.Writer, resource core.Resource
...
+				if r == nil {
+					return
+				}
+				panic(errors.Errorf("panic rendering field %s: %v", fieldName, r))
+			}()
+			// dependsOn will be a reserved field for us to use to map dependencies. If specified as an Arg we will automatically call resolveDependencies
+			if fieldName == "dependsOn" {
+				inputArgs[fieldName] = stringTemplateValue{value: tc.resolveDependencies(resource)}
+				return
 			}
```
jhsinger-klotho, line 215: why make this change? Also wont this cause depends on and protect to be handled differently when they are both our reserved names and should be handled in the same manner?

pkg/infra/kubernetes/exec_unit.go
```diff
@@ -173,12 +173,27 @@ func (unit *HelmExecUnit) transformDeployment() ([]HelmChartValue, error) {
...
+			Type:         string(ExecUnitNetworkPlacement),
+			Key:          fmt.Sprintf("{{ .Values.%sNetworkPlacement }}", unit.Name),
+		},
+		HelmChartValue{
+			ExecUnitName: unit.Name,
+			Kind:         deployment.Kind,
+			Type:         string(ExecUnitNodeGroup),
+			Key:          fmt.Sprintf("{{ .Values.%sNodeGroup }}", unit.Name),
+		},
+	)
```
jhsinger-klotho, line 195: this value may not always exist. If the user doesnt specify any constrainst on an exec unit we dont need to care what node it gets launched on

pkg/infra/kubernetes/manifests/deployment.yaml.tmpl
```diff
@@ -25,4 +25,7 @@ spec:
       - image: REPLACE_ME
         name: {{ .ExecUnitName }}
       serviceAccount: {{ .ServiceAccountName }}
-      serviceAccountName: {{ .ServiceAccountName }}
\ No newline at end of file
+      serviceAccountName: {{ .ServiceAccountName }}
+      nodeSelector:
+        network_placement: REPLACE_ME
+        eks.amazonaws.com/nodegroup: REPLACE_ME
```
jhsinger-klotho, line 31: similarly here, i think its fine for network placement to be always put on for now, but this one should not exist on every manifest since we wont always have a node selector for node group

pkg/provider/aws/config.go
```diff
@@ -34,8 +34,9 @@ const (
 
 var (
 	eksDefaults = config.KubernetesTypeParams{
-		NodeType: "fargate",
-		Replicas: 2,
+		NodeType:     "fargate",
+		InstanceType: "t3.medium",
```
jhsinger-klotho, line 38: this combination doesnt quite make sense since you cant choose instance type on fargate. Not sure if it would actually make our logic have issues though, but its a bit confusing

pkg/provider/aws/resources/eks.go
```diff
@@ -65,43 +111,90 @@ type (
...
-	dag.AddDependency(cluster, clusterRole)
+	dag.AddDependenciesReflect(cluster)
+
+	for groupKey, spec := range groupSpecs {
+		groupName := NodeGroupName(groupKey.NetworkType, groupKey.InstanceType)
+		nodeGroup := &EksNodeGroup{
+			Name:          nodeGroupSanitizer.Apply(groupName),
+			ConstructsRef: spec.refs,
+			Cluster:       cluster,
+			DiskSize:      spec.DiskSizeGiB,
```
jhsinger-klotho, line 160: if this isnt set on any exec units, wouldnt we set this as 0 then?

pkg/infra/iac2/templates_compiler.go
gordon-klotho: The only thing this changed was to put it into a function with the `defer func()` to wrap panics with the relevant field name and resource Id. All the reflection stuff likes to panic, and the resulting message makes it hard to diagnose where the problem was. This is a band-aid solution (there's more robust / cleaner solutions), but good enough for now

pkg/provider/aws/config.go
gordon-klotho: Yeah, but there's no "per-nodetype default" configuration, so this is as good as we can do for now.

pkg/infra/iac2/templates_compiler.go
jhsinger-klotho: i see, why is depends on in here instead of the other portion or why isnt protect in here as well since thats also a reserved field for us

gordon-klotho: The code is the same as before except instead of `continue` it's `return`. Anything else is likely a merge/rebase error, I'll take another look at it.

pkg/infra/iac2/templates/eks_node_group/factory.ts
```diff
@@ -19,11 +19,11 @@ interface Args {
 // noinspection JSUnusedLocalSymbols
 function create(args: Args): aws.eks.NodeGroup {
     return new aws.eks.NodeGroup(args.Name, {
-        clusterName: args.Cluster.name,
+        clusterName: args.Cluster.name.apply((n) => n!),
```
jhsinger-klotho, line 22: why do you need to run apply here? I thought input can be pulumi output

gordon-klotho: `clusterName: pulumi.Input<string>`
whereas
`args.Cluster.name: pulumi.Output<string | undefined>`
😢

pkg/infra/iac2/templates_compiler.go
gordon-klotho: I don't think we need a full try/catch - these can be resolved by properly `Kind` checking or `IsNil` checking in various places. I didn't want to go through the current code with a fine-toothed comb at this point though. I think it originally came up when looking into https://github.com/klothoplatform/klotho/issues/476 (but I dropped the investigation and just commented out the expose and let Jordan fix it).

yuval-klotho: But you have implemented a full try-catch, just in a kinda weird way (and specifically, by doing it as a catch-try, which is what Go forced you to do).

gordon-klotho: I meant a generalized try/catch. The idiomatic Go thing would be to split `func(fieldName string)` into a real function and call it from within the loop, which can then more easily catch and handle any panics. Again, there's much better/cleaner ways to do this - and nothing panics right now so we don't _need_ it, but I figured it was better than nothing. I can revert this part if it's distracting from the other parts of the PR.

yuval-klotho: fwiw, I got distracted (😬) and looked into what a nice try-catch API might look like, cause it piqued my interest. This is what I came up with: https://gist.github.com/yuval-klotho/c8788725a8967b33464b4b7489e55cae

pkg/infra/iac2/templates/eks_node_group/factory.ts
jhsinger-klotho: i wonder why it can be undefined according to pulumi, but 🤷

pkg/provider/aws/resources/eks.go
jhsinger-klotho: im confused, i still think this may happen right?

yuval-klotho (commented):
Generally lgtm, but I'll let Jordan take a look as well since I don't know the k8s stuff as well.

jhsinger-klotho (commented):
can we see a rendered manifest and corresponding IaC Because i dont think this would quite work yet

pkg/provider/aws/execution_unit.go
```diff
@@ -107,21 +107,28 @@ func (a *AWS) GenerateExecUnitResources(unit *core.ExecutionUnit, result *core.C
...
 								khChart.Values[val.Key] = core.IaCValue{
 									Resource: role,
 									Property: core.ARN_IAC_VALUE,
 								}
 								dag.AddDependency(khChart, role)
-							case string(kubernetes.TargetGroupTransformation):
+							case kubernetes.InstanceTypeKey:
+								khChart.Values[val.Key] = core.IaCValue{
+									Property: "eks.amazonaws.com/nodegroup",
+								}
```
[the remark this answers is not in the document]
gordon-klotho: Yeah, I thought the k8s provider was supposed to be cloud agnostic - at least it seems that way right now. Wasn't too hard to add when the value needs to templated anyway.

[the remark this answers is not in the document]
jhsinger-klotho: yeah we havent split out aws specific logic yet, but i like starting to do that so fine by this

pkg/infra/kubernetes/exec_unit.go
```diff
@@ -150,6 +156,27 @@ func (unit *HelmExecUnit) transformDeployment(cfg config.ExecutionUnit) ([]HelmC
...
+		instanceTypeValue := fmt.Sprintf("{{ .Values.%sInstanceTypeValue }}", unit.Name)
+		deployment.Spec.Template.Spec.NodeSelector[instanceTypeKey] = instanceTypeValue
+		values = append(values,
+			HelmChartValue{
+				ExecUnitName: unit.Name,
+				Kind:         deployment.Kind,
+				Type:         string(InstanceTypeKey),
+				Key:          instanceTypeKey,
+			},
+		)
```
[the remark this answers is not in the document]
gordon-klotho: Yeah, I forgot to go back and add it when I added to EKS as a value. I wasn't sure whether the value could go directly in here like network_placement or not.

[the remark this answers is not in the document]
jhsinger-klotho: oh i guess it depends if we have that name at the time of writing this, i know we actually used pulumi output to set it so thats what i was expecting. If it works without that we could do it, but then you are modifying the manifest again rather than setting values on the pulumi helm chart, which seems odd to do outside of the kubernetes plugin

pkg/provider/aws/resources/eks_test.go
```diff
@@ -36,9 +44,9 @@ func Test_CreateEksCluster(t *testing.T) {
 					{Source: "aws:eks_fargate_profile:test-app-test-cluster", Destination: "aws:eks_cluster:test-app-test-cluster"},
 					{Source: "aws:eks_fargate_profile:test-app-test-cluster", Destination: "aws:iam_role:test-app-test-cluster-FargateExecutionRole"},
 					{Source: "aws:eks_fargate_profile:test-app-test-cluster", Destination: subnet.Id()},
-					{Source: "aws:eks_node_group:test-app-test-cluster", Destination: "aws:eks_cluster:test-app-test-cluster"},
-					{Source: "aws:eks_node_group:test-app-test-cluster", Destination: "aws:iam_role:test-app-test-cluster-NodeGroupRole"},
-					{Source: "aws:eks_node_group:test-app-test-cluster", Destination: subnet.Id()},
+					{Source: "aws:eks_node_group:private_t3.medium", Destination: "aws:eks_cluster:test-app-test-cluster"},
+					{Source: "aws:eks_node_group:private_t3.medium", Destination: "aws:iam_role:test-app-test-cluster.private_t3.medium"},
+					{Source: "aws:eks_node_group:private_t3.medium", Destination: subnet.Id()},
```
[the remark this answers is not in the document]
gordon-klotho: https://github.com/klothoplatform/klotho/blob/ifc2-foundations/pkg/sanitization/aws/eks.go#L24-L37

/shrug

pkg/infra/kubernetes/exec_unit.go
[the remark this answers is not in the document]
gordon-klotho: Added the value in the latest commit. 👍

jhsinger-klotho:
is the `mainInstanceTypeValue` from the recent post due to new updates or whats in here?

gordon-klotho:
> is the `mainInstanceTypeValue` from the recent post due to new updates or whats in here?

After the most recent commit

pkg/provider/aws/resources/eks_test.go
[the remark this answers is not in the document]
jhsinger-klotho: right im saying we need to update this. We never had node group sanitization in deploy lib, but looking at our ts code we do remove .s https://github.com/klothoplatform/klotho/blob/main/pkg/infra/pulumi_aws/iac/eks.ts#L539. So we need to update our sanitization here since before we only had a static name we never ran into it

jhsinger-klotho:
> > is the `mainInstanceTypeValue` from the recent post due to new updates or whats in here?
> 
> After the most recent commit

Maybe you havent pushed yet? I dont see a commit outside of the imports one

pkg/provider/aws/resources/eks.go
```diff
@@ -65,43 +111,89 @@ type (
...
+
+	for _, unit := range units {
+		unitCfg := cfg.GetExecutionUnit(unit.ID)
+		params := unitCfg.GetExecutionUnitParamsAsKubernetes()
+		key := groupKey{InstanceType: params.InstanceType, NetworkType: unitCfg.NetworkPlacement}
+		spec := groupSpecs[key]
+		if spec == nil {
+			spec = &groupSpec{}
+			groupSpecs[key] = spec
+		}
```
jhsinger-klotho, line 140: if someone doesnt set the params for instance type then do we generate a blank spec for it?

Similarly if a unit has a requirement on disk size, but not instance type will it be ignored and not factored into the existing groups?

jhsinger-klotho: may be helpful to break out the determination of node groups given a set of exec units so we can unit test it

pkg/provider/aws/config.go
jhsinger-klotho: So i think we actually dont want this default expose, but only internally in our plugin. the reasoning being, if someone sets an instanceType on one unit but not the other and that instance type isnt t3.medium, this would force us to create 2 node groups which would be wasteful. Instead we should know that because the unit2 doesnt have an instance type or disk space specified, it can launch on any node

pkg/provider/aws/resources/eks.go
gordon-klotho: At this point, after defaults have been applied, the instance type and disk size shouldn't be empty / zero.

jhsinger-klotho: disk size isnt in defaults it looks like but also this will cause us to create unnecessary node groups which could be wildly expensive.

pkg/infra/kubernetes/exec_unit_test.go
```diff
@@ -505,6 +506,75 @@ func Test_transformDeployment(t *testing.T) {
...
+      - image: '{{ .Values.testUnitImage }}'
+        name: nginx
+        resources: {}
+      nodeSelector:
+        '{{ .Values.testUnitInstanceTypeKey }}': '{{ .Values.testUnitInstanceTypeValue
+          }}'
+        network_placement: private
+      serviceAccountName: testUnit
+status: {}
+`, // ?? Not sure why yaml marshalling adds the newline and indentation within the value of the instance type
```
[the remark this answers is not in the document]
gordon-klotho: Yeah, it does - confirmed in a compilation (posted as a comment above)

pkg/provider/aws/resources/eks.go
gordon-klotho: Is `aws/config.go` not the right place to put the defaults?
```go
	eksDefaults = config.KubernetesTypeParams{
		NodeType:     "fargate",
		InstanceType: "t3.medium",
		DiskSizeGiB:  20,
		Replicas:     2,
	}
```

I took these from what the default node group was in the deploylib:
```ts
    private determineNodeGroupSpecs(execUnits: EksExecUnit[]): Map<string, NodeGroupSpecs> {
        let nodeGroupSpecs: Map<string, NodeGroupSpecs> = new Map<string, NodeGroupSpecs>()
        const defaultDiskSize = 20
        const defaultInstanceType = 't3.medium'
```

> this will cause us to create unnecessary node groups

Not sure I understand why, can you go into details on that?

> see my comment about why we didnt have defaults for more info

Which comment?

pkg/provider/aws/config.go
jhsinger-klotho: if we want to set this it would need to be at the cluster level now that i think about it

gordon-klotho: > if someone sets an instanceType on one unit but not the other and that instance type isnt t3.medium, this would force us to create 2 node groups which would be wasteful.

The same could be said about many other parameters. If the intention is to change the instance type for all units, we provide a mechanism for that via the `defaults`. I think this should behave as all the other parameters and not special cased.

jhsinger-klotho: its not special cased, its just a different scenario. If i was a user and i didnt specify anything, i wouldnt expect a node group to be created for a specific execution unit which didnt have needs for instance type or disk size because there are other node groups which satisfy its constraints.

I also just dont think we should be regressing in terms of optimizing architectures for cost

jhsinger-klotho: a node doesnt belong to a kubernetes exec unit, like it does to an rds instance, etc

pkg/provider/aws/resources/eks.go
```diff
@@ -65,43 +110,122 @@ type (
...
+			Labels: map[string]string{
+				"network_placement": groupKey.NetworkType,
+			},
+			// TODO make these configurable
+			DesiredSize:    2,
+			MaxSize:        2,
+			MinSize:        1,
+			MaxUnavailable: 1,
+		}
+		nodeGroup.NodeRole = createNodeRole(cfg.AppName, fmt.Sprintf("%s.%s", clusterName, nodeGroup.Name), spec.refs)
```
jhsinger-klotho, line 205: nit: i believe the cluster name already contains the app name so we might have a bit of a long redundant name here. Will still work though

pkg/provider/aws/execution_unit.go
```diff
@@ -107,21 +107,28 @@ func (a *AWS) GenerateExecUnitResources(unit *core.ExecutionUnit, result *core.C
...
+							case kubernetes.InstanceTypeValue:
+								khChart.Values[val.Key] = core.IaCValue{
+									Property: resources.NodeGroupNameFromConfig(cfg),
+								}
```
jhsinger-klotho, line 130: wouldnt this only be conditionally set? If instance type and disk size arent set on the config then we wouldnt want to place it on a node group. Same with the key

pkg/provider/aws/resources/eks.go
```diff
@@ -65,43 +110,122 @@ type (
...
+			break
+		}
+	}
+	if !hasInstanceType {
+		for gk, spec := range groupSpecs {
+			if gk.InstanceType == "" {
+				groupSpecs[groupKey{InstanceType: "t3.medium", NetworkType: gk.NetworkType}] = spec
+			}
+		}
+	}
```
jhsinger-klotho, line 183: this will still leave blank instance types in the group specs list, which i believe would cause failures if there is another node grup thats set

pkg/provider/aws/resources/eks.go
gordon-klotho: I tested this because I thought that too. `clusterName` doesn't have the app name in it, but `cluster.Name` does.

pkg/provider/aws/execution_unit.go
gordon-klotho: The Key and Value are only conditionally set in the provider values, so that is handled here: https://github.com/klothoplatform/klotho/pull/487/files#diff-9da19c3bee3a0431ac52dba95b3211cfd88e3a40eb4f484064ddf771f0b0a2abR166

pkg/provider/aws/resources/eks.go
```diff
@@ -135,23 +154,38 @@ func CreateEksCluster(cfg *config.Application, clusterName string, subnets []*Su
...
 
 	for groupKey, spec := range groupSpecs {
+		if groupKey.InstanceType == "" {
+			continue
+		}
```
jhsinger-klotho, line 188: the issue with having this check down here is it practically ignores the DiskSize param on an exec unit if it doesnt set the instance type

pkg/provider/aws/resources/eks_test.go
```diff
@@ -139,3 +139,91 @@ func TestNodeGroupNameFromConfig(t *testing.T) {
...
+			},
+			want: []NodeGroupExpect{
+				{Name: "private_c1_medium", DiskSize: 20, AmiType: "AL2_x86_64"},
+				{Name: "private_g2_medium", DiskSize: 20, AmiType: "AL2_x86_64_GPU"},
+			},
+		},
+		{
+			name: "disk size maximum",
+			units: map[string]*config.ExecutionUnit{
+				"a": {InfraParams: config.InfraParams{"disk_size_gib": 10}},
```
jhsinger-klotho, line 196: try switching this to 100 and see if it has an effect, we want to make sure if its larger than the default its obeyed

pkg/infra/kubernetes/exec_unit.go
```diff
@@ -150,6 +156,33 @@ func (unit *HelmExecUnit) transformDeployment(cfg config.ExecutionUnit) ([]HelmC
...
 	deployment.Spec.Template.Spec.ServiceAccountName = unit.getServiceAccountName()
 
+	if deployment.Spec.Template.Spec.NodeSelector == nil {
+		deployment.Spec.Template.Spec.NodeSelector = make(map[string]string)
+	}
+
+	if cfg.NetworkPlacement != "" {
+		deployment.Spec.Template.Spec.NodeSelector["network_placement"] = cfg.NetworkPlacement
+	}
+	if kconfig := cfg.GetExecutionUnitParamsAsKubernetes(); kconfig.InstanceType != "" {
```
jhsinger-klotho, line 166: if they dont put instance type but put DiskSizeGB how would we make the guarantee they would get enough disk space?

pkg/provider/aws/resources/eks.go
gordon-klotho: > this will still leave blank instance types in the group specs list, which i believe would cause failures if there is another node grup thats set

It shouldn't due to the `continue` here:
```go
	for groupKey, spec := range groupSpecs {
		if groupKey.InstanceType == "" {
			continue
		}
```
I could delete the key instead if you think that's clearer.

> might want to split this logic out so it can be unit tested, i think thatll help us understand what will get created

Split out more than `createNodeGroups` (with tests in https://github.com/klothoplatform/klotho/pull/487/files#diff-03d25bd6b71fc8fcc391be68236b07406fbb298a04ada456b9f28794e7c50b9aR143)?

pkg/infra/kubernetes/exec_unit.go
jhsinger-klotho: we could either make every node group obey the min disk size or make one obey it and still put a node selector on

pkg/provider/aws/execution_unit.go
jhsinger-klotho: ah good point 👍

jhsinger-klotho (commented):
just questions around how disk size will be handled, but thats it

pkg/provider/aws/resources/eks.go
```diff
@@ -65,43 +110,122 @@ type (
...
+}
+
+func NodeGroupNameFromConfig(cfg config.ExecutionUnit) string {
+	params := cfg.GetExecutionUnitParamsAsKubernetes()
+	return NodeGroupName(cfg.NetworkPlacement, params.InstanceType)
+}
+
+func NodeGroupName(networkPlacement string, instanceType string) string {
+	return nodeGroupSanitizer.Apply(fmt.Sprintf("%s_%s", networkPlacement, instanceType))
 }
```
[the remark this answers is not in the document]
gordon-klotho: Does it need a unique name or is it scoped to the cluster? It seemed to me that they're specific to the cluster (ie, `"nodegroupArn": "arn:aws:eks:us-west-2:012345678910:nodegroup/my-cluster/my-nodegroup/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",` from https://docs.aws.amazon.com/eks/latest/APIReference/API_CreateNodegroup.html)

pkg/provider/aws/resources/eks.go
gordon-klotho: That's in parity with the logic before:
https://github.com/klothoplatform/klotho/blob/main/pkg/infra/pulumi_aws/iac/eks.ts#L479-L502

I can add a comment though for the future.

pkg/provider/aws/resources/eks_test.go
gordon-klotho: Unit `b` (the next line from this one) has the disk size of 50 to cover this

jhsinger-klotho (approved):

jhsinger-klotho (approved):

PR #488, opened 2023-04-10
expose for eks

### LINKED ISSUE #421: Expose with EKS
This is dependent on both Phase 1 and 2 completion

the definition of done:

jhsinger-klotho:
• Does it relate to or fix any issue? closes #421 

adds load balancer, target group, and listener

adds the ability to get the URI for nlbs, to use for api integrations with expose

```
const awsLoadBalancerMpMain = new aws.lb.LoadBalancer(`mp-main`, {
                    internal: `internal` == "internal",
                    loadBalancerType: `network`,
                    subnets: [awsVpcSubnetMpPrivate2,awsVpcSubnetMpPrivate1].map(subnet => subnet.id),
                    tags: {},
                    securityGroups: [awsSecurityGroupMp].map(sg => sg.id)
                });

const awsApiIntegrationMpAppWriteFilePathPOST = new aws.apigateway.Integration(`mp-app-write-file/-path-/-POST`, {
        restApi: awsRestApiMpApp.id,
        resourceId: awsApiResourceMpAppWriteFilePath.id,
        httpMethod: awsApiMethodMpAppWriteFilePathPOST.httpMethod,
        integrationHttpMethod: awsApiMethodMpAppWriteFilePathPOST.httpMethod,
        type: `VPC_LINK`,
        
        
        connectionId: awsVpcLinkAwsLoadBalancerMpMain.id,
        
        uri: pulumi.interpolate`http://${awsLoadBalancerMpMain.dnsName}/write-file/{path}`,
        requestParameters: {},
    });

const awsTargetGroupMpMain = new aws.lb.TargetGroup(`mp-main`, {
        port: 3000,
        protocol: `TCP`,
        targetType: `ip`,
        vpcId: awsVpcMp.id,
        tags: {},
    });

const awsLoadBalancerListenerMpMainMain = new aws.lb.Listener(`mp-main-main`, {
        loadBalancerArn: awsLoadBalancerMpMain.arn,
        defaultActions: [{targetGroupArn: awsTargetGroupMpMain.arn,
type: `forward`,
}],
        port: 80,
        protocol: `TCP`,
    });
```

### Standard checks


pkg/provider/aws/execution_unit_test.go
```diff
@@ -35,6 +35,11 @@ func Test_GenerateExecUnitResources(t *testing.T) {
 				Type:         string(kubernetes.ImageTransformation),
 				Key:          "image",
 			},
+			{
+				ExecUnitName: unit.ID,
+				Type:         string(kubernetes.TargetGroupTransformation),
+				Key:          "tgb",
```
yuval-klotho, line 41: tgb? (what's the "b"?)

yuval-klotho (approved):

pkg/provider/aws/resources/lb.go
```diff
@@ -0,0 +1,135 @@
...
+)
+
+type (
+	LoadBalancer struct {
+		Name                   string
+		ConstructsRef          []core.AnnotationKey
+		IpAddressType          string
+		LoadBalancerAttributes map[string]string
+		Scheme                 string
+		SecurityGroups         []*SecurityGroup
```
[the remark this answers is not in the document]
jhsinger-klotho: well if the security group gets modified we just want to make sure that its in sync everywhere was my thought. Also all of the IRs imlpement the core.Resource interface as pointers so its more natural it feels like. not sure if it would have issues not using pointers though

pkg/provider/aws/execution_unit_test.go
jhsinger-klotho: target group binding is the term for the actual manifest, but it doesnt matter what the key value is so i just put tgb

PR #490, opened 2023-04-11
cleanup: refactor in AssignFileToUnits

yuval-klotho:
I need to add another file to this switch, and it's getting a bit unruly with all the duplication. This pulls the shared logic (which is pretty much all of the `case` bodies) out.

### Standard checks

- **Unit tests**: passing (and they do fail if, for example, I comment out `*handle = ast`)
- **Docs**: n/a

### DIFF
--- a/pkg/infra/kubernetes/helm_chart.go
+++ b/pkg/infra/kubernetes/helm_chart.go
@@ -94,13 +94,12 @@ func (t *HelmChart) AssignFilesToUnits() error {
 			// only one exec unit or (b) there are multiple units, but this one's name matches the k8s object name.
 			// It returns whether that condition matched.
 			setAst := func(k8sObject metav1.Object, handle **core.SourceFile) bool {
-				needsMetadataName := len(t.ExecutionUnits) > 1
-				shouldSet := (!needsMetadataName) || (k8sObject.GetName() == unit.Name)
-				if shouldSet {
+				if len(t.ExecutionUnits) <= 1 || (k8sObject.GetName() == unit.Name) {
 					log.Debugf("Found unit, %s's, pod manifest in file, %s", unit.Name, f.Path())
 					*handle = ast
+					return true
 				}
-				return shouldSet
+				return false
 			}
 
 			obj, err := readFile(ast)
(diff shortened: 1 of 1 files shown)


pkg/infra/kubernetes/helm_chart.go
```diff
@@ -98,61 +111,21 @@ func (t *HelmChart) AssignFilesToUnits() error {
...
-					log.Debug("Found unit, %s's, pod manifest in file, %s", unit.Name, f.Path())
-					if unit.Deployment != nil {
-						return fmt.Errorf("can not support multiple pod specifications for unit %s", unit.Name)
-					}
-					unit.Pod = ast
+				if setAst(o, &unit.Pod) && unit.Deployment != nil {
+					// Don't set this pod if there's already a spec for deployment. That means there's both a deployment
+					// manifest and a pod manifest for the same exec unit, which is a confusing scenario (since a
+					// deployment itself contains a pod spec). For now, we just disallow it.
+					return fmt.Errorf("can not support multiple pod specifications for unit %s", unit.Name)
```
gordon-klotho, line 117: This slightly changes the behaviour - before only one of `.Pod` and `.Deployment` would be set; however, when this error occurs, case they would now both be set. Might not be a big issue, but just pointing that out.

gordon-klotho (approved):

pkg/infra/kubernetes/helm_chart.go
yuval-klotho: Ah yes, I did realize that and forgot to note it in the PR — thanks for calling it out. I had assumed it's not a problem, since we're just returning an `error` anyway; but @jhsinger-klotho, can you confirm that it's okay?

It also is a bit odd that an error would still result in changed state. It's not completely unprecedented (I'd expect the same from a stream read, for example), but certainly nonstandard for an object that's completely in-memory with no external resources.

One alternative (in fact, the original version of this refactor) was to have the helper not actually do the setting, but just return a `bool`. That just makes it slightly more boilerplatey to use, so I opted against it.

I could also just have this branch manually set `unit.Pod = nil`.

yuval-klotho: Discussed it offline, we're all good.

PR #491, opened 2023-04-12
rest of the k8s configs

### LINKED ISSUE #419: Kubernetes plugin work

yuval-klotho:
The rest of the k8s configs, including Horizontal Pod Autoscaling (HPA).

If the user supplies an HPA manifest attached to this unit (per the usual rules), we use that. Otherwise, we'll create one only if there's (a) a deployment, and (b) the user has specified HPA configs in their klotho.yaml.

The intent is that this resolves #419, so if you don't think it does, please let me know. :)

> **Warning**
> I'm using the `autoscaling/v2beta2` autoscaler, because it looks like it's the latest? I don't know how much the k8s world like cutting-edge. I think it should be pretty easy to go to the standard `v2` (not beta), if we prefer — it should basically just be some yaml changes, mostly in tests.

### Standard checks

- **Unit tests**: I think I added the appropriate unit tests, but please do check (if you can do it without your eyes glazing over...mine would 😩 )
- **Docs**: not added; please lmk if there's anywhere you'd like comments

### DIFF
--- a/pkg/infra/kubernetes/exec_unit.go
+++ b/pkg/infra/kubernetes/exec_unit.go
@@ -186,8 +186,7 @@ func (unit *HelmExecUnit) transformHorizontalPodAutoscaler(cfg config.ExecutionU
 	}
 	hpa, ok := obj.(*autoscaling.HorizontalPodAutoscaler)
 	if !ok {
-		err = fmt.Errorf("expected file %s to contain HorizontalPodAutoscaler Kind", unit.HorizontalPodAutoscaler.Path())
-		return nil, nil
+		return nil, fmt.Errorf("expected file %s to contain HorizontalPodAutoscaler Kind", unit.HorizontalPodAutoscaler.Path())
 	}
 	k8Cfg := cfg.GetExecutionUnitParamsAsKubernetes()
 	hpaCfg := k8Cfg.HorizontalPodAutoScalingConfig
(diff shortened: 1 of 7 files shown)


pkg/infra/kubernetes/exec_unit.go
```diff
@@ -167,7 +177,91 @@ func (unit *HelmExecUnit) transformDeployment(cfg config.ExecutionUnit) ([]HelmC
...
+	hpa, ok := obj.(*autoscaling.HorizontalPodAutoscaler)
+	if !ok {
+		err = fmt.Errorf("expected file %s to contain HorizontalPodAutoscaler Kind", unit.HorizontalPodAutoscaler.Path())
+		return nil, nil
+	}
+	k8Cfg := cfg.GetExecutionUnitParamsAsKubernetes()
+	hpaCfg := k8Cfg.HorizontalPodAutoScalingConfig
+
+	if k8Cfg.Replicas != 0 {
+		minReplicas := int32(k8Cfg.Replicas)
```
jhsinger-klotho, line 196: im not sure we want to set this to what they specify the replicas add because i could see scenarios in which they would want to scale down. Not blocking though since its our current UX

jhsinger-klotho (approved):
Just a couple of lint errors, but the rest looks good to me

PR #492, opened 2023-04-12
recurse in dependencies reflect and vpc config for lambda

jhsinger-klotho:

- puts lambdas in vpcs if needed
- makes DependenciesReflect recursive

### Standard checks

### DIFF
--- a/pkg/core/resource_graph.go
+++ b/pkg/core/resource_graph.go
@@ -111,22 +111,6 @@ func (rg *ResourceGraph) AddDependenciesReflect(source Resource) {
 		sourceValue = sourceValue.Elem()
 		sourceType = sourceType.Elem()
 	}
-	add := func(targetValue reflect.Value) {
-		if targetValue.Kind() == reflect.Pointer && targetValue.IsNil() {
-			return
-		}
-		if !targetValue.CanInterface() {
-			return
-		}
-		switch value := targetValue.Interface().(type) {
-		case Resource:
-			rg.AddDependency(source, value)
-		case *IaCValue:
-			rg.AddDependency(source, value.Resource)
-		case IaCValue:
-			rg.AddDependency(source, value.Resource)
-		}
-	}
 	for i := 0; i < sourceType.NumField(); i++ {
 		// TODO maybe add a tag for options for things like ignoring fields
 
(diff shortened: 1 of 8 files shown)


pkg/core/resource_graph.go
```diff
@@ -135,17 +119,48 @@ func (rg *ResourceGraph) AddDependenciesReflect(source Resource) {
 		case reflect.Slice, reflect.Array:
 			for elemIdx := 0; elemIdx < fieldValue.Len(); elemIdx++ {
 				elemValue := fieldValue.Index(elemIdx)
-				add(elemValue)
+				rg.addDependenciesReflect(source, elemValue)
```
gordon-klotho, line 122: The godoc for this function needs to be updated to reflect (heh) that it is recursive and not just first-level anymore.

pkg/core/resource_graph_test.go
```diff
@@ -96,8 +104,9 @@ func TestResourceGraph_AddDependenciesReflect(t *testing.T) {
 		"value_arr1", "value_arr2",
 		"value_ptr_arr1", "value_ptr_arr2",
 		"value_map1", "value_map2",
-		"value_ptr_map1", "value_ptr_map2",
+		"value_ptr_map1", "value_ptr_map2", "nested",
```
gordon-klotho, line 107: nit: these were generally only on the same line if they were related - since this one isn't related to anything it should get its own line

pkg/provider/aws/execution_unit.go
```diff
@@ -49,6 +49,12 @@ func (a *AWS) GenerateExecUnitResources(unit *core.ExecutionUnit, result *core.C
 	switch execUnitCfg.Type {
 	case Lambda:
 		lambdaFunction := resources.NewLambdaFunction(unit, a.Config.AppName, role, image)
+		if resources.VpcExists(dag) {
```
gordon-klotho, line 52: Would this have a race condition with whether the VPC was created first or not? Or is there something that always sets the VPC up first before iterating through the resources.

pkg/provider/aws/resources/vpc.go
```diff
@@ -139,6 +148,17 @@ func (vpc *Vpc) GetPrivateSubnets(dag *core.ResourceGraph) []*Subnet {
 	return subnets
 }
 
+func (vpc *Vpc) GetSecurityGroups(dag *core.ResourceGraph) []*SecurityGroup {
```
gordon-klotho, line 151: At this point, why not have the SecurityGroups as a field on VPC?

gordon-klotho (commented):
Only "main" blocking question is the iteration order one. The rest are non-blocking and generally looks good.

pkg/provider/aws/resources/vpc.go
jhsinger-klotho: because then dependencies reflect would go the incorrect way in the dependencies it initializes

pkg/provider/aws/execution_unit.go
jhsinger-klotho: yeah this would if it required to proxy to an eks exec unit, but not for persist resources. I think we may need to determine if we are creating a vpc first thing

pkg/provider/aws/resources/vpc.go
gordon-klotho: Ah, gotcha - the SGs are created in the VPC

pkg/provider/aws/execution_unit.go
gordon-klotho: That's fine then. Can you add a comment explaining why this is safe?

jhsinger-klotho: i added the method to just check if it needs to be created and do that first instead just because its clearer and it may end up catching some other cases we dont think about

pkg/provider/aws/plugin.go
```diff
@@ -63,6 +71,26 @@ func (a *AWS) Translate(result *core.ConstructGraph, dag *core.ResourceGraph) (l
 	return
 }
 
+// shouldCreateNetwork determines whether any of our aws resources will need to be within a VPC
+func (a *AWS) shouldCreateNetwork(result *core.ConstructGraph) (bool, error) {
+	constructIds, err := result.TopologicalSort()
+	if err != nil {
+		return false, err
+	}
+	for _, id := range constructIds {
```
gordon-klotho, line 80: Does this need to be in topological order or could you just use ListConstructs (or equiv.)? If not, that'd get rid of needing `error` too.

jhsinger-klotho: youre right, can just list constructs here, let me change that

DavidSeptimus-Klotho (approved):

PR #493, opened 2023-04-12
make security group rules dynamic

### LINKED ISSUE #465: Dynamic ingress rules for SecurityGroup
Currently, the ingress rules are created statically within the factory - these should come from args so specific rules can be added by the provider.

jhsinger-klotho:
• Does it relate to or fix any issue? closes #465 

Makes our rules dynamic so we can manipulate them in the provider. This makes it so we only include ones that are necessary, where as previously we would always input the eks rule

```
const awsSecurityGroupDemoKube = new aws.ec2.SecurityGroup(`demo-kube`, {
        name: `demo-kube`,
        vpcId: awsVpcDemoKube.id,
        egress: [{description: `Allows all outbound IPv4 traffic.`,
cidrBlocks: [`0.0.0.0/0`],
protocol: `-1`,
}],
        ingress: [{description: `Allow ingress traffic from ip addresses within the vpc`,
cidrBlocks: [awsVpcDemoKube.cidrBlock],
protocol: `-1`,
},{description: `Allow ingress traffic from within the same security group`,
protocol: `-1`,
self: true,
},{description: `Allows ingress traffic from the EKS control plane`,
cidrBlocks: [`0.0.0.0/0`],
fromPort: 9443,
protocol: `TCP`,
toPort: 9443,
}],
    });
```

### Standard checks

- **Unit tests**: Any special considerations? added new and missing unit tests.

### DIFF
--- a/pkg/provider/aws/plugin.go
+++ b/pkg/provider/aws/plugin.go
@@ -109,6 +109,16 @@ func (a *AWS) createEksClusters(result *core.ConstructGraph, dag *core.ResourceG
 	}
 
 	vpc := resources.CreateNetwork(a.Config, dag)
+	sg := resources.GetSecurityGroup(a.Config, dag)
+	sg.IngressRules = append(sg.IngressRules, resources.SecurityGroupRule{
+		Description: "Allows ingress traffic from the EKS control plane",
+		FromPort:    9443,
+		Protocol:    "TCP",
+		ToPort:      9443,
+		CidrBlocks: []core.IaCValue{
+			{Property: "0.0.0.0/0"},
+		},
+	})
 	subnets := vpc.GetVpcSubnets(dag)
 	for clusterId, units := range clusterIdToUnit {
 		resources.CreateEksCluster(a.Config.AppName, clusterId, subnets, nil, units, dag)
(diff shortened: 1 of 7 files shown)


pkg/provider/aws/plugin.go
```diff
@@ -109,6 +109,16 @@ func (a *AWS) createEksClusters(result *core.ConstructGraph, dag *core.ResourceG
...
 
 	vpc := resources.CreateNetwork(a.Config, dag)
+	sg := resources.GetSecurityGroup(a.Config, dag)
+	sg.IngressRules = append(sg.IngressRules, resources.SecurityGroupRule{
+		Description: "Allows ingress traffic from the EKS control plane",
+		FromPort:    9443,
+		Protocol:    "TCP",
+		ToPort:      9443,
+		CidrBlocks: []core.IaCValue{
+			{Property: "0.0.0.0/0"},
```
yuval-klotho, line 119: Do we want this to be configurable? or are we okay with...

a **hard**<sup>coded</sup> **cidr**?

yuval-klotho (approved):

pkg/provider/aws/plugin.go
jhsinger-klotho: this is a specific rule for the eks control plane and im not sure if it can be changed ever for the control plane to work so hardcoded should be fine. We maybe can look into if theres a narrower cidr but i didnt think they provided one

yuval-klotho: I was just making a pun

jhsinger-klotho: oh i see it now lol

PR #494, opened 2023-04-12
add validation back and make provider validation specific to the prov…

### LINKED ISSUE #470: Rework Validation back in
We currently removed validation while we work on iac2.

We can now split our validation into 2 phases:
 - Compiler Validation: This performs validation checks on the constructs itself and will always be run
 - Provider Validation: These are the validation functions that take in provider and are provider specific

Definition of done:
- Have the validation being run in 2 phases
- Have compiler validation always being run

jhsinger-klotho:
• Does it relate to or fix any issue? closes #470 

Split up validation into 2 parts
- Core validation - This is validation around the annotations/ids/config overrides themselves and has nothing to do with provider specific knowledge
- Provider validtion - This is very much provider specific and i added it as a part of the plugin interface so we can run it for every provider

The provider validation step has a common utility for handling provider validation so that as long as we map types to our constructs, we can have the same reusable logic

The kubernetes plugin really shouldnt be a part of the provider plugin interface as it stands today and is really a analysis and transform. Weve had this discussion before, but dont have the time to do a complete refactor there so for now we just return nil in validation since there is nothing to validate.

### Standard checks

- **Unit tests**: Any special considerations? made sure our unit tests still covered what we are doing

### DIFF
--- a/pkg/compiler/compiler.go
+++ b/pkg/compiler/compiler.go
@@ -6,7 +6,9 @@ import (
 	"fmt"
 
 	"github.com/klothoplatform/klotho/pkg/annotation"
+	"github.com/klothoplatform/klotho/pkg/config"
 	"github.com/klothoplatform/klotho/pkg/core"
+	"github.com/klothoplatform/klotho/pkg/validation"
 	"github.com/pelletier/go-toml/v2"
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
@@ -27,6 +29,7 @@ type (
 	ProviderPlugin interface {
 		Plugin
 		Translate(result *core.ConstructGraph, dag *core.ResourceGraph) ([]core.CloudResourceLink, error)
+		Validate(config *config.Application, constructGraph *core.ConstructGraph) error
 	}
 
 	IaCPlugin interface {
@@ -49,6 +52,9 @@ type (
 )
 
 func (c *Compiler) Compile() error {
+
+	userOverridesConfiguration := *c.Document.Configuration
+
 	// Add our internal resource to be used for provider specific implementations. ex) aws dispatcher requires the payloads bucket and so does proxy
 	// TODO: We could likely move this into runtime, but until we refactor that to be common we can keep this here so it lives in one place.
 	// We previously always created the payloads bucket so the behavior is no different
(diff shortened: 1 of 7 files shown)


pkg/provider/provider.go
```diff
@@ -38,3 +44,36 @@ type (
...
+		mapping := p.GetKindTypeMappings(resource)
+		if len(mapping) == 0 {
+			errs.Append(errors.Errorf("Provider, %s, Does not support %s ", p.Name(), reflect.ValueOf(resource).Kind()))
+			continue
+		}
+		resourceType := config.GetResourceType(resource)
+		log.Debugf("Checking if provider, %s, supports %s and type, %s, pair.", p.Name(), resource.Provenance().Capability, resourceType)
+		for _, validType := range mapping {
+			if validType == resourceType {
+				resourceValid = true
```
yuval-klotho, line 70: nit: you can just `return nil` at this point, and then you don't need `resourceValid` at all.

pkg/provider/provider_test.go
```diff
@@ -0,0 +1,92 @@
...
+			}
+
+			err := HandleProviderValidation(provider, &tt.cfg, result)
+			if tt.wantErr {
+				assert.Error(err)
+				return
+			} else {
+				assert.NoError(err)
+				return
+			}
```
yuval-klotho, line 67: minor, only if you want it:

```suggestion
			assert.Equal(tt.wantErr, err != nil)
```

pkg/compiler/compiler.go
```diff
@@ -27,6 +29,7 @@ type (
 	ProviderPlugin interface {
 		Plugin
 		Translate(result *core.ConstructGraph, dag *core.ResourceGraph) ([]core.CloudResourceLink, error)
+		Validate(config *config.Application, constructGraph *core.ConstructGraph) error
```
yuval-klotho, line 32: Rather than all plugins having `Validate`, what about creating a new interface for just this method, and then checking in `compiler.go` whether it's of that type? That way, the default "no validation needed" (like in the k8s plugin" becomes just "don't write the method at all".

yuval-klotho (commented):
I mostly gave this a cursory look, but it lgtm. Had a couple minor comments, nothing blocking.

yuval-klotho (approved):

pkg/compiler/compiler.go
jhsinger-klotho: i think that would cause the import cycle which i was fighting with was the issue otherwise i fully agree. I was hoping to put that method on the provider interface itself, but then we cant type check it in the compiler. Thoughts?

yuval-klotho: It shouldn't cause any import cycle, because the structs don't need to (and in fact can't) import this interface to implement it. I think all you'd need to do is, here do:

```go
ValidatingPlugin interface
	Validate(config *config.Application, constructGraph *core.ConstructGraph) error
}
```

and then below, where you have:

```
err := p.Validate(c.Document.Configuration, c.Document.Constructs)
if err != nil { ...
```

... instead do:

```go
if validator, ok := p.(ValidatingPlugin); ok {
	err := validator.Validate(c.Document.Configuration, c.Document.Constructs)
	if err != nil { ...
}
```

and then just rm the `Kubernetes.Validate` method.

Just to be clear, this isn't a blocker.

jhsinger-klotho: right just seeing if theres a more optimal way. i think i understand what you are mentioning and may like that better so could give it a try

yuval-klotho (approved):

PR #497, opened 2023-04-12
remove all render tags and add parsVal function back for security gro…

### LINKED ISSUE #474: Remove render tag from all core.Resource objects
Because this logic is really iac specific it should just be handled in the iac plugin.

We have the logic to render sub resources as a document and keep PolicyDoc upper case so this should be our default logic if the resource is not of type core.Resource or core.IaCValue

jhsinger-klotho:
• Does it relate to or fix any issue? closes #474 

We remove all tags from fields within Provider IR resources

our logic is now
- If core.Resource then handle
- if core.IaCValue then handle
- If there is a nested tempalte for the field, then render the nested tempalte
- If not, render as a document

A couple notes:
- When we render as a document, any zeroValue is not rendered, so if '' or 0 or nil is valid, you would need to create a nested tempalte
- added back the template function of parseVal so that we can render resources and iacValues within nested templates.

David pointed out the document version broke the security group, so added a nested template for that and heres the end result

```
const awsSecurityGroupMp = new aws.ec2.SecurityGroup(`mp`, {
        name: `mp`,
        vpcId: awsVpcMp.id,
        egress: [{
    cidrBlocks: [`0.0.0.0/0`],
    description: "Allows all outbound IPv4 traffic.",
    fromPort: 0,
    protocol: "-1",
    toPort: 0,
}
],
        ingress: [{
    cidrBlocks: [awsVpcMp.cidrBlock],
    description: "Allow ingress traffic from ip addresses within the vpc",
    fromPort: 0,
    protocol: "-1",
    toPort: 0,
}
,{
    description: "Allow ingress traffic from within the same security group",
    fromPort: 0,
    protocol: "-1",
    self: true,
    toPort: 0,
}
,{
    cidrBlocks: [`0.0.0.0/0`],
    description: "Allows ingress traffic from the EKS control plane",
    fromPort: 9443,
    protocol: "TCP",
    toPort: 9443,
}
],
    });
```
### Standard checks

### DIFF
--- a/pkg/infra/iac2/templates_compiler.go
+++ b/pkg/infra/iac2/templates_compiler.go
@@ -25,7 +25,6 @@ type (
 	nestedTemplateValue struct {
 		parentValue            reflect.Value
 		childValue             reflect.Value
-		iacTag                 string
 		tc                     *TemplatesCompiler
 		useDoubleQuotedStrings bool
 	}
@@ -58,6 +57,13 @@ type (
 		resourceVarNames map[string]struct{}
 		// resourceVarNamesById is a map from resource id to the variable name for that resource
 		resourceVarNamesById map[string]string
+		// ctx is a pointer to the current context being used within the templates compiler. This context is used when parsing values within nested templates.
+		ctx *NestedCtx
+	}
+	NestedCtx struct {
+		useDoubleQuotes bool
+		appliedOutputs  *[]AppliedOutput
+		rootVal         *reflect.Value
 	}
 )
 
@@ -76,7 +82,7 @@ func (s stringTemplateValue) Raw() interface{} {
 
 func (v nestedTemplateValue) Parse() (string, error) {
 	childVal := v.childValue
-	return v.tc.resolveStructInput(&v.parentValue, childVal, v.useDoubleQuotedStrings, v.iacTag, nil)
+	return v.tc.resolveStructInput(&v.parentValue, childVal, v.useDoubleQuotedStrings, nil)
 }
 
 func (v nestedTemplateValue) Raw() interface{} {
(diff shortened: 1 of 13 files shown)


pkg/infra/iac2/templates_compiler.go
```diff
@@ -565,7 +572,7 @@ func (tp templatesProvider) getTemplateForType(typeName string) (ResourceCreatio
 	return template, nil
 }
 
-func (tp templatesProvider) getNestedTemplate(templatePath string) (*template.Template, error) {
+func (tp templatesProvider) getNestedTemplate(templatePath string, funcs template.FuncMap) (*template.Template, error) {
```
DavidSeptimus-Klotho, line 575: I don't think this method of supplying functions is going to work with caching of parsed templates since the function and its context are ignored after the first time we get the template. We probably need a single up-front `parseVal` function declaration that we supply with a reference to some context holder that we update prior to each attempt to parse a nested template value.

DavidSeptimus-Klotho (changes requested):

pkg/infra/iac2/templates_compiler.go
jhsinger-klotho: alright, i had just copied what you had done previously, but i can rework it. Youre thinking of something like 

```
func(val reflect.Value) (string, error) {
					return tc.resolveStructInput(nil, val, useDoubleQuotedStrings, "", appliedOutputs)
				},
			}
```

needs to be defined staticly somewhere and just referenced in the `	tmpl, err := template.New(templatePath).Funcs(funcs).Parse(string(contents))
`
line?

DavidSeptimus-Klotho: partially, but you'd also need to do something like this:

```go
type (
  TemplatesCompiler struct {
    ...
    Ctx *NestedCtx
  }

  NestedCtx struct {
    useDoubleQuotes bool
    appliedOutputs []AppliedOutput
    rootVal reflect.Value
  }
)

...

// parseVal
func(val reflect.Value) (string, error) {
  return tc.resolveStructInput(tc.Ctx.rootVal, val, tc.Ctx.useDoubleQuotedStrings, "", tc.Ctx.appliedOutputs)
}
```

Then make sure to update `tc.Ctx` before each attempt to parse a value in a nested template.

PR #503, opened 2023-04-13
Don't use the greedy marker for intergration URIs

gordon-klotho:
Having `{rest+}` on the endpoint url causes "illegal character". 

### Standard checks


yuval-klotho:
Are there any tests we can add to this?

Issue #420, opened 2023-03-28
EKS prereq chart setup


jhsinger-klotho:
Dependent on the task for helm charts being deployable

Setup all prerequisite charts if required:

- [ ] VPC CNI Plugin -> Always
  - [ ] in progress
  - [ ] done
- [ ] ALB chart -> if exposed exec unit
  - [x] in progress
  - [ ] done
- [x] Metrics server -> Always
  - [x] in progress
  - [x] done
- [x] Cert Manager -> always 
  - [x] in progress
  - [x] done
- [ ] NVidia driver -> If instance type is GPU (Configurable if we arent owning the cluster)
  - [ ] in progress
  - [ ] done
- [ ] Cloud map chart -> If proxy (lower-pri; only needed for multi exec. possibly defer to a separate task for multi-exec. this also uses kustomize instead of helm, so it may be worth spinning off to its own)
  - [ ] in progress
  - [ ] done

Stretch Goals **[deferred]**:
- [ ] Include chart dependencies within our klotho generated Chart.yaml (as opposed to pulumi)
    - not needed for now; requires some work on our `dependsOn` handling to call `ready()` as appropriate
- [ ] Include a high level cluster config that determines if these can be installed (This will be important when imported clusters are being used because we could negatively effect their clusters by installing these no matter what).
    - not needed for now; will be needed for imports

Definition Of Done:
- Determinations are made on which charts need to be installed
- Proper IaC Is rendered for these charts

yuval-klotho:
Looking at the legacy `add_on` dirs: do we need to handle fluentbit as part of this work?

jhsinger-klotho:
theres a separate task for fluentbit, which im handling at the moment

PR #507, opened 2023-04-13
Klotho 420 prereqs

yuval-klotho:
As part of this, i'm also changing the helmchart.values map to have values of type `any`, instead of `iacvalue`. this because the cert-manager chart wants values that include a boolean (for `installcrds`) and a nested record/map (for `webhook`). in order to accomplish that, i also changed the helm_chart/factory.ts args to take `pulumi.output<any>` instead of `...<string>`. this is what the pulumi field is actually typed as, so we don't actually lose any checks by doing it this way.

also fix a typo in the factory.ts's handling of args.repo.

also also provider a prettier error in coretesting, because otherwise it's very hard to read the diff when two lists have different elements.

in service of #420

### Standard checks

- **Unit tests**: n/a
- **Docs**: n/a

### DIFF
--- a/pkg/core/coretesting/resources.go
+++ b/pkg/core/coretesting/resources.go
@@ -2,6 +2,7 @@ package coretesting
 
 import (
 	"fmt"
+	"reflect"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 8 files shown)


pkg/core/coretesting/resources.go
```diff
@@ -32,11 +33,40 @@ func (expect ResourcesExpectation) Assert(t *testing.T, dag *core.ResourceGraph)
 		assert.Subset(t, got.Nodes, expect.Nodes)
 		assert.Subset(t, got.Deps, expect.Deps)
 	} else {
-		assert.ElementsMatch(t, expect.Nodes, got.Nodes)
-		assert.ElementsMatch(t, expect.Deps, got.Deps)
+		expect.ElementsMatchPretty(t, expect.Nodes, got.Nodes)
```
jhsinger-klotho, line 36: does this do a deep equals?

jhsinger-klotho: oh i see your comment below, so that seems like what it is

pkg/provider/aws/resources/eks.go
```diff
@@ -102,6 +102,36 @@ func CreateEksCluster(appName string, clusterName string, subnets []*Subnet, sec
 		dag.AddDependency(nodeGroup, s)
 		dag.AddDependency(profile, s)
 	}
+
+	for _, addOn := range createAddOns(clusterName, references) {
+		dag.AddResource(addOn)
+		dag.AddDependency(addOn, nodeGroup)
```
jhsinger-klotho, line 108: we have the dependsOn in the helm chart factory right? Pretty sure we do, but if not will need to add it

jhsinger-klotho (approved):
LGTM!

pkg/provider/aws/resources/eks.go
yuval-klotho: yep, we do

PR #508, opened 2023-04-13
setting up eks logging for fargate and fluent bit

### LINKED ISSUE #444: EKs Set up fluentBit and fargateLogging
To be able to setup proper logging for eks, we currently have 2 functions in deploylibs `k8s/add_ons` directory
- enableFargateLogging
- installFluentBitForCW
We need these to be replicated to the new format in the kubernetes or aws plugin.

Definition of done.
- The cluster has both of these pieces of functionality working.
- Manifests are installed and output to the compiled dir

jhsinger-klotho:
• Does it relate to or fix any issue? closes #444 

sets up fargate logging and fluent bit for container insights.

We now can install manifests via ConfigFile in pulumi.

We also output our manifests to the `compiled/manifests` directory

### Standard checks

- **Unit tests**: Any special considerations? added test case

### DIFF
--- a/pkg/cli/klothomain.go
+++ b/pkg/cli/klothomain.go
@@ -254,7 +254,7 @@ func (km KlothoMain) run(cmd *cobra.Command, args []string) (err error) {
 			cmd.SilenceUsage = true
 		},
 	}
-	// defer analyticsClient.PanicHandler(&err, errHandler)
+	defer analyticsClient.PanicHandler(&err, errHandler)
 
 	updateStream := options.Update.Stream.OrDefault(km.DefaultUpdateStream)
 	analyticsClient.AppendProperty("updateStream", updateStream)
--- a/pkg/provider/aws/resources/eks.go
+++ b/pkg/provider/aws/resources/eks.go
@@ -1,10 +1,15 @@
 package resources
 
 import (
+	"embed"
 	"fmt"
+	"io/fs"
+	"path"
+
 	"github.com/klothoplatform/klotho/pkg/core"
 	"github.com/klothoplatform/klotho/pkg/infra/kubernetes"
 	"github.com/klothoplatform/klotho/pkg/sanitization/aws"
+	"go.uber.org/zap"
 )
 
 const (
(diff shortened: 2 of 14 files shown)


pkg/cli/klothomain.go
```diff
@@ -254,7 +254,7 @@ func (km KlothoMain) run(cmd *cobra.Command, args []string) (err error) {
 			cmd.SilenceUsage = true
 		},
 	}
-	// defer analyticsClient.PanicHandler(&err, errHandler)
+	defer analyticsClient.PanicHandler(&err, errHandler)
```
yuval-klotho, line 257: 😬 lol

pkg/provider/aws/resources/eks.go
```diff
@@ -79,6 +95,15 @@ func CreateEksCluster(appName string, clusterName string, subnets []*Subnet, sec
 	dag.AddResource(cluster)
 	dag.AddDependency(cluster, clusterRole)
 
+	err := cluster.createFargateLogging(references, dag)
+	if err != nil {
+		zap.S().Warnf("Unable to set up Fargate logging manifests for cluster %s: %s", clusterName, err.Error())
```
yuval-klotho, line 100: instead of (or in addition to, your pick) just printing the err, also add it as a zap field (`zap.Error(err)` or something like that, I forget exactly). That will let us get the full stack trace on telemetry.

ditto below

yuval-klotho (commented):
lgtm after the logging thing. no need for second review/approval from me after that, though.

pkg/cli/klothomain.go
jhsinger-klotho: yeah i saw that as i was debugging... not sure when that got commented out

pkg/provider/aws/resources/eks.go
jhsinger-klotho: well wouldnt we actually fail compilation with zap.Error? The reason i opted for warn is because these charts just affect streaming logs to CW and wouldnt affect any bit of the functionality

yuval-klotho (approved):
hitting the approve button so that GH won't yell at you when you try to merge, since my last review was really a soft-approve anyway

pkg/provider/aws/resources/eks.go
yuval-klotho: I'm not saying log at error level, I'm just saying attach the field. When we send the telemetry, we look at the fields, and any that are of type `error` we extract the full stack trace from and send that.

Issue #505, opened 2023-04-13
add permissions for exec unit proxying to lambda


jhsinger-klotho:
we need proper permissions to call lambdas during proxy.

Issue #495, opened 2023-04-12
ServiceDiscovery for Proxy


jhsinger-klotho:
We need to create the IR and factories for service discovery so we can proxy from exec unit -> eks exec unit:

Definition of Done:
- Service discovery is properly rendered in index.ts
- We have a proper IR of service discovery

PR #511, opened 2023-04-14
adding proxy code for lambda and kubernetes

jhsinger-klotho:
• Does it relate to or fix any issue? closes both #505 and #495

Adding service discovery namespace (cloudmap) for proxying to eks
Adding policy for invoke:function for lambda proxying

- Adds a method in templates compiler to renderSubResources. This is because in aws-classic you cant specify roles on a policy to attach them (which you can do in every other iac) and have to do an aws iam role policy attachment. The RolePolicyattachment is not an aws resource so it does not belong in the graph.
- This method will be for specific resources if you need to code in some intermediate resource without injecting it into the graph.

- Adding FileDependencies to the compilation document
- Making Analysis and transformation plugins take in file dependencies so we can properly do exec split and proxy stubs

Note: This work cannot be tested fully until we have upgraded klotho pro - It now has been tested since this corresponds with the klotho pro updates.

### Standard checks

- **Unit tests**: Any special considerations? added new ones and modify where necessary


pkg/sanitization/aws/cloud_map.go
```diff
@@ -0,0 +1,17 @@
...
+
+	"github.com/klothoplatform/klotho/pkg/sanitization"
+)
+
+// PrivateDnsNamespaceSanitizer returns a sanitized private dns namespace when applied.
+var PrivateDnsNamespaceSanitizer = sanitization.NewSanitizer(
+	[]sanitization.Rule{
+		// strip any leading non alpha characters
+		{
+			Pattern:     regexp.MustCompile(`^[!-~]+$`),
```
DavidSeptimus-Klotho, line 14: This rule seems off:
1. `// strip any leading non alpha characters` seems like a copy/paste issue -- that's not what this pattern does
2. `[!-~]` should be `[^!-~]` since this rule is supposed to detect invalid characters for replacement.
3. The max length should be 253 per documentation

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-servicediscovery-privatednsnamespace.html

I don't know how much this matters if we're still planning a resource name validation pass on the codebase at the end.

DavidSeptimus-Klotho (approved):

PR #514, opened 2023-04-14
adding ALB controller to Helm

yuval-klotho:
Still very much of a WIP

### Standard checks

### DIFF
--- a/pkg/core/resource_graph.go
+++ b/pkg/core/resource_graph.go
@@ -1,6 +1,7 @@
 package core
 
 import (
+	"github.com/pkg/errors"
 	"reflect"
 
 	"github.com/klothoplatform/klotho/pkg/graph"
@@ -147,5 +148,26 @@ func (rg *ResourceGraph) AddDependenciesReflect(source Resource) {
 		default:
 			add(fieldValue)
 		}
+	}
+}
+
+// FindOnlyResource looks through the graph to find a resource of the specified type. It returns that resource, or an
+// error if it did not find exactly one such resource.
+func FindOnlyResource[R Resource](rg *ResourceGraph) (R, error) {
+	var possibilities []R
+	for _, res := range rg.ListResources() {
+		if res, ok := res.(R); ok {
+			possibilities = append(possibilities, res)
+		}
+	}
+	switch len(possibilities) {
+	case 1:
+		return possibilities[0], nil
+	case 0:
+		var zero R
+		return zero, errors.New(`could not find element`)
+	default:
+		var zero R
+		return zero, errors.Errorf(`expected 1 element, found %d`, len(possibilities))
 	}
 }
--- a/pkg/infra/iac2/templates/helm_chart_alb_controller/factory.ts
+++ b/pkg/infra/iac2/templates/helm_chart_alb_controller/factory.ts
@@ -1,0 +1,61 @@
+// A specialization of the more generic helm_chart/ template, for the ALB controller. This needs to create a service
+// account, which itself needs things like VPC id and a role, so it's easier to just create it as a separate template.
+import * as aws from '@pulumi/aws'
+import * as pulumi from '@pulumi/pulumi'
+import * as pulumi_k8s from '@pulumi/kubernetes'
+
+interface Args {
+    Name: string
+    Namespace: string
+    ClustersProvider: pulumi_k8s.Provider
+    ClusterName: string
+    Region: pulumi.Output<pulumi.UnwrappedObject<aws.GetRegionResult>>
+    Vpc: aws.ec2.Vpc
+    Role: aws.iam.Role
+    dependsOn: pulumi.Input<pulumi.Input<pulumi.Resource>[]> | pulumi.Input<pulumi.Resource>
+}
+
+// noinspection JSUnusedLocalSymbols
+function create(args: Args): pulumi_k8s.helm.v3.Chart {
+    return new pulumi_k8s.helm.v3.Chart(
+        args.Name,
+        {
+            chart: 'https://aws.github.io/eks-charts',
+            fetchOpts: {
+                repo: 'https://aws.github.io/eks-charts',
+            },
+            version: '1.4.7',
+            namespace: args.Namespace,
+            values: {
+                clusterName: args.ClusterName,
+                serviceAccount: {
+                    create: false,
+                    name: new pulumi_k8s.core.v1.ServiceAccount(
+                        `${args.ClusterName}-alb-controller`,
+                        {
+                            automountServiceAccountToken: true,
+                            metadata: {
+                                name: `${args.ClusterName}-alb-controller`,
+                                namespace: args.Namespace,
+                                annotations: {
+                                    'eks.amazonaws.com/role-arn': pulumi.interpolate`${args.Role.arn}`,
+                                },
+                            }
+                        }
+                    ).metadata.name,
+                },
+                region: pulumi.interpolate`${args.Region.name}`,
+                vpcId: pulumi.interpolate`${args.Vpc.id}`,
+                podLabels: {
+                    app: 'aws-lb-controller',
+                },
+            },
+        },
+        {
+            provider: args.ClustersProvider,
+            //TMPL {{- if .dependsOn.Raw }}
+            dependsOn: args.dependsOn,
+            //TMPL {{- end }}
+        }
+    )
+}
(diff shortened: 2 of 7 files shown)


pkg/core/resource_graph.go
```diff
@@ -149,3 +150,24 @@ func (rg *ResourceGraph) AddDependenciesReflect(source Resource) {
 		}
 	}
 }
+
+// FindOnlyResource looks through the graph to find a resource of the specified type. It returns that resource, or an
+// error if it did not find exactly one such resource.
+func FindOnlyResource[R Resource](rg *ResourceGraph) (R, error) {
```
yuval-klotho, line 156: oh, I don't think I need this anymore

pkg/infra/iac2/templates/helm_chart_alb_controller/factory.ts
```diff
@@ -0,0 +1,61 @@
...
+    ClustersProvider: pulumi_k8s.Provider
+    ClusterName: string
+    Region: pulumi.Output<pulumi.UnwrappedObject<aws.GetRegionResult>>
+    Vpc: aws.ec2.Vpc
+    Role: aws.iam.Role
+    dependsOn: pulumi.Input<pulumi.Input<pulumi.Resource>[]> | pulumi.Input<pulumi.Resource>
+}
+
+// noinspection JSUnusedLocalSymbols
+function create(args: Args): pulumi_k8s.helm.v3.Chart {
```
jhsinger-klotho, line 19: why a separate factory?

pkg/provider/aws/resources/eks.go
```diff
@@ -283,6 +286,37 @@ func createAddOns(clusterName string, provenance []core.AnnotationKey) []*kubern
...
+					{
+						Effect:    `Allow`,
+						Action:    []string{`sts:AssumeRoleWithWebIdentity`},
+						Resource:  nil,
+						Principal: nil,
+						Condition: &Condition{
+							StringEquals: map[core.IaCValue]string{
+								nil: fmt.Sprintf(`system:serviceaccount:${namespace}:${name}`), // TODO fill out the key, and namenamespace, and name
+							},
+						},
```
jhsinger-klotho, line 315: id maybe make the code which is in execution_unit.go reusable for generating the service account roles because we do that already and its a bit annoying to generate the weird assume role policy

pkg/infra/iac2/templates/helm_chart_alb_controller/factory.ts
yuval-klotho: I thought it makes for a simpler way to get the various nested values into the `values` map, especially with all the `pulumi.interpolate` that needs to happen. It seemed easier than having to create new IacValues, though I could do that if you prefer.

pkg/infra/iac2/templates/helm_chart_alb_controller/factory.ts
```diff
@@ -0,0 +1,61 @@
...
+            values: {
+                clusterName: args.ClusterName,
+                serviceAccount: {
+                    create: false,
+                    name: new pulumi_k8s.core.v1.ServiceAccount(
+                        `${args.ClusterName}-alb-controller`,
+                        {
+                            automountServiceAccountToken: true,
+                            metadata: {
+                                name: args.ClusterName + '-alb-controller',
```
jhsinger-klotho, line 38: nit: i dont think we need to set the metadata name but wont make a difference

pkg/infra/iac2/templates/helm_chart_alb_controller/factory.ts
jhsinger-klotho: well we have the Arn, name and id iac value i think so they should all fit.

We can try to do it after though, would rather get it in so we can start testing, but maybe add a card to fit and finish to clean this up?

pkg/provider/aws/resources/eks.go
```diff
@@ -288,35 +289,75 @@ func NodeGroupName(networkPlacement string, instanceType string) string {
...
+		&HelmChartAlbController{
+			HelmChart: kubernetes.HelmChart{
+
+				Name:             cluster.Name + `-alb-c`,
+				Chart:            `aws-load-balancer-controller`,
+				ConstructRefs:    provenance,
+				Repo:             `https://aws.github.io/eks-charts`,
+				Version:          `1.4.7`,
+				ClustersProvider: clustersProvider,
+				Namespace:        "default", // kube-system if on fargate, default otherwise
```
jhsinger-klotho, line 328: i wouldnt delete these because well clean it up, but they wont do anything since all the values are hardcoded within the factory

pkg/provider/aws/resources/eks.go
```diff
@@ -288,35 +289,75 @@ func NodeGroupName(networkPlacement string, instanceType string) string {
...
+			},
+			ClusterName: cluster.Name,
+			Region:      NewRegion(),
+			Vpc:         vpc,
+			Role:        NewIamRole(cluster.Name, `alb-controller-role`, provenance, ServiceRolePolicyDocument(cluster, `alb-controller`)),
```
jhsinger-klotho, line 333: this role needs a specific policy for it to be able to work. we have all the policy defined in deploylib but its a custom one we have to write

pkg/infra/iac2/templates/helm_chart_alb_controller/factory.ts
yuval-klotho: We had it in the deploylib version, so I kept it. 🤷‍♂️

pkg/infra/iac2/templates/helm_chart_alb_controller/factory.ts
yuval-klotho: Alright, I can convert it to use IaCValue.

I assume I'll need to manually set up the DAG dependencies then, right?

pkg/provider/aws/resources/eks.go
yuval-klotho: Is that policy not what this sets up, via `ServiceRolePolicyDocument`?

The deploylib (`eks.ts`, more precisely) code is:

```typescript
const assumeRolePolicyDoc = pulumi
    .all([this.oidcProvider.arn, this.oidcProvider.url])
    .apply(([oidc_arn, oidc_url]) => {
        return {
            Version: '2012-10-17',
            Statement: [
                {
                    Effect: 'Allow',
                    Principal: {
                        Federated: oidc_arn,
                    },
                    Action: 'sts:AssumeRoleWithWebIdentity',
                    Condition: {
                        StringEquals: {
                            [`${oidc_url}:sub`]: `system:serviceaccount:${namespace}:${name}`,
                        },
                    },
                },
            ],
        }
    })
```

the new `index.ts` we generate, after prettying, is:

```typescript
 pulumi
  .all([
    awsEksClusterTsEksHelmEksCluster.openIdConnectIssuerUrl,
    awsEksClusterTsEksHelmEksCluster.arn,
  ])
  .apply(([cluster_oidc_url, cluster_arn]) => {
    return {
      Version: `2013-10-17`,
      Statement: [
        {
          Effect: `Allow`,
          Principal: {
            Federated: `arn:aws:iam::${
              cluster_arn.split(":")[5]
            }:oidc-provider/${cluster_oidc_url}`,
          },
          Condition: {
            StringEquals: {
              [`${cluster_oidc_url}:sub`]: `system:serviceaccount:default:alb-controller`,
            },
          },
        },
      ],
    };
  })
```

They seem pretty much the same, except for the `oidc_arn`, which is `this.oidcProvider.arn` in the deployib version and `'arn:aws:iam::${cluster_arn.split(":")[5]}:oidc-provider/${cluster_oidc_url}'` in the new. Would that evaluate to something different?

pkg/infra/iac2/templates/helm_chart_alb_controller/factory.ts
```diff
@@ -0,0 +1,61 @@
...
+                                namespace: args.Namespace,
+                                annotations: {
+                                    'eks.amazonaws.com/role-arn': pulumi.interpolate`${args.Role.arn}`,
```
jhsinger-klotho, line 41: we should be creating service accounts via manifests if possible as well so that we have the outputted yaml and it can work directly into the kubernetes plugin

pkg/provider/aws/resources/eks.go
jhsinger-klotho: sorry, referencing https://github.com/klothoplatform/klotho/blob/main/pkg/infra/pulumi_aws/iac/k8s/add_ons/alb_controller/index.ts#L38 as in the permissions it needs to call the ec2 api, etc, to spin up the resources for load balancers and other aws resources.

This policy should be good for assuming the role though

yuval-klotho:
It sounds like my remaining tasks are:

- [ ] resolve merge conflict
- [ ] use IaCValue instead of the custom `helm_chart_alb_controller/factory.ts`
- [ ] create service accounts via manifests
- [ ] create a `ingressController-iam-policy` and attach it to the service account's role

please lmk if I missed anything

PR #517, opened 2023-04-17
fixing bugs

jhsinger-klotho:

fixes circular dependency of making helm chart a direct representative of units. (Caused role -> helm chart and helm chart -> role)

fixes getting env vars into the helm chart

prevents us from entering cyclic edges into a resource graph

### Standard checks


pkg/core/resource_graph.go
```diff
@@ -48,8 +48,12 @@ func (rg *ResourceGraph) AddDependency(deployedSecond Resource, deployedFirst Re
 	for _, res := range []Resource{deployedSecond, deployedFirst} {
 		rg.AddResource(res)
 	}
-	rg.underlying.AddEdge(deployedSecond.Id(), deployedFirst.Id())
-	zap.S().Debugf("adding %s -> %s", deployedSecond.Id(), deployedFirst.Id())
+	if cycle, _ := rg.underlying.CreatesCycle(deployedSecond.Id(), deployedFirst.Id()); cycle {
+		zap.S().Errorf("Not Adding Dependency, Cycle would be created from edge %s -> %s", deployedSecond.Id(), deployedFirst.Id())
```
gordon-klotho, line 52: nice!

pkg/graph/graph.go
```diff
@@ -239,7 +243,7 @@ func handleIncomingEdges[V Identifiable, O any](d *Directed[V], to V, generate f
 	for _, v := range fullAdjacency {
 		for _, edge := range v {
 			if edge.Target != to.Id() {
-				zap.S().Debugf(`Ignoring unexpected edge source from %v`, edge)
+				// zap.S().Debugf(`Ignoring unexpected edge target from %v`, edge)
```
yuval-klotho, line 246: intentionally commented out? if so, just rm it.

pkg/core/resource_graph.go
yuval-klotho, line 52: I think we should add `(internal error)` or such

pkg/provider/aws/execution_unit.go
```diff
@@ -289,10 +303,18 @@ func (a *AWS) convertExecUnitParams(result *core.ConstructGraph, dag *core.Resou
 		resourceEnvVars["EXECUNIT_NAME"] = core.IaCValue{
 			Property: unit.ID,
 		}
-
 		// Retrieve the actual resource and set the environment variables on it
-		resList, ok := a.GetResourcesDirectlyTiedToConstruct(unit)
-		if !ok {
+		resList, _ := a.GetResourcesDirectlyTiedToConstruct(unit)
```
yuval-klotho, line 307: if we don't care about the `ok` arg, maybe just rm it?

gordon-klotho (approved):

pkg/graph/graph.go
jhsinger-klotho: nope, let me add it back, was just endlessly printing because of circular dep. Going to add to this pr though as well

pkg/core/resource_graph.go
jhsinger-klotho: yeah i think well need to clean this up, ill add a fit and finish task, but for now want this in because its hard to trace these down if theyre from many commits ago

yuval-klotho: I just meant in the error text. If a user sees this, they're going to wonder what they did to create the cycle, and how to fix it; we should let them know it's a bug on our end.

PR #518, opened 2023-04-17
Adds EKS VPC CNI addon and NVIDIA device plugin

DavidSeptimus-Klotho:
This PR is associated with #507.

It adds support for the following:
- Installing the NVIDIA Device k8s plugin when any of an EKS cluster's node groups use a gpu-enabled AMI type.
- Installing the EKS VPC CNI addon on all EKS clusters

Fixes:
- Added  missing `ClustersProvider` reference for fluentbit

### DIFF
--- a/pkg/provider/aws/resources/eks.go
+++ b/pkg/provider/aws/resources/eks.go
@@ -22,6 +22,7 @@ const (
 	EKS_FARGATE_PROFILE_TYPE = "eks_fargate_profile"
 	EKS_NODE_GROUP_TYPE      = "eks_node_group"
 	DEFAULT_CLUSTER_NAME     = "eks-cluster"
+	EKS_ADDON_TYPE           = "eks_addon"
 
 	CLUSTER_OIDC_URL_IAC_VALUE = "cluster_oidc_url"
 	CLUSTER_OIDC_ARN_IAC_VALUE = "cluster_oidc_arn"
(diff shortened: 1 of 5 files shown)


pkg/provider/aws/resources/eks.go
```diff
@@ -326,6 +342,25 @@ func (cluster *EksCluster) GetOutputFiles() []core.File {
...
+		},
+	}
+
+	dag.AddResource(manifest)
+	dag.AddDependenciesReflect(manifest)
+
+	for _, ng := range cluster.getClustersNodeGroups(dag) {
+		dag.AddDependency(manifest, ng)
+	}
+}
```
jhsinger-klotho, line 362: the references here may not be correct because were passing in the clusters references and not the node groups which caused this creation. We ideally want the ref to just be exec units that need this. Not blocking for now, but better for tracking

jhsinger-klotho (approved):
one minor comment which would be good to address but wont consider blocking

pkg/provider/aws/resources/eks.go
```diff
@@ -326,6 +342,25 @@ func (cluster *EksCluster) GetOutputFiles() []core.File {
...
+		ConstructRefs: references,
+		FilePath:      "https://raw.githubusercontent.com/NVIDIA/k8s-device-plugin/v1.10/nvidia-device-plugin.yml",
+		ClustersProvider: core.IaCValue{
+			Resource: cluster,
+			Property: CLUSTER_PROVIDER_IAC_VALUE,
+		},
+	}
+
+	dag.AddResource(manifest)
+	dag.AddDependenciesReflect(manifest)
```
gordon-klotho, line 357: nit: you don't need `AddResource` if you call `AddDependenciesReflect`

gordon-klotho (approved):

Issue #502, opened 2023-04-13
Add core.Config to create secrets in IAC2


jhsinger-klotho:
Currently our `GenerateSecretsResources` method is made specifically for the core.Secrets struct. We also want to create a secret when someone specifies `klotho::config { secret = true}`, since this is what is used in golang. The difficulty here is that we allow non static paths so we cannot always determine where the path to the secret is.

For now we could force that to be input within our klotho configuration if necessary.

Definition of Done:
- the klotho config capability outputs the secret and secret version as expected

gordon-klotho:
https://github.com/klothoplatform/klotho/pull/519

PR #519, opened 2023-04-17
Config secrets

gordon-klotho:
Support config secrets for https://github.com/klothoplatform/klotho/issues/502

from `go-secrets` sample app
```ts
const awsSecretGgDevMySecretMySecretKey = new aws.secretsmanager.Secret(
        `gg-dev-mySecret-my_secret.key`,
        {
            name: `gg-dev-mySecret-my_secret.key`,
            recoveryWindowInDays: 0,
        },
        { protect: protect }
    );

if (fs.existsSync(`my_secret.key`)) {
        new aws.secretsmanager.SecretVersion(
            `gg-dev-mySecret-my_secret.key`,
            {
                secretId: awsSecretGgDevMySecretMySecretKey.id,
                secretBinary: fs.readFileSync(`my_secret.key`, 'base64').toString(),
                
            },
            { protect: protect }
        )
    };

const awsIamPolicyGgDevConfigMySecret = new aws.iam.Policy(`gg-dev-config_mySecret`, {
        policy: {Version: `2012-10-17`,
Statement: [{Effect: `Allow`,
Action: [`secretsmanager:DescribeSecret`,`secretsmanager:GetSecretValue`],
Resource: [awsSecretGgDevMySecretMySecretKey.arn],
}],
},
    });
```

### Standard checks

### DIFF
--- a/pkg/infra/iac2/resource_creation_template.go
+++ b/pkg/infra/iac2/resource_creation_template.go
@@ -224,6 +224,9 @@ func deduplicateAppliedOutputs(outputs []AppliedOutput) ([]AppliedOutput, error)
 	return uniqueList, nil
 }
 
+// bodyContents returns the contents of a 'statement_block' with the surrounding {}
+// and indentation removed so that the contents of a void function
+// can be inlined with the rest of the index.ts.
 func bodyContents(node *sitter.Node) string {
 	if node.ChildCount() == 0 || node.Child(0).Content() != "{" {
 		return node.Content()
@@ -236,5 +239,5 @@ func bodyContents(node *sitter.Node) string {
 		}
 		buf.WriteString(node.NamedChild(i).Content())
 	}
-	return strings.TrimSuffix(buf.String(), ";")
+	return strings.TrimSuffix(buf.String(), ";") // Remove any trailing ';' since one is added later to prevent ';;'
 }
--- a/pkg/infra/iac2/resource_creation_template.go
+++ b/pkg/infra/iac2/resource_creation_template.go
@@ -106,11 +106,11 @@ func ParseResourceCreationTemplate(name string, contents []byte) ResourceCreatio
 	result.OutputType = create["return_type"].Content()
 	var expressionBody string
 	if result.OutputType == "void" {
-		expressionBody = create["body"].Content()
+		expressionBody = bodyContents(create["body"])
 	} else {
 		body, found := doQuery(create["body"], findReturn)()
 		if !found {
-			panic(errors.Errorf("No 'return' found in body in %s", name))
+			panic(errors.Errorf("No 'return' found in %s body : `%s`", name, create["body"].Content()))
 		}
 		expressionBody = body["return_body"].Content()
 	}
@@ -222,4 +222,19 @@ func deduplicateAppliedOutputs(outputs []AppliedOutput) ([]AppliedOutput, error)
 		}
 	}
 	return uniqueList, nil
+}
+
+func bodyContents(node *sitter.Node) string {
+	if node.ChildCount() == 0 || node.Child(0).Content() != "{" {
+		return node.Content()
+	}
+	var buf strings.Builder
+	buf.Grow(len(node.Content()))
+	for i := 0; i < int(node.NamedChildCount()); i++ {
+		if i > 0 {
+			buf.WriteRune('\n')
+		}
+		buf.WriteString(node.NamedChild(i).Content())
+	}
+	return strings.TrimSuffix(buf.String(), ";")
 }
--- a/pkg/infra/iac2/resource_creation_template.go
+++ b/pkg/infra/iac2/resource_creation_template.go
@@ -51,7 +51,7 @@ var (
 
 	parameterizeArgsRegex = regexp.MustCompile(`\bargs(\.\w+)`)
 	curlyEscapes          = regexp.MustCompile(`({+)(args\.)`)
-	templateComments      = regexp.MustCompile(`//*TMPL\s+(.*)`)
+	templateComments      = regexp.MustCompile(`//*TMPL\s+`)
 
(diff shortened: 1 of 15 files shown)


pkg/infra/iac2/templates/secret_version/factory.ts
```diff
@@ -10,17 +10,19 @@ interface Args {
...
-            //TMPL {{ if eq .Type.Raw "string" }}
-            //TMPL secretString: fs.readFileSync(args.Path, 'utf-8').toString()
-            //TMPL {{ else }}
-            secretBinary: fs.readFileSync(args.Path, 'base64').toString(),
-            //TMPL {{ end }}
-        },
-        { protect: args.protect }
-    )
+function create(args: Args): void {
+    if (fs.existsSync(args.Path)) {
```
jhsinger-klotho, line 14: this will cause failures if we return void since other things may depend on the secret. I ran into this earlier, so here we would regress and not be deployable

pkg/provider/aws/config_resources.go
```diff
@@ -0,0 +1,53 @@
...
+					return errors.Errorf("expected resource with id, %s, to be an iam policy", res.Id())
+				}
+			} else {
+				dag.AddResource(policy)
+				dag.AddDependency(policy, secretVersion)
+				a.PolicyGenerator.AddAllowPolicyToUnit(unit.Id(), policy)
+			}
+		}
+		return nil
+	}
```
jhsinger-klotho, line 50: why not break out the generateSecrets function so we can just reuse it rather than duplicating it if possible?

pkg/infra/iac2/resource_creation_template.go
```diff
@@ -210,3 +223,18 @@ func deduplicateAppliedOutputs(outputs []AppliedOutput) ([]AppliedOutput, error)
...
+	}
+	var buf strings.Builder
+	buf.Grow(len(node.Content()))
+	for i := 0; i < int(node.NamedChildCount()); i++ {
+		if i > 0 {
+			buf.WriteRune('\n')
+		}
+		buf.WriteString(node.NamedChild(i).Content())
+	}
+	return strings.TrimSuffix(buf.String(), ";")
```
jhsinger-klotho, line 239: nit: not quite sure what this is doing, could we add a comment?

pkg/provider/aws/config_resources.go
```diff
@@ -0,0 +1,53 @@
...
+			if res := dag.GetResource(policy.Id()); res != nil {
+				if existingPolicy, ok := res.(*resources.IamPolicy); ok {
+					existingPolicy.Policy.Statement = append(existingPolicy.Policy.Statement, policyDoc.Statement...)
+					dag.AddDependency(existingPolicy, secretVersion)
+				} else {
+					return errors.Errorf("expected resource with id, %s, to be an iam policy", res.Id())
+				}
+			} else {
+				dag.AddResource(policy)
+				dag.AddDependency(policy, secretVersion)
```
yuval-klotho, line 45: any reason not to use `AddResourceReflect`?

pkg/provider/aws/config_resources.go
```diff
@@ -0,0 +1,53 @@
...
+				a.PolicyGenerator.AddAllowPolicyToUnit(unit.Id(), policy)
+			}
+		}
+		return nil
+	}
+
+	return errors.New("unsupported")
```
yuval-klotho, line 52: Is this something a user could trigger, or does it imply a bug on our part?

If it's something a user could trigger, we should make sure the error message is understandable and actionable (`"unsupported"` by itself isn't, but I'm not sure if this would then be wrapped in some context that would make it be). If it's an internal bug, I like specifying that in the error text (though we don't have any official policy about that).

pkg/infra/iac2/templates/secret_version/factory.ts
gordon-klotho: It won't cause any failures because the dependency was for the policy which was erroneously pointing to the version's ARN instead of the secret's ARN. This is corrected in this PR (though I only fixed it for config, I'll go fix that for secret now) and the resulting index.ts has no compilation errors.

pkg/provider/aws/config_resources.go
gordon-klotho: Nope, was just copy/paste. I fixed when I factored it out.

pkg/provider/aws/config_resources.go
gordon-klotho: Yeah, I'll do that

pkg/infra/iac2/templates/secret_version/factory.ts
jhsinger-klotho: well for rds we depend on the secret version and need that secret versions arn i thought? Or does it automatically work if you put the secrets arn?

jhsinger-klotho (approved):
Approved, but can you run this on the go-secrets app before pushing? or i think with this we should be good to test the go-mega-app so thats also an option (minues a few bug fixes which david and i have in flight)

pkg/infra/iac2/templates/secret_version/factory.ts
gordon-klotho: Yeah, the way it works is access is done at the `Secret` level and the `SecretVersion` is only used if you need something other than "latest" or to write a new version. In fact, looks like RDS already correctly uses secret:
```go
secretPolicyDoc := CreateAllowPolicyDocument([]string{"secretsmanager:GetSecretValue"}, []core.IaCValue{{Resource: secret, Property: ARN_IAC_VALUE}})
```

gordon-klotho:
Successful test using go-secret

image

1. Initial deploy skipped the SecretVersion because the file did not exist. Curl results in error
2. Copy secret in
3. Deploy shows new SecretVersion. Curl succeeds

PR #521, opened 2023-04-17
(optional) refactor exec unit transformation functions

yuval-klotho:
This PR is optional cleanup; we should only merge it if consensus is that it improves readability.

This pulls all the common code in the transformers (reading, writing, casting, error handling, etc.) to a single method, so that each transformer focuses on just the logic it needs to provide.

> **Note**
> A lot of this diff is indentation change, so I recommend viewing with "ignore whitespace"

### Standard checks

- **Unit tests**: passing
- **Docs**: n/a

### DIFF
--- a/pkg/infra/kubernetes/exec_unit.go
+++ b/pkg/infra/kubernetes/exec_unit.go
@@ -55,11 +55,8 @@ func (transformer manifestTransformer[K]) apply(unit *HelmExecUnit, cfg config.E
 	}
 	transformObj, ok := obj.(K)
 	if !ok {
-		t := reflect.TypeOf((*K)(nil))
-		for t.Kind() == reflect.Pointer {
-			t = t.Elem()
-		}
-		err = fmt.Errorf("expected file %s to contain %s Kind", source.Path(), t.Name())
+		var k K
+		err = fmt.Errorf("expected file %s to contain %T Kind", source.Path(), t)
 		return nil, err
 	}
 
--- a/pkg/infra/kubernetes/exec_unit.go
+++ b/pkg/infra/kubernetes/exec_unit.go
@@ -3,11 +3,12 @@ package kubernetes
 import (
 	"errors"
 	"fmt"
+	autoscaling "k8s.io/api/autoscaling/v2beta2"
+	"k8s.io/apimachinery/pkg/runtime"
 	"regexp"
 
 	"go.uber.org/zap"
 	apps "k8s.io/api/apps/v1"
-	autoscaling "k8s.io/api/autoscaling/v2beta2"
 	corev1 "k8s.io/api/core/v1"
 	elbv2api "sigs.k8s.io/aws-load-balancer-controller/apis/elbv2/v1beta1"
 	"sigs.k8s.io/yaml"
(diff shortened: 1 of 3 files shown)


yuval-klotho:
Oh, and to mention: an alternate form of this could be to define the transformers via an interface instead of fields; it would be much the same (a method for name, a method for getting the `*SourceFile`, a method for the transform), and at the call site, instead of something like

```
deploymentValues, err := deploymentTransformer.apply(unit, cfg)
```

you would have

```
deploymentValues, err := unit.transform(deploymentTransformer, cfg)
```

If people prefer that, I can easily make the change.

pkg/infra/kubernetes/exec_unit.go
```diff
@@ -3,11 +3,12 @@ package kubernetes
 import (
 	"errors"
 	"fmt"
+	autoscaling "k8s.io/api/autoscaling/v2beta2"
+	"k8s.io/apimachinery/pkg/runtime"
```
gordon-klotho, line 7: nit: would it be possible to standardize to `gopls` import ordering and grouping? I believe `goimports` is the equivalent that intellij can use

gordon-klotho (commented):
I haven't looked it over for correctness, but the pattern I'm 👍 on

gordon-klotho:
Oh, one thing that I wanted to do, at some point, and might touch on this (so you may or may not want to consider) is that I wanted to change all the `HelmExecUnit` fields to be their actual k8s types instead of in yaml form as `*core.SourceFile`. This would make it much easier to work with, reason about, and potentially modify. We could wrap the k8s types with one that implements `core.File` by marshalling to yaml on WriteTo. Not sure if that affects this at all, and even if it does that doesn't necessarily mean we should do anything about it now.

yuval-klotho:
> Oh, one thing that I wanted to do, at some point, and might touch on this (so you may or may not want to consider) is that I wanted to change all the HelmExecUnit fields to be their actual k8s types instead of in yaml form as *core.SourceFile.

Yeah, I agree it would make them much nicer to work with. I would like to keep that off of this PR, though, just for the sake of keeping it smaller (this is a relatively fast-moving part of the code base). I think this change is compatible with that idea though.

pkg/infra/kubernetes/exec_unit.go
```diff
@@ -30,6 +33,52 @@ type HelmExecUnit struct {
...
+	transform        func(unit *HelmExecUnit, cfg config.ExecutionUnit, target K, log *zap.SugaredLogger) ([]HelmChartValue, error)
+	// readF is the function to read a given file to the runtime.Object that we will then cast down to K. You may leave
+	// it blank, in which case the transformer will assume readFile.
+	readF func(f *core.SourceFile) (runtime.Object, error)
+}
+
+func (transformer manifestTransformer[K]) apply(unit *HelmExecUnit, cfg config.ExecutionUnit) ([]HelmChartValue, error) {
+	source := transformer.fieldToTransform(unit)
+	log := zap.L().Sugar().With(logging.FileField(source), zap.String("unit", unit.Name))
+	log.Debugf("Transforming file, %s, for exec unit, %s", source.Path(), unit.Name)
```
gordon-klotho, line 47: nit: the unit and source are redundant since they're already fields on the logger.

pkg/infra/kubernetes/exec_unit.go
```diff
@@ -30,6 +33,52 @@ type HelmExecUnit struct {
...
+	if err != nil {
+		return nil, err
+	}
+	transformObj, ok := obj.(K)
+	if !ok {
+		t := reflect.TypeOf((*K)(nil))
+		for t.Kind() == reflect.Pointer {
+			t = t.Elem()
+		}
+		err = fmt.Errorf("expected file %s to contain %s Kind", source.Path(), t.Name())
```
gordon-klotho, line 62:
```suggestion
		var k K
		err = fmt.Errorf("expected file %s to contain %T Kind", source.Path(), t)
```
nit: a little more succinct

pkg/infra/kubernetes/exec_unit.go
```diff
@@ -76,197 +125,165 @@ func shouldTransformServiceAccount(unit *core.ExecutionUnit) bool {
...
+		output, err := yaml.Marshal(pod)
+		if err != nil {
+			return nil, err
+		}
+		err = unit.Pod.Reparse([]byte(output))
+		if err != nil {
+			return nil, err
+		}
+		return []HelmChartValue{
+			HelmChartValue{
```
gordon-klotho, line 154: I think linter would be unhappy about this (the `HelmChartValue` can be omitted)

gordon-klotho (approved):

pkg/infra/kubernetes/exec_unit.go
yuval-klotho: We do run `golangci-lint`, and it didn't complain, but I can rm it.

pkg/infra/kubernetes/exec_unit.go
yuval-klotho: Your version is a bit different, but probably good enough to warrant the succinctness. The output is:

```diff
- expected file test/templates/unit-deployment.yaml to contain Deployment Kind
+ expected file test/templates/unit-deployment.yaml to contain *v1.Deployment Kind
```

pkg/infra/kubernetes/exec_unit.go
yuval-klotho: This is what they were before, so I'll keep them for now — I don't feel like investigating what the UX difference would be. :)

pkg/infra/kubernetes/exec_unit.go
gordon-klotho: Huh, nevermind then. I swear I've done similar things and it complained to me. 🤷

PR #523, opened 2023-04-18
remove applied outputs and k8s related fixes

jhsinger-klotho:

lots of fixes from deploying a k8s mega app

- Fixes the addition of the openIdConnectProvider
- removes using aws-native for now to prevent the need to set region
- removed the appliedOutputs since it doesnt work anyways when set on a field. using pulumi.Jsonstringify instead
- other small fixes to api integration/other resources

### Standard checks

### DIFF
--- a/pkg/infra/iac2/templates_compiler.go
+++ b/pkg/infra/iac2/templates_compiler.go
@@ -55,7 +55,6 @@ type (
 	}
 	NestedCtx struct {
 		useDoubleQuotes bool
-		appliedOutputs  *[]AppliedOutput
 		rootVal         *reflect.Value
 	}
 )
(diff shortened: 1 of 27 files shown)


yuval-klotho (approved):

DavidSeptimus-Klotho (approved):

pkg/infra/iac2/templates_compiler.go
```diff
@@ -278,20 +259,24 @@ func (tc TemplatesCompiler) resolveDependencies(resource core.Resource) string {
 	numDeps := len(upstreamResources)
 	for i := 0; i < numDeps; i++ {
 		res := upstreamResources[i]
-		buf.WriteString(tc.getVarName(res))
-		if i < (numDeps - 1) {
-			buf.WriteRune(',')
+		switch res.(type) {
+		case *resources.Region, *resources.AvailabilityZones, *resources.AccountId:
```
gordon-klotho, line 263: (non-blocking) I wonder if there's a better way to tell which resources to ignore rather than hardcoding.

gordon-klotho (approved):

DavidSeptimus-Klotho (approved):

pkg/infra/iac2/templates_compiler.go
jhsinger-klotho: yeah im not sure, but its iac specific so i wouldnt want to include it on the actual resource. i also dont think it will be more than just these 3 for a while unless we end up having more special cased things

Issue #524, opened 2023-04-18
HTTP_PROXY fails with wildcard routes


yuval-klotho:
As reported [on discord][1], wildcard routes break on ECS and EKS because of how we set up the integration.

The [AWS docs explain][2]:

> To set up a route with an HTTP proxy integration that catches all requests, create an API route with a greedy path variable (for example, `/parent/{proxy+}`). Integrate the route with an HTTP endpoint (for example, `https://petstore-demo-endpoint.execute-api.com/petstore/{proxy}`) on the ANY method.

We are adding the `+` on both the route and the integration today. We should only be adding it on the route.

We should:

- [x] fix it in legacy iac
- [x] fix it in iac2
- [ ] (optional) add an integration test for it

[1]: https://discord.com/channels/870150286059114576/896496825681453158/1095817894740373634
[2]: https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-http.html

yuval-klotho:
I'm spinning the integ test task off to https://github.com/klothoplatform/infra-tests/issues/20. We're working on integ tests in general as part of the iac2 work, and I don't want to potentially conflict with that work. Plus, I think there's a larger question of whether we really want to rely on integration tests for all these sorts of permutations — if we do, the integ test suite could get very large.

With that task spun off to its own ticket, this one is now done.

PR #527, opened 2023-04-18
don't use greedy vars in integration routes

yuval-klotho:
I had considered also adding a fuller test of the APIG and all that, but it's a _lot_ of setup and makes for a pretty  unreadable test. Instead, I added just a simpler unit test of `convertPath`

In service of #524.

### Standard checks

- **Unit tests**: added
- **Docs**: n/a

### DIFF
--- a/pkg/provider/aws/expose_test.go
+++ b/pkg/provider/aws/expose_test.go
@@ -1,7 +1,6 @@
 package aws
 
 import (
-	"fmt"
 	"testing"
 
 	"github.com/klothoplatform/klotho/pkg/annotation"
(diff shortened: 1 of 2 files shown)


pkg/provider/aws/expose_test.go
```diff
@@ -389,3 +390,54 @@ func Test_CreateRestApi(t *testing.T) {
...
+					actual := convertPath(tt.given, greedy)
+					var want string
+					if greedy {
+						want = tt.wantIfGreedy
+					} else {
+						want = tt.wantIfNoGreedy
+					}
+					assert.Equal(want, actual)
+				})
+			}
```
gordon-klotho, line 440: very nitpick: I'd prefer just writing this out, doesn't seem like too much to warrant the for loop and not like we'd ever add more things to that list.
```go
assert := assert.New(t)

actualGreedy := convertPath(tt.given, true)
assert.Equal(tt.wantIfGreedy, actualGreedy, "greedy")

actualNotGreedy := convertPath(tt.given, false)
assert.Equal(tt.wantIfNoGreedy, actualNotGreedy, "not greedy")
```

gordon-klotho (approved):

PR #528, opened 2023-04-18
stable ordering of the graph topology

### LINKED ISSUE #459: provide stable ordering of generated IaC across all runs
The trickiest big will be the topological ordering. The library we currently use doesn't support it, so we would need to either work around that or propose a change to the library.

Some options:

* propose a change to the library (we should first check to see if they'd be willing)
* soft-fork the library to add our change, even if we don't get it into upstream; or, just cp the `TopologicalSort` method specifically into our code base, and use that rather than the library's
* use a different library

yuval-klotho:
I think the topological sorting is currently the only non-deterministic aspect of our output (citation needed?), in which case this resolves #459.

This provides a stable topological ordering by copying the graph lib's `TopologicalSort`, and replacing the two `range`s with an iterator impl. The hope is that this approach might be something the upstream lib would be willing to accept as a change, but for now I don't want to block on that.

### Standard checks

- **Unit tests**: I added a test that gets the vertices in topological order, and then repeats that 10k times and ensures that none of those have different order. It failed before this change, and then succeeded after. (The chance of it succeeding even if the order is random is 0.5<sup>10,000</sup>, which is basically 0.)
- **Docs**: n/a


pkg/graph/topological_test.go
```diff
@@ -0,0 +1,59 @@
...
+	g := NewDirected[dummyVertex]()
+	g.AddVerticesAndEdge("a", "c")
+	g.AddVerticesAndEdge("b", "c")
+
+	archetype, err := g.VertexIdsInTopologicalOrder()
+	if !assert.NoError(err) {
+		return
+	}
+	archetypeStr := strings.Join(archetype, "")
+	assert.True(archetypeStr == "abc" || archetypeStr == "bac")
```
gordon-klotho, line 23: Shouldn't this be in a single specific ordering to make sure it doesn't flip across compilations?

yshavit: I was trying to convey in the test that it doesn't really matter _which_ it is, as long as it's then the same for the next 10k. I could be even more explicit, to make sure that it's stable not just across invocations but also across processes.

yshavit: https://github.com/klothoplatform/klotho/pull/530

PR #531, opened 2023-04-18
make test more explicit

yuval-klotho:
A more explicit `expected` that ensures stability across processes (and indeed even across Go versions). See https://github.com/klothoplatform/klotho/pull/528.

### Standard checks

- **Unit tests**: got 'em
- **Docs**: n/a


gordon-klotho (approved):

yuval-klotho:
(This is the same as #530, just with the right GH account this time)

Issue #477, opened 2023-04-06
Defaults config regression


gordon-klotho:
With config:
```toml
[defaults.execution_unit]
  type = 'kubernetes'
```

Get the following error:
```
 error error in plugin go/AddExecRuntimeFiles:Go: unsupported execution unit type: ''
github.com/klothoplatform/klotho/pkg/lang/golang/aws_runtime.(*AwsRuntime).AddExecRuntimeFiles
        /home/gordon/workspace/klotho/pkg/lang/golang/aws_runtime/aws.go:50
github.com/klothoplatform/klotho/pkg/lang/golang.(*AddExecRuntimeFiles).Transform
        /home/gordon/workspace/klotho/pkg/lang/golang/plugin_add_exec_runtime_files.go:25
github.com/klothoplatform/klotho/pkg/lang/golang.GoPlugins.Transform
        /home/gordon/workspace/klotho/pkg/lang/golang/plugins.go:33
github.com/klothoplatform/klotho/pkg/compiler.(*Compiler).Compile
        /home/gordon/workspace/klotho/pkg/compiler/compiler.go:65
github.com/klothoplatform/klotho/pkg/cli.KlothoMain.run
        /home/gordon/workspace/klotho/pkg/cli/klothomain.go:415
github.com/spf13/cobra.(*Command).execute
        /home/gordon/.gvm/pkgsets/go1.20.2/global/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:916
github.com/spf13/cobra.(*Command).ExecuteC
        /home/gordon/.gvm/pkgsets/go1.20.2/global/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:1044
github.com/spf13/cobra.(*Command).Execute
        /home/gordon/.gvm/pkgsets/go1.20.2/global/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:968
github.com/klothoplatform/klotho/pkg/cli.KlothoMain.Main
        /home/gordon/workspace/klotho/pkg/cli/klothomain.go:129
main.main
        /home/gordon/workspace/klotho/cmd/klotho/main.go:20
runtime.main
        /home/gordon/.gvm/gos/go1.20.2/src/runtime/proc.go:250
runtime.goexit
        /home/gordon/.gvm/gos/go1.20.2/src/runtime/asm_amd64.s:1598
```

jhsinger-klotho:
[...]

PR #536, opened 2023-04-19
Apply defaults even when specific config is present

gordon-klotho:
Resolves https://github.com/klothoplatform/klotho/issues/477

With config
```toml
[defaults.execution_unit]
type = 'kubernetes'

[execution_units.main.infra_params]
instance_type = "g3s.xlarge"
```

Before, this results in an error that type cannot be "".
After, this compiles successfully.

### Standard checks


pkg/config/config.go
```diff
@@ -255,60 +243,72 @@ func ConvertToInfraParams(p any) InfraParams {
...
 
-func merge(dst, src map[string]interface{}, depth int) map[string]interface{} {
+// merge applies all k,v pairs from src into dst. If the value is a map, it will
+// try to recusively merge those as well. When keys conflict, dst will win out
+// because this is used for ApplyDefaults where dst is the specific annotation's
+// values and src is the default values.
+func merge(dst, src map[string]interface{}, depth int) {
 	if depth > MaxDepth {
 		panic("too deep!")
 	}
```
jhsinger-klotho, line 295: nit: I realize you didnt write this, but should we change the panic message. Kind of a funny message and shouldnt ever happen, but would be weird for a user to see that

jhsinger-klotho (approved):

PR #539, opened 2023-04-24
be able to parse constsruct graph for a run

jhsinger-klotho:

Mainly want some opinions, but this gets us to a full end to end flow we are looking for in ICoP

Introduces a new hidden flag for `construct-graph` which takes in a path to a file.
If the flag is passed, we remove all analysis and transformation plugins from the compiler.
We create the construct graph from the file passed in and run provider + IaC Plugins.

### Standard checks

- **Unit tests**: Any special considerations? will add, but wanted to make sure this approach can get some alignment

### DIFF
--- a/pkg/core/construct_graph.go
+++ b/pkg/core/construct_graph.go
@@ -1,8 +1,13 @@
 package core
 
 import (
+	"os"
+
+	"github.com/klothoplatform/klotho/pkg/annotation"
 	"github.com/klothoplatform/klotho/pkg/graph"
+	"github.com/pkg/errors"
 	"go.uber.org/zap"
+	"gopkg.in/yaml.v2"
 )
 
 type (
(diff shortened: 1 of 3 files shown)


pkg/core/construct_graph.go
```diff
@@ -116,3 +121,82 @@ func (cg *ConstructGraph) FindUpstreamGateways(unit *ExecutionUnit) []*Gateway {
...
+			keys[id] = unit.AnnotationKey
+		case "orm":
+			unit := &Orm{AnnotationKey: AnnotationKey{ID: id, Capability: annotation.PersistCapability}}
+			graph.AddConstruct(unit)
+			keys[id] = unit.AnnotationKey
+		case "kv":
+			unit := &Kv{AnnotationKey: AnnotationKey{ID: id, Capability: annotation.PersistCapability}}
+			graph.AddConstruct(unit)
+			keys[id] = unit.AnnotationKey
+		case "redis_node":
```
gordon-klotho, line 176: nit: missing `"redis_cluster'`?

gordon-klotho (approved):

pkg/core/construct_graph.go
jhsinger-klotho: in ifc2 we dont support redis_cluster so i didnt add it

PR #540, opened 2023-04-26
Ifc2/eks integ test fixes

DavidSeptimus-Klotho:
Implements a variety of fixes required to get our EKS integration test applications working.

Check the individual commit messages for additional details. Reviewing the commits individually is recommended.


jhsinger-klotho (approved):

pkg/infra/iac2/resource_creation_template.go
```diff
@@ -178,6 +183,54 @@ func parseTS(val reflect.Value) (string, error) {
 	return "", fmt.Errorf("invalid template value detected: %s: %v: template values must implement the templateValue interface", val.Kind().String(), val.Interface())
 }
 
+func appliedOutputsToString(outputs []AppliedOutput) string {
```
gordon-klotho, line 186: I thought Jordan removed this - is this a merge/rebase error?

DavidSeptimus-Klotho: I added it back. There are still cases where something like this is needed. An example of this is where you need to use a pulumi output as an attribute name in an object.

here's an example of how it's used:
```ts
const myTaskExecutionRole = new aws.iam.Role(`my-task-ExecutionRole`, {
    assumeRolePolicy: pulumi.jsonStringify(
      // we invoke .apply() on a specific applied output instead of using pulumi.All() when there's only 1 applied output
      myEksCluster.url.apply(cluster_oidc_url => {
          return {
              Version: `2012-10-17`,
              Statement: [{
                  Effect: `Allow`,
                  Principal: {
                      Federated: myEksCluster.arn,
                  },
                  Condition: {
                      StringEquals: {
                          // cluster_oidc_url needs to be an actual string here
                          [`${cluster_oidc_url}:sub`]: `system:serviceaccount:default:my-task`,
                          [`${cluster_oidc_url}:aud`]: `sts.amazonaws.com` 
                      },
                  },
              }],
          }
    })),
```

PR #548, opened 2023-04-26
fix our fluent bit cm for fargate

jhsinger-klotho:

we need to input the region in to the config map so do that via transformation
we also can name the log group on a cluster by cluster basis

### Standard checks

### DIFF
--- a/pkg/infra/iac2/templates_compiler.go
+++ b/pkg/infra/iac2/templates_compiler.go
@@ -619,6 +619,10 @@ func (tc TemplatesCompiler) handleIaCValue(v core.IaCValue, appliedOutputs *[]Ap
 		}
 	case resources.CIDR_BLOCK_IAC_VALUE:
 		return fmt.Sprintf(`%s.cidrBlock`, tc.getVarName(resource)), nil
+	case resources.AWS_OBSERVABILITY_CONFIG_MAP_REGION_IAC_VALUE:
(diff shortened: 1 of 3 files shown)


pkg/infra/iac2/templates_compiler.go
```diff
@@ -619,6 +619,10 @@ func (tc TemplatesCompiler) handleIaCValue(v core.IaCValue, appliedOutputs *[]Ap
 		}
 	case resources.CIDR_BLOCK_IAC_VALUE:
 		return fmt.Sprintf(`%s.cidrBlock`, tc.getVarName(resource)), nil
+	case resources.AWS_OBSERVABILITY_CONFIG_MAP_REGION_IAC_VALUE:
+		region := resources.NewRegion()
+		return fmt.Sprintf(`pulumi.all([obj.data["output.conf"], %s.name, %s.name]).apply(([obj, regionName, clusterName]) => obj.replace("region-code",regionName).replace("my-logs","/fargate/" +clusterName))`,
```
DavidSeptimus-Klotho, line 624: so ugly!

DavidSeptimus-Klotho (approved):

pkg/infra/iac2/templates_compiler.go
jhsinger-klotho: indeed

PR #551, opened 2023-04-26
Copies config secret path when overriding values

DavidSeptimus-Klotho:
This PR copies over the `Config` struct's `Path` field when merging config overrides.


jhsinger-klotho (approved):

gordon-klotho (approved):
I regressed my own change, oops. Good catch

PR #562, opened 2023-04-27
Change python fargate dockerfile to be debian based

gordon-klotho:
Tested with py-redis
image

Can't get a request through the API Gateway yet though - likely due to unrelated issues, but want to get a complete test before it's ready for merge.

### Standard checks


gordon-klotho:
Confirmed working requests on py-redis.

PR #566, opened 2023-05-01
Detect more import statements for python dependencies

gordon-klotho:
Tested with a modified py-redis to use `from shared import data` from `app/main.py`.
Source:
image
Compiled:
image

### Standard checks

- **Unit tests**: Updated all tests, added new cases to cover new functionality

### DIFF
--- a/pkg/lang/python/imports_test.go
+++ b/pkg/lang/python/imports_test.go
@@ -19,44 +19,44 @@ func TestFindImports(t *testing.T) {
 			name:   "import module",
 			source: "import mymodule",
 			want: map[string]Import{
-				"mymodule": {Name: "mymodule", ImportedSelf: true},
+				"mymodule": {Name: "mymodule"},
 			},
 		},
 		{
 			name:   "import module aliased",
 			source: "import mymodule as m",
 			want: map[string]Import{
-				"mymodule": {Name: "mymodule", ImportedSelf: true, Alias: "m"},
+				"mymodule": {Name: "mymodule", Alias: "m"},
 			},
 		},
 		{
 			name:   "import modules",
 			source: "import mymodule1, mymodule2",
 			want: map[string]Import{
-				"mymodule1": {Name: "mymodule1", ImportedSelf: true},
-				"mymodule2": {Name: "mymodule2", ImportedSelf: true},
+				"mymodule1": {Name: "mymodule1"},
+				"mymodule2": {Name: "mymodule2"},
 			},
 		},
 		{
 			name:   "import modules aliased",
 			source: "import mymodule1 as m1, mymodule2 as m2",
 			want: map[string]Import{
-				"mymodule1": {Name: "mymodule1", ImportedSelf: true, Alias: "m1"},
-				"mymodule2": {Name: "mymodule2", ImportedSelf: true, Alias: "m2"},
+				"mymodule1": {Name: "mymodule1", Alias: "m1"},
+				"mymodule2": {Name: "mymodule2", Alias: "m2"},
 			},
 		},
 		{
 			name:   "import submodule",
 			source: "import mymodule.submodule",
 			want: map[string]Import{
-				"mymodule.submodule": {ParentModule: "mymodule", Name: "submodule", ImportedSelf: true},
+				"mymodule.submodule": {ParentModule: "mymodule", Name: "submodule"},
 			},
 		},
 		{
 			name:   "import submodule aliased",
 			source: "import mymodule.submodule as w",
 			want: map[string]Import{
-				"mymodule.submodule": {ParentModule: "mymodule", Name: "submodule", ImportedSelf: true, Alias: "w"},
+				"mymodule.submodule": {ParentModule: "mymodule", Name: "submodule", Alias: "w"},
 			},
 		},
 		{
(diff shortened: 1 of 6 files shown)


pkg/lang/python/imports_test.go
```diff
@@ -19,44 +19,44 @@ func TestFindImports(t *testing.T) {
...
-				"mymodule": {Name: "mymodule", ImportedSelf: true},
+				"mymodule": {Name: "mymodule"},
 			},
 		},
 		{
 			name:   "import module aliased",
 			source: "import mymodule as m",
 			want: map[string]Import{
-				"mymodule": {Name: "mymodule", ImportedSelf: true, Alias: "m"},
+				"mymodule": {Name: "mymodule", Alias: "m"},
```
yuval-klotho, line 29: I'm probably missing something, but `"."` seems odd as the key here. What does the key essentially represent? I _thought_ it was "the name that the code will refer to this as" — which here would be the alias, but that doesn't seem to be the case.

What happens if someone does:

```python
# klotho::whatever { id = "my_m" }
import mymodule as m
import mymodule as x

x.foo()
```

Wouldn't this key structure confuse `m` and `x`?

pkg/lang/python/imports_test.go
```diff
@@ -134,14 +128,20 @@ func TestFindImports(t *testing.T) {
...
 						"attribute1": {Name: "attribute1", Alias: "a1"},
 						"attribute2": {Name: "attribute2", Alias: "a2"},
 					}},
 			},
 		},
+		{
+			name:   "import sibling",
+			source: "from . import mymodule",
+			want: map[string]Import{
+				".": {ParentModule: ".", ImportedAttributes: map[string]Attribute{"mymodule": {Name: "mymodule"}}},
```
yuval-klotho, line 142: Similar to my comment above, but now even more easy to trigger: what happens if someone does:

```python
from . import mymodule
from . import anothermodule
```

pkg/lang/python/imports_test.go
```diff
@@ -253,14 +262,14 @@ func TestResolveFileDependencies(t *testing.T) {
 			name: "import full module and use method",
 			input: map[string]string{
 				"main.py": `
-import other
-other.hello_world()
-`,
+		import other
```
yuval-klotho, line 265: minor, but why (a) the indentation change, and (b) indenting to a rather odd column (not full-left, but also not indented relative to the surrounding code)

pkg/lang/python/imports_test.go
```diff
@@ -298,48 +307,74 @@ print(other.hello_world)
...
 				"main.py": map[string]core.References{
-					"other/hello.py": NewSet[string]("say_hi"),
+					"other/hello.py": NewSet("say_hi"),
 				},
 				"other.py": map[string]core.References{},
 			},
 		},
 		{
-			expectFailureDueTo: "#492",
+			expectFailureDueTo: "#492",                                        // TODO https://github.com/klothoplatform/klotho-history/issues/492
```
yuval-klotho, line 366: mionr: why so much &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; whitespace?

pkg/lang/python/imports_test.go
```diff
@@ -606,3 +615,143 @@ func NewSet[K comparable](elements ...K) map[K]struct{} {
...
+			}
+
+			got := referencesForImport(parsed.Tree().RootNode(), tt.importModule)
+			assert.Equal(tt.want, got)
+		})
+	}
+}
+
+func Test_pythonModuleToPath(t *testing.T) {
+	// For ease of understanding, all test cases operate in the following directory structure:
```
yuval-klotho, line 664: 👍 nice!

yuval-klotho (commented):
Did a very cursory look at the prod code, focused mostly on tests. I had a couple questions there, that once I understand what's going on I'll probably circle back and do a more real review of the prod code.

pkg/lang/python/imports_test.go
gordon-klotho: FindImports is just the code parsing, so it's just `from <x> import <y>` would result in `Import{Name: <x>, Attributes: [<y>]}`.

The case you pointed out is tested in case `import multiple attributes from module in separate imports`

pkg/lang/python/imports_test.go
gordon-klotho: Should be the same as `import multiple attributes from module in separate imports` but the key of the import is `"."` instead.

pkg/lang/python/imports_test.go
gordon-klotho: Probably happened when I commented the tests out, then un-commented. I'll revert.

pkg/lang/python/imports_test.go
gordon-klotho: So that it lines up with the comment in the next line. gofmt does that automatically /shrug
I'll put it in the preceding line

pkg/lang/python/imports_test.go
yuval-klotho: Alright, works for me. I added the "bug" label to that ticket, because this really is a bug — not just a lacking feature. But we can discuss it in grooming.

PR #568, opened 2023-05-02
Changes lambda invocation permissions to wildcards per-gateway

DavidSeptimus-Klotho:
Reduces lambda permission clutter and potential policy size overruns for large applications by changing our lambda permissions from 1 per API Gateway integration to 1 per API Gateway to Lambda relationship using wildcards. Now, if you have multiple routes in an API Gateway all invoking the same lambda, you'll only have a single resource-based policy on the lambda allowing invocation from any of the invoking API Gateway's child resources.

### Standard checks

- **Unit tests**: Any special considerations? n/a
- **Backwards compatibility**: Will this break existing apps? If so, what would be the extra work required to keep them working? Backwards compatible. Next deployment of an existing stack will cause existing lambda permissions to be consolidated.

### DIFF
--- a/pkg/infra/pulumi_aws/iac/api_gateway.ts
+++ b/pkg/infra/pulumi_aws/iac/api_gateway.ts
@@ -68,13 +68,16 @@ export class ApiGateway {
         }`
         const lambda = this.lib.execUnitToFunctions.get(r.execUnitName)!
         return new aws.lambda.Permission(
-            `${gwName}-${verb}-${path}`,
+            `${gwName}-http-${r.execUnitName}-invoke-permission`,
             {
-                statementId: `${gwName}-http-${verb}-${r.path.replace(/[^a-zA-Z0-9]+/g, '-')}`,
+                statementId: `${gwName}-httpapi-${r.execUnitName}-invoke`.replace(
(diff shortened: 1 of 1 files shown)


pkg/infra/pulumi_aws/iac/api_gateway.ts
```diff
@@ -68,13 +68,16 @@ export class ApiGateway {
...
-                statementId: `${gwName}-http-${verb}-${r.path.replace(/[^a-zA-Z0-9]+/g, '-')}`,
+                statementId: `${gwName}-httpapi-${r.execUnitName}-invoke`.replace(
+                    /[^a-zA-Z0-9]+/g,
+                    '-'
+                ),
                 action: 'lambda:InvokeFunction',
                 function: lambda,
                 principal: 'apigateway.amazonaws.com',
-                sourceArn: pulumi.interpolate`arn:aws:execute-api:${this.lib.region}:${this.accountId}:${resourceId}`,
+                sourceArn: pulumi.interpolate`${api.executionArn}/*`,
```
gordon-klotho, line 80: nice

gordon-klotho (approved):

PR #572, opened 2023-05-04
Refactor resource ID to be a struct

gordon-klotho:
This is a prerequisite for import work. Since this touched a lot of files, doing this separately to get it in before too many conflicts.

### Standard checks


pkg/core/resource_graph.go
```diff
@@ -15,7 +15,9 @@ type (
 
 func NewResourceGraph() *ResourceGraph {
 	return &ResourceGraph{
-		underlying: graph.NewDirected[Resource](),
+		underlying: graph.NewDirected(func(r Resource) string {
+			return r.Id().String()
+		}),
```
jhsinger-klotho, line 20: im a little bit confused on this change

pkg/infra/iac2/inputs_outputs_test.go
```diff
@@ -150,23 +151,40 @@ func TestKnownTemplates(t *testing.T) {
 					t.Run(inputName, func(t *testing.T) {
 						assert := assert.New(t)
 
-						field, fieldFound := resType.FieldByName(inputName)
+						var inputType reflect.Type
 
-						if !assert.Truef(fieldFound, `missing field`, field.Name) {
+						if field, fieldFound := resType.FieldByName(inputName); fieldFound {
```
yuval-klotho, line 156: Is this part of the refactor, or unrelated additional testing?

pkg/infra/iac2/resources.go
```diff
@@ -35,28 +33,28 @@ type (
...
 
-func (e *KubernetesProvider) Id() string {
-	return fmt.Sprintf("%s:%s:%s", e.Provider(), "kubernetes_provider", e.Name)
-}
-
-func (e *RouteTableAssociation) Provider() string {
-	return "pulumi"
+func (e *KubernetesProvider) Id() core.ResourceId {
+	return core.ResourceId{
+		Provider: "pulumi",
```
yuval-klotho, line 42: While we're here, can we pull this to a `const`?

pkg/infra/iac2/templates_compiler.go
```diff
@@ -174,6 +178,22 @@ func (tc TemplatesCompiler) RenderPackageJSON() (*javascript.NodePackageJson, er
 	return &mainPJson, nil
 }
 
+func validTemplateMethod(method reflect.Value) error {
```
yuval-klotho, line 181: Part of refactor, or unrelated changes?

pkg/core/resource_graph.go
gordon-klotho: Because resources no longer conform to the `Identifiable` interface and the `Id()` signature no longer matches constructs, we need a way to specify the hashing function is a more flexible way (which is also why it's used this way in the underlying graph).

pkg/provider/aws/static_unit_test.go
```diff
@@ -77,19 +76,7 @@ func Test_GenerateStaticUnitResources(t *testing.T) {
 				return
 			}
 
-			tiedResources, found := aws.GetResourcesDirectlyTiedToConstruct(unit)
```
yuval-klotho, line 80: part of refactor?

yuval-klotho (commented):
overall approach lgtm. There are some parts in input_outputs_test.go that will conflict with my #571 — are those actually needed for the refactor?

pkg/infra/iac2/inputs_outputs_test.go
gordon-klotho: Yes because the `Name` field has been removed on some resources that delegate their name to another resource (VpcLink for eg).

pkg/infra/iac2/templates_compiler.go
gordon-klotho: This is to support eg. VpcLink's `Name()` function instead of `.Name` field.

pkg/core/resources.go
```diff
@@ -15,12 +15,16 @@ type (
...
-		Provider() string
 		// KlothoConstructRef returns AnnotationKey of the klotho resource the cloud resource is correlated to
 		KlothoConstructRef() []AnnotationKey
 		// Id returns the id of the cloud resource
-		Id() string
+		Id() ResourceId
+	}
+
+	ResourceId struct {
+		Provider string
```
yuval-klotho, line 25: while we're here, I'd personally like to see this as a new `type ProviderId string`, and have all values pulled out to consts. There have been a couple times I've wondered "what are all the valid provider ids?", and today it's not very easy to find that set.

pkg/infra/iac2/resources.go
jhsinger-klotho: im a little confused why the provider is pulumi for id function of the kubernetes provider here

jhsinger-klotho: oh this is in the pulumi plugin ok

pkg/provider/aws/static_unit_test.go
gordon-klotho: Yeah, because some of the signatures on the graph were changed, I had to change the test. Instead of updating the functions on the dag, I just switched it to use the ResourcesExpectation assert method instead.

pkg/infra/iac2/resources.go
gordon-klotho: I kept them all the same /shrug

PR #579, opened 2023-05-08
start of expansion, just lambda and network

jhsinger-klotho:

start of the construct expansion spike,

each resource now has a create method which is used to create itself based on the metadata passed in.

ResourceGraph has a new method to add all resources that are prereqs

### Standard checks

- **Unit tests**: Any special considerations? everything should be well unit tested in here
- **Docs**: Do we need to update any docs, internal or public? not public facing
- **Backwards compatibility**: Will this break existing apps? If so, what would be the extra work required to keep them working? yeah tried to make sure that the naming was consistent, but we also have not released this version yet


pkg/cli/klothomain.go
```diff
@@ -257,7 +257,7 @@ func (km KlothoMain) run(cmd *cobra.Command, args []string) (err error) {
 			cmd.SilenceUsage = true
 		},
 	}
-	defer analyticsClient.PanicHandler(&err, errHandler)
+	// defer analyticsClient.PanicHandler(&err, errHandler)
```
gordon-klotho, line 260: Don't forget to uncomment before merge

pkg/core/resource_graph.go
```diff
@@ -62,6 +63,18 @@ func (rg *ResourceGraph) GetResource(id ResourceId) Resource {
...
+	vertices := g.underlying.GetAllVertices()
+	for _, v := range vertices {
+		if vT, ok := v.(T); ok {
+			if vT.Id().String() == id {
+				return &vT
+			}
+		}
+	}
+	return nil
+}
```
gordon-klotho, line 76: Isn't this the same as
```go
g.GetResourceByVertexId(id).(*T)
```
?

pkg/core/resources.go
```diff
@@ -19,6 +19,8 @@ type (
 		KlothoConstructRef() []AnnotationKey
 		// Id returns the id of the cloud resource
 		Id() ResourceId
+		// Create modifies itself to be a completed instance of the resource given a set of metadata
+		Create(dag *ResourceGraph, metadata map[string]any) (Resource, error)
```
gordon-klotho, line 23: I like that `Create` mutates (similar to `UnmarshalJSON`) but why does it return a `Resource`?

Also, would we expect every resource to implement this? If not, perhaps this could be a separate interface for resources to implement. Could save on a bunch of "Not Implemented" boilerplate.

This is a bit of a nit pick, but "metadata" doesn't really describe it well IMO - something like "parameters" or "arguments" feel closer to how it's used.

pkg/core/resource_graph_test.go
```diff
@@ -38,6 +40,17 @@ type testResource struct {
...
 }
+func (tr *testResource) Create(dag *ResourceGraph, metadata map[string]any) (Resource, error) {
+	currName := ""
+	if strName, ok := metadata["name"].(string); ok {
+		currName = strName + uuid.NewString()
+	} else {
+		return tr, fmt.Errorf("Failure")
+	}
+
+	tr = &testResource{Name: currName}
```
gordon-klotho, line 51: This should probably be `*tr = testResource` so it mutates `tr` (I believe this as written won't actually change the struct, just what the `tr` variable points to).

pkg/cli/klothomain.go
jhsinger-klotho: ah good catch, let me update

gordon-klotho (commented):
partial review

pkg/core/resource_graph.go
jhsinger-klotho: yeah i removed this

Issue #499, opened 2023-04-13
Visualization Translation


jhsinger-klotho:
Convert Resource DAG to our visualization IR

Definition of Done: 

- On every klotho compilation, we are able to output a diagram as a png

gordon-klotho:
Putting this down for now to pivot onto other things.

🚧 WIP 🚧 
https://github.com/klothoplatform/klotho/tree/ifc2/vis_output (output the topology yaml and call visualizer for image)
+
https://github.com/klothoplatform/visualizer/tree/klotho_integration (change some group directions, converting types, other fixes)

Added `topology.yaml` which is the topology as output from the `vis_output` branch for the `go-expose` sample app.
Currently running into this error on the visualizer:
```
Traceback (most recent call last):
  File "/home/gordon/.local/share/virtualenvs/visualizer-N6ijiUmh/lib/python3.10/site-packages/uvicorn/protocols/http/h11_impl.py", line 429, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
  File "/home/gordon/.local/share/virtualenvs/visualizer-N6ijiUmh/lib/python3.10/site-packages/uvicorn/middleware/proxy_headers.py", line 78, in __call__
    return await self.app(scope, receive, send)
  File "/home/gordon/.local/share/virtualenvs/visualizer-N6ijiUmh/lib/python3.10/site-packages/fastapi/applications.py", line 276, in __call__
    await super().__call__(scope, receive, send)
  File "/home/gordon/.local/share/virtualenvs/visualizer-N6ijiUmh/lib/python3.10/site-packages/starlette/applications.py", line 122, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/home/gordon/.local/share/virtualenvs/visualizer-N6ijiUmh/lib/python3.10/site-packages/starlette/middleware/errors.py", line 184, in __call__
    raise exc
  File "/home/gordon/.local/share/virtualenvs/visualizer-N6ijiUmh/lib/python3.10/site-packages/starlette/middleware/errors.py", line 162, in __call__
    await self.app(scope, receive, _send)
  File "/home/gordon/.local/share/virtualenvs/visualizer-N6ijiUmh/lib/python3.10/site-packages/starlette/middleware/exceptions.py", line 79, in __call__
    raise exc
  File "/home/gordon/.local/share/virtualenvs/visualizer-N6ijiUmh/lib/python3.10/site-packages/starlette/middleware/exceptions.py", line 68, in __call__
[...]

PR #581, opened 2023-05-10
use viz.klo.dev by default

yuval-klotho:
- use the domain instead of the hard-coded gateway url
- tweak the env var name

### Standard checks

- **Unit tests**: none; manually tested
- **Docs**: n/a
- **Backwards compatibility**: no issues (the env var is only an internal/debug thing)


yuval-klotho:
in service of #499

jhsinger-klotho (approved):

PR #584, opened 2023-05-12
fix python multiple imports

### LINKED ISSUE #567: Support module with multiple imports with different alias
Currently, this test case in `imports_test` demonstrates how it doesn't work
```go
{
	name: "import module aliased twice",
	source: `
		import module1 as a
		import module1 as b
`,
	want: map[string]Import{
		"module1": {Name: "module1", Alias: "a"},
	},
},
```

This shows that `FindImports` ignores the alias `b`.

yuval-klotho:
Python lets you import modules multiple times:

```python
import my_module
import my_module as x
import my_module as y
```

Previously, we would drop all but the last of these. Now, we'll capture all of them.

A big structural change in this is to remove the `Alias string` and `ImportedAs() string` in `Import struct`, and similarly for `Attribute`. Instead, there is only a `UsedAs map[string]struct{}`:

- `import my_module`
  ↳ UsedAs = {"my_module"}`
- `import my_module as a`
  ↳ UsedAs = {"a"}
- `import my_module` + `import my_module as a`
  ↳ UsedAs = {"my_module", "a"}

This resolves #567 

### Standard checks

- **Unit tests**: added a bunch, and modified a bunch, to handle multi-imports
- **Docs**: none needed
- **Backwards compatibility**: should be fully backwards compatible

### DIFF
--- a/pkg/lang/python/plugin_persist.go
+++ b/pkg/lang/python/plugin_persist.go
@@ -436,8 +436,8 @@ func (p *persister) queryKV(file *core.SourceFile, annotation *core.Annotation, 
 	}
 	aiocacheImported := len(aiocacheImport.ImportedAttributes) == 0
 	cacheImport, cacheImported := aiocacheImport.ImportedAttributes["Cache"]
-	functionHostName := aiocacheImport.Name
-	cacheFunction := cacheImport.Name
+	//functionHostName := aiocacheImport.Name
+	//cacheFunction := cacheImport.Name
 
 	nextMatch := DoQuery(annotation.Node, persistKV)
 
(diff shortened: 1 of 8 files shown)


pkg/testutil/sitter.go
```diff
@@ -0,0 +1,24 @@
+package testutil
+
+import (
+	sitter "github.com/smacker/go-tree-sitter"
+)
+
+// FindNodeByContent returns a node whose content is the given content string. If there are multiple such nodes, this
+// will return one of them, but it's unspecified which.
+func FindNodeByContent(tree *sitter.Tree, content string) *sitter.Node {
```
gordon-klotho, line 9: nit: if this used node instead of tree for input, the inner function wouldn't be needed

pkg/lang/python/plugin_persist.go
```diff
@@ -515,11 +511,15 @@ func (p *persister) queryFS(file *core.SourceFile, annotation *core.Annotation,
 	module, aliasedModule, alias, importStatement := match["module"], match["aliasedModule"], match["alias"], match["importStatement"]
 
 	// this assignment/invocation is unrelated to aiofile instantiation found from the matching import
+	var varName string
 	if aliasedModule != nil {
-		if !query.NodeContentEquals(alias, varName) {
-			return nil
+		if content, matched := query.GetMatchingNodeContent(alias, varNames); matched {
```
gordon-klotho, line 516: nit: (opinion) I don't feel like this is easier to read than `_, matched := varNames[alias.Content()]` (and using `alias.Content()` inside). Especially not clear that the first argument is just the `.Content()` of the node passed in.

pkg/lang/python/plugin_persist.go
```diff
@@ -496,12 +494,10 @@ func (p *persister) queryFS(file *core.SourceFile, annotation *core.Annotation,
...
-	varName := ""
-	if fsSpecImport.Alias != "" {
-		varName = fsSpecImport.Alias
-	} else if len(fsSpecImport.ImportedAttributes) == 0 {
-		varName = fsSpecImport.Name
-	} else {
+	varNames := fsSpecImport.UsedAs
+	if len(varNames) == 0 {
+		// this means it's an attribute-import: "from aiofiles import f", which we don't support
+		log.Warn("Unsupported import")
```
gordon-klotho, line 500: Good comment - could this go into the warning instead?

gordon-klotho (approved):

pkg/lang/python/plugin_persist.go
yuval-klotho: Is that true of the `query.NodeContentIn` call? If you like that one but not this one, I could also just do:

```go
if query.NodeContentIn(alias, varNames) {
  content := alias.Content() // an extra call to Content(), but that's probably fine
  ...
```

pkg/lang/python/plugin_persist.go
yuval-klotho: I didn't do that because we output the annotation along with the log anyway, so it felt like it'd feel redundant. It would be something like:

```
Unsupported import: from aiofiles import f
| @klotho::persist {
|   id = 'hello'
| }
from aiofiles import f
```

What about something like: `Unsupported import. Please use "import aiofiles" without "from"`?

pkg/testutil/sitter.go
yuval-klotho: I went for `Tree` just because it's a _little_ more obvious (at least to me) that it's a recursive/deep call. That is, finding a Node within a Tree has to be recursive, there's no other choice. Finding a Node from a Node _might_ be "return the given node, or nil`.

But maybe that's a silly enough alternative ("why wouldn't you just return bool?") that people won't assume it, or if they do assume it, they'll at least be 🤨'ed enough that they'll read the docs.

pkg/lang/python/plugin_persist.go
yuval-klotho: ☝️ I went with that

pkg/lang/python/plugin_persist.go
yuval-klotho: ☝️ I went with that

Issue #589, opened 2023-05-15
Support relative python imports


gordon-klotho:
Check & fix relative imports (such as `import .foo` or `from .foo import blah`)

PR #590, opened 2023-05-15
Add sibling module to python import test

gordon-klotho:
No actual changes here, just adding a case to the test. Thought this was broken, but either it wasn't broken, or the changes in https://github.com/klothoplatform/klotho/pull/584 fixed it.

Addresses https://github.com/klothoplatform/klotho/issues/589

### Standard checks

### DIFF
--- a/pkg/lang/python/queries/find_imports.scm
+++ b/pkg/lang/python/queries/find_imports.scm
@@ -1,22 +1,22 @@
 [
-  (import_statement [
-      name: (dotted_name) @module
-      name: (aliased_import
+  (import_statement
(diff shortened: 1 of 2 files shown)


pkg/lang/python/queries/find_imports.scm
yuval-klotho, line 1: Is this file just formatting changes, or anything substantive?

gordon-klotho: Just formatting

yuval-klotho (approved):

PR #593, opened 2023-05-16
Enables ALB Controller readiness gates for exec unit deployments

### LINKED ISSUE #557: Configure k8s deployments + target group health checks to result in 0 downtime
right now there is slight downtime on new deployments if you have 2 replicas (our default) because the k8s control plane cycles pods faster than our target group health checks allow them to be marked as healthy

DavidSeptimus-Klotho:
Resolves #557 

This PR enables the ALB Controller pod readiness gate for pods associated with an execution unit ([docs](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/deploy/pod_readiness_gate/)) by implementing the following:

- Setting the following values on the ALB Controller Chart:
  1. `webhookNamespaceSelector` as an empty `matchExpressions` array to enable readiness gate injection in all namespaces
  2. `objectSelector.matchLabels` to enable pod readiness gate injection for pods labeled with the `elbv2.k8s.aws/pod-readiness-gate-inject: enabled`
- Adding the `elbv2.k8s.aws/pod-readiness-gate-inject: enabled` label to pods or deployment templates associated with target group bindings
- 
### Standard checks

### DIFF
--- a/pkg/infra/kubernetes/manifests/deployment.yaml.tmpl
+++ b/pkg/infra/kubernetes/manifests/deployment.yaml.tmpl
@@ -19,6 +19,7 @@ spec:
     metadata:
       labels:
         execUnit: {{ .ExecUnitName }}
+        elbv2.k8s.aws/pod-readiness-gate-inject: enabled
     spec:
       containers:
       - image: REPLACE_ME
--- a/pkg/infra/kubernetes/manifests/deployment.yaml.tmpl
+++ b/pkg/infra/kubernetes/manifests/deployment.yaml.tmpl
@@ -19,7 +19,6 @@ spec:
     metadata:
       labels:
         execUnit: {{ .ExecUnitName }}
-        elbv2.k8s.aws/pod-readiness-gate-inject: enabled
     spec:
       containers:
       - image: REPLACE_ME
--- a/pkg/provider/aws/resources/eks.go
+++ b/pkg/provider/aws/resources/eks.go
@@ -35,6 +35,7 @@ const (
 	ID_IAC_VALUE                                  = "id"
 	AWS_OBSERVABILITY_CONFIG_MAP_REGION_IAC_VALUE = "aws_observ_cm_region"
 	NODE_GROUP_NAME_IAC_VALUE                     = "node_group_name"
+	ALB_CONTROLLER_READINESS_GATE_IAC_VALUE       = "alb_controller_readiness_gate"
 
 	AWS_OBSERVABILITY_NS_PATH         = "aws_observability_namespace.yaml"
 	AWS_OBSERVABILITY_CONFIG_MAP_PATH = "aws_observability_configmap.yaml"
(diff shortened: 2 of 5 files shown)


pkg/infra/iac2/templates_compiler.go
```diff
@@ -664,6 +664,17 @@ func (tc TemplatesCompiler) handleIaCValue(v core.IaCValue, appliedOutputs *[]Ap
...
+		return `
+(obj: any, opts: pulumi.CustomResourceOptions) => {
+	if (obj.kind === "MutatingWebhookConfiguration") {
+		obj.webhooks.filter(w => w.name === "mpod.elbv2.k8s.aws").forEach((webhook: any) => {
+                webhook.namespaceSelector = undefined
+                webhook.objectSelector.matchLabels = {"elbv2.k8s.aws/pod-readiness-gate-inject": "enabled"}
+            }
+		)
+	}	
+}`, nil
```
jhsinger-klotho, line 677: i dont think this is needed.
look for enablePodReadinessGateInject in
https://github.com/kubernetes-sigs/aws-load-balancer-controller/tree/main/helm/aws-load-balancer-controller

 i think that covers it

pkg/infra/kubernetes/manifests/deployment.yaml.tmpl
```diff
@@ -19,6 +19,7 @@ spec:
     metadata:
       labels:
         execUnit: {{ .ExecUnitName }}
+        elbv2.k8s.aws/pod-readiness-gate-inject: enabled
```
jhsinger-klotho, line 22: we wont want this to always be injected because it really is only necessary for pods which are exposed.

I would make this be written only if the pod is exposed and also set in the aws provider since this is aws specific. If its too much to make this be set in the aws provider its ok to accept that as tech debt for now

gordon-klotho:
I don't have a review to add other than the links to docs are great - can you find a way to include them as godoc/comments in the code?

pkg/infra/iac2/templates_compiler.go
```diff
@@ -664,6 +664,17 @@ func (tc TemplatesCompiler) handleIaCValue(v core.IaCValue, appliedOutputs *[]Ap
 			tc.getVarName(region), tc.getVarName(v.Resource)), nil
 	case resources.NODE_GROUP_NAME_IAC_VALUE:
 		return fmt.Sprintf(`%s.nodeGroupName`, tc.getVarName(resource)), nil
+	case resources.ALB_CONTROLLER_READINESS_GATE_IAC_VALUE:
+		return `
```
yuval-klotho, line 668: would it be helpful to move `testutils.UnIndent` to a prod-scope module, so we could use it here? (Basically just wrap this string with that func, and then you can indent the string and it'll appear un-indented in output.)

Obviously not a blocker.

pkg/infra/iac2/templates_compiler.go
DavidSeptimus-Klotho: Thanks for pointing me in the direction of the ALB Controller chart's values. I was looking at an older version that didn't template the selectors we need to modify for this change.

`enablePodReadinessGateInject` isn't required unless you want to disable pod readiness gate injection.

I've reworked this PR to make use of the chart's `objectSelector.matchLabels` and `webhookNamespaceSelectors` values to apply the same changes originally included in the transformation.

pkg/infra/kubernetes/manifests/deployment.yaml.tmpl
DavidSeptimus-Klotho: I was wondering about this. I can take a closer look.

pkg/infra/iac2/templates_compiler.go
DavidSeptimus-Klotho: I like that idea, but not as part of this change now that I've removed this `IaCValue`.

pkg/provider/aws/resources/eks.go
```diff
@@ -35,6 +35,7 @@ const (
 	ID_IAC_VALUE                                  = "id"
 	AWS_OBSERVABILITY_CONFIG_MAP_REGION_IAC_VALUE = "aws_observ_cm_region"
 	NODE_GROUP_NAME_IAC_VALUE                     = "node_group_name"
+	ALB_CONTROLLER_READINESS_GATE_IAC_VALUE       = "alb_controller_readiness_gate"
```
jhsinger-klotho, line 38: is this still needed?

jhsinger-klotho (approved):

PR #594, opened 2023-05-17
initial knowledge base class and other utility/helper functions

jhsinger-klotho:

This PR Defines all utilities which were used in the edge expansion spike.

1. CreateDependencies: Used to recurively create direct dependencies which exist on an IR. These are dependencies which we know are critical to the functioning of a resource and are not optional
2. CreateResource: A wrapper to add safety around calling a Resource's Create method. The safety is in regards to making sure the resource returned is truly what exists in the graph since the .create method takes in a pointer
3. The Knowledge Base

The knowledge base should have go docs documenting whats going on so going to default to that rather than rewriting

### Standard checks

- **Unit tests**: Any special considerations? tests should be in place
- **Backwards compatibility**: Will this break existing apps? If so, what would be the extra work required to keep them working? not used anywhere yet


pkg/core/edge_kb.go
```diff
@@ -0,0 +1,296 @@
...
+
+	"github.com/klothoplatform/klotho/pkg/multierr"
+	"go.uber.org/zap"
+)
+
+type (
+	// Edge defines an entry in a Knowledge base. An Edge represents a valid linking between two types of resources
+	Edge struct {
+		// From represents the source resource in the edge
+		From reflect.Type
```
DavidSeptimus-Klotho, line 15: We might want to rename `From` and `To` to `Source` and `Destination` (or `Target`) to be consistent with how we already refer to either end of an edge.

pkg/core/edge_kb.go
```diff
@@ -0,0 +1,296 @@
...
+		Constraint EdgeConstraint
+		// Source refers to the initial source node when edge expansion is called
+		Source Resource
+		// Destination refers to the initial target node when edge expansion is called
+		Destination Resource
+	}
+)
+
+// GetEdgeDetails takes in a source and target to retrieve the edge details for the given key. Will return nil if no edge exists for the given source and target
+func GetEdgeDetails(kb EdgeKB, source reflect.Type, target reflect.Type) EdgeDetails {
```
DavidSeptimus-Klotho, line 62: Is there a reason this function and the related functions below don't have an `EdgeKB` receiver?

pkg/core/resource_graph.go
```diff
@@ -235,3 +259,214 @@ func (rg *ResourceGraph) getAllDownstreamResourcesSet(source Resource, upstreams
...
+	rg.AddDependenciesReflect(res)
+	return merr.ErrOrNil()
+}
+
+// CreateResource is a wrapper around a Resources .Create method
+//
+// CreateResource provides safety in assuring that the up to date resource (which is inline with what exists in the ResourceGraph) is returned
+func CreateResource[T Resource](rg *ResourceGraph, params any) (resource T, err error) {
+
+	res := reflect.New(reflect.TypeOf(resource).Elem()).Interface()
```
DavidSeptimus-Klotho, line 319: Do you need to use reflection to create an instance of T here? Isn't `resource` already a zero value of type `T`?

pkg/core/annotation.go
```diff
@@ -49,6 +49,22 @@ func (a *Annotation) MarshalJSON() ([]byte, error) {
 	return json.Marshal(m)
 }
 
+func DedupeAnnotationKeys(a []AnnotationKey) []AnnotationKey {
```
atorres-klo, line 52: you can speed this up a fair amount if the list gets large with a map:

`func DedupeAnnotationKeys(a []AnnotationKey) []AnnotationKey {
	result := make(map[AnnotationKey]struct{})
	for _, key := range a {
		result[key] = struct{}{}
	}
	keys := make([]AnnotationKey, 0, len(result))
	for key := range result {
		keys = append(keys, key)
	}
	return keys
}`

pkg/core/edge_kb.go
```diff
@@ -0,0 +1,296 @@
...
+	delete(visited, source)
+}
+
+// isValidForPath determines if an edge is valid for an instance of path generation.
+//
+// The criteria is:
+//   - if there is no expansion function and no ValidDestinations, assume all destinations are valid.
+//   - otherwise check to see if the path generations destination is valid for the edge
+func isValidForPath(kb EdgeKB, edge Edge, dest reflect.Type) bool {
+	edgeDetail := kb[edge]
```
atorres-klo, line 132: is it possible for edgeDetail to be nil here if edge isn't in kb?

pkg/core/resource_graph.go
```diff
@@ -47,13 +50,30 @@ func (rg *ResourceGraph) AddResource(resource Resource) {
 //
 //	rg.AddDependency(lambda, lambda.Role)
 func (rg *ResourceGraph) AddDependency(deployedSecond Resource, deployedFirst Resource) {
+
 	for _, res := range []Resource{deployedSecond, deployedFirst} {
```
atorres-klo, line 54: Can't we just called rg.AddResource twice here? Are we thinking there will be more resources down the road?

pkg/core/resource_graph.go
```diff
@@ -235,3 +259,214 @@ func (rg *ResourceGraph) getAllDownstreamResourcesSet(source Resource, upstreams
...
+				return fmt.Errorf("return type should be an error")
+			}
+			return err
+		}
+	}
+	return nil
+}
+
+func (rg *ResourceGraph) checkChild(child reflect.Value, res Resource, metadata any) error {
+	var merr multierr.Error
```
atorres-klo, line 432: Are we on go 1.20? Maybe not worth it now, but https://go.dev/doc/go1.20#errors may make more sense than multierrors when we wanna convert

pkg/core/resource_graph_test.go
```diff
@@ -121,3 +133,127 @@ func TestResourceGraph_AddDependenciesReflect(t *testing.T) {
 	}
 	assert.Nil(dag.GetDependencyByVertexIds(tr.Id().String(), (&testResource{Name: "nested_single"}).Id().String()))
 }
+
+func TestResourceGraph_CreateDependencies(t *testing.T) {
```
atorres-klo, line 137: Thank you for tests on this one

atorres-klo (commented):
Just a few comments, nothing is blocking

pkg/core/edge_kb_test.go
```diff
@@ -0,0 +1,232 @@
...
+	D struct {
+		Name string
+	}
+	E struct {
+		Name string
+	}
+)
+
+func (f *A) Id() ResourceId                      { return ResourceId{Name: "A" + f.Name} }
+func (f *A) Provider() string                    { return "DummyProvider" }
```
gordon-klotho, line 215: This isn't on the `Resource` interface anymore (should be in the `Id` function instead)

pkg/core/resource_graph.go
gordon-klotho: I originally wrote this - the adding used to be more complex. The "if not present" check used to be here, which is why I did the for loop. Once I moved that check into `AddResource`, I didn't bother to change this from a loop.

It should be two `AddResource` calls, but it's not anything Jordan changed.

pkg/core/resource_graph.go
gordon-klotho: The zero value would be `nil`, but we can add another type parameter to make it work simpler:

```go
func CreateResource[T interface{*IT; Resource}, IT any](rg *ResourceGraph, params any) (resource T, err error) {

	res := new(IT)
```
This says that "T is a Resource and a pointer to IT", then you can use `new(IT)` to make a new, non-nil T (that is, a `*IT`)

gordon-klotho (approved):

pkg/core/edge_kb.go
jhsinger-klotho: nope thats something i can change and probably is a good idea as we expand what our knowledge base is

pkg/core/resource_graph.go
jhsinger-klotho: oh i can try that out. Yeah to gordons point it is nil, but if the other method works then ill go with that

pkg/core/annotation.go
jhsinger-klotho: makes sense, can do that, just since we really only ever have like 1 or 2 in the list that we are watching for dedupes i wasnt too worried about speed, but might as well optimize

pkg/core/resource_graph.go
jhsinger-klotho: yeah i can go look at the code since i didnt change this method i need to take a peek, but if theres something we know needs changing, gordon maybe we can sync about it

pkg/core/resource_graph.go
jhsinger-klotho: i believe we are now, can move to the go errors

pkg/core/edge_kb.go
```diff
@@ -0,0 +1,296 @@
...
+		Configure EdgeConfigurer
+		// ValidDestinations is a list of end destinations the edge supports. This field is used within determining the path (edge expansion) between two resources.
+		ValidDestinations []reflect.Type
+	}
+
+	// EdgeKB is a map (knowledge base) of edges and their respective details used to configure ResourceGraphs
+	EdgeKB map[Edge]EdgeDetails
+
+	// EdgeExpander is a function used to create the To and From resource and necessary intermediate resources, if any do not exist, to ensure the nodes are in place for correct functionality.
+	EdgeExpander func(from, to Resource, dag *ResourceGraph, data EdgeData) error
```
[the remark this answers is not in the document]
jhsinger-klotho: 👍

pkg/core/edge_kb.go
```diff
@@ -0,0 +1,296 @@
...
+	// EdgeData is an object attached to edges in the ResourceGraph to help the knowledge base understand context when performing expansion and configuration tasks
+	EdgeData struct {
+		// AppName refers to the application name of the global ResourceGraph
+		AppName string
+		// EnvironmentVaribles specify and environment variables which will need to be configured during the edge expansion process
+		EnvironmentVariables []EnvironmentVariable
+		// Constraint refers to the EdgeConstraints defined during the edge expansion
+		Constraint EdgeConstraint
+		// Source refers to the initial source node when edge expansion is called
+		Source Resource
```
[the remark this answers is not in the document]
jhsinger-klotho: directly to a resource actually, im trying to avoid ever referencing constructs in the engine, so that its a purely compiler concept. By doing so we will need to translate constructs to resources + constraints in the compiler and not in the engine (long term view) but it allows us to be able to do more in terms of how we can make determinations. Well need generics to fully get there though but i think thats known.

pkg/core/edge_kb.go
```diff
@@ -0,0 +1,296 @@
...
+
+// FindPaths takes in a source and destination type and finds all valid paths to get from source to destination.
+//
+// Find paths does a Depth First Search to search through all edges in the knowledge base.
+// The function tracks visited edges to prevent cycles during execution
+// It also checks the ValidDestinations for each edge against the original destination node to ensure that the edge is allowed to be used in the instance of the path generation
+//
+// The method will return all paths found
+func FindPaths(kb EdgeKB, source reflect.Type, dest reflect.Type) [][]Edge {
+	zap.S().Debugf("Finding Paths from %s -> %s", source.String(), dest.String())
```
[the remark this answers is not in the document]
jhsinger-klotho: 👍

pkg/core/annotation.go
gordon-klotho: I mention this somewhere else I think, but might be worth making the annotation keys a set, that way we don't have to remember when to dedupe or not.

pkg/core/resource_graph.go
gordon-klotho: Yeah, I think we can but I'd prefer we file that as tech debt and take it on later (or at least, deliberately and separate from this work)

pkg/core/edge_kb.go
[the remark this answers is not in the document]
gordon-klotho: I meant the resources created at the first step of the engine, the construct -> resource step. I suppose this is what you might have meant by "initial" but I was sure if that was an established term.

pkg/core/resource_graph.go
jhsinger-klotho: by going out this route i see `invalid operation: res (variable of type *IT) is not an interface` when trying to typecast res to a Resource. Going to leave for now to continue addressing comments but we can go back and change since it doesnt have any downstream affect

pkg/core/edge_kb.go
jhsinger-klotho: yeah also the only thing being put into here are valid edges since its used by findPaths so we should be ok

pkg/core/resource_graph.go
jhsinger-klotho: i cut a ticket for it so we dont forget, but going to address everything to unblock all the next tasks first

pkg/core/edge_kb.go
```diff
@@ -0,0 +1,296 @@
+package core
```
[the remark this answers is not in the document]
jhsinger-klotho: ok i can move it out into its own package, i think thats fine

pkg/core/resource_graph.go
gordon-klotho: You wouldn't need to typecast it, since the generic already specifies that it is a Resource via `T`. Might have to do `res := T(new(It))` to make it a T explicitly.

pkg/core/edge_kb.go
```diff
@@ -0,0 +1,296 @@
...
+	visitedEdges := map[reflect.Type]bool{}
+	stack := []Edge{}
+	findPaths(kb, source, dest, stack, visitedEdges, &result)
+	return result
+}
+
+// findPaths performs the recursive calls of the parent FindPath function
+//
+// It works under the assumption that an edge is bidirectional and uses the edges ValidDestinations field to determine when that assumption is incorrect
+func findPaths(kb EdgeKB, source reflect.Type, dest reflect.Type, stack []Edge, visited map[reflect.Type]bool, result *[][]Edge) {
```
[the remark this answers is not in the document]
jhsinger-klotho: sure, changing it

pkg/core/edge_kb.go
```diff
@@ -0,0 +1,296 @@
...
+				if len(p) == 2 {
+					zap.S().Debug("Defaulting to direct path")
+					validPath = p
+				}
+			}
+			if len(validPath) == 0 {
+				merr.Append(fmt.Errorf("found multiple paths which satisfy constraints for edge %s -> %s. \n Paths: %s", dep.Source.Id().String(), dep.Destination.Id().String(), validPaths))
+			}
+		} else {
+			validPath = validPaths[0]
```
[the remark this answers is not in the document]
jhsinger-klotho: good catch adding error handling for it

pkg/core/edge_kb.go
```diff
@@ -0,0 +1,296 @@
...
+				}
+			}
+
+			// resourceCache is used to always pass the graphs nodes into the Expand functions if they exist. We do this so that we operate on nodes which already exist
+			resourceCache := map[reflect.Type]Resource{}
+			for _, edge := range validPath {
+				from := edge.From
+				to := edge.To
+				edgeDetail := GetEdgeDetails(kb, from, to)
+				fromNode := reflect.New(from.Elem()).Interface().(Resource)
```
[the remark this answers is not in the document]
jhsinger-klotho: flipped

pkg/core/edge_kb_test.go
jhsinger-klotho: oh nice, i copy pasted from other tests so didnt remove, but removing now

PR #595, opened 2023-05-17
Eu lambda expansion

jhsinger-klotho:

This PR shows how the createDependencies function from the [helpers pr](https://github.com/klothoplatform/klotho/pull/594) is used to create a lambda and its dependent resources.

This pr shows how by just calling the lambda.Create method we would be able to achieve a functional setup for that execution unit running on lambda in base node creation, without any configuration of the nodes

### Standard checks

### DIFF
--- a/pkg/provider/aws/execution_unit.go
+++ b/pkg/provider/aws/execution_unit.go
@@ -12,6 +12,25 @@ import (
 	"go.uber.org/zap"
 )
 
+// expandExecutionUnit takes in a single execution unit and expands the generic construct into a set of resource's based on the units configuration.
+func (a *AWS) expandExecutionUnit(dag *core.ResourceGraph, unit *core.ExecutionUnit) error {
+	switch a.Config.GetExecutionUnit(unit.ID).Type {
+	case Lambda:
+		lambda, err := core.CreateResource[*resources.LambdaFunction](dag, resources.LambdaCreateParams{
+			AppName: a.Config.AppName,
+			Refs:    []core.AnnotationKey{unit.AnnotationKey},
+			Name:    unit.ID,
+		})
+		a.MapResourceDirectlyToConstruct(lambda, unit)
+		if err != nil {
+			return err
+		}
+	default:
+		return fmt.Errorf("unsupported execution unit type %s", a.Config.GetExecutionUnit(unit.ID).Type)
+	}
+	return nil
+}
+
 // GenerateExecUnitResources generates the necessary AWS resources for a given execution unit and adds them to the resource graph
 func (a *AWS) GenerateExecUnitResources(unit *core.ExecutionUnit, result *core.ConstructGraph, dag *core.ResourceGraph) error {
 	log := zap.S()
--- a/pkg/provider/aws/resources/cloudwatch.go
+++ b/pkg/provider/aws/resources/cloudwatch.go
@@ -20,6 +20,27 @@ type (
 	}
 )
 
+type CloudwatchLogGroupCreateParams struct {
+	AppName string
+	Refs    []core.AnnotationKey
+	Name    string
+}
+
+// Create takes in an all necessary parameters to generate the LogGroup name and ensure that the LogGroup is correlated to the constructs which required its creation.
+func (logGroup *LogGroup) Create(dag *core.ResourceGraph, params CloudwatchLogGroupCreateParams) error {
+	logGroup.Name = logGroupSanitizer.Apply(fmt.Sprintf("%s-%s", params.AppName, params.Name))
+	logGroup.ConstructsRef = params.Refs
+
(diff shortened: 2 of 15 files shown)


pkg/provider/aws/execution_unit.go
```diff
@@ -12,6 +12,25 @@ import (
 	"go.uber.org/zap"
 )
 
+// expandExecutionUnit takes in a single execution unit and expands the generic construct into a set of resource's based on the units configuration.
+func (a *AWS) expandExecutionUnit(dag *core.ResourceGraph, unit *core.ExecutionUnit) error {
+	switch a.Config.GetExecutionUnit(unit.ID).Type {
+	case Lambda:
+		lambda, err := core.CreateResource[*resources.LambdaFunction](dag, resources.LambdaCreateParams{
+			AppName: a.Config.AppName,
+			Refs:    []core.AnnotationKey{unit.AnnotationKey},
```
gordon-klotho, line 21: Won't lambda always have exactly 1 Ref?

pkg/provider/aws/resources/cloudwatch.go
```diff
@@ -20,6 +20,27 @@ type (
...
+}
+
+// Create takes in an all necessary parameters to generate the LogGroup name and ensure that the LogGroup is correlated to the constructs which required its creation.
+func (logGroup *LogGroup) Create(dag *core.ResourceGraph, params CloudwatchLogGroupCreateParams) error {
+	logGroup.Name = logGroupSanitizer.Apply(fmt.Sprintf("%s-%s", params.AppName, params.Name))
+	logGroup.ConstructsRef = params.Refs
+
+	existingLogGroup := dag.GetResource(logGroup.Id())
+	if existingLogGroup != nil {
+		graphLogGroup := existingLogGroup.(*LogGroup)
```
gordon-klotho, line 36: nit: I think there's a generic form of `GetResource` that will hide this casting.

pkg/provider/aws/resources/iam.go
```diff
@@ -147,6 +147,30 @@ type (
...
+func (role *IamRole) Create(dag *core.ResourceGraph, params RoleCreateParams) error {
+	role.Name = roleSanitizer.Apply(fmt.Sprintf("%s-%s", params.AppName, params.Name))
+	role.ConstructsRef = params.Refs
+
+	existingRole := dag.GetResource(role.Id())
+	if existingRole != nil {
+		return fmt.Errorf("iam role with name %s already exists", role.Name)
+	}
+
+	dag.AddResource(role)
```
gordon-klotho, line 166: This pattern happens a lot, maybe add a method on the dag to reduce copy/paste?

pkg/provider/aws/resources/lambda_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/klothoplatform/klotho/pkg/core/coretesting"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_LambdaCreate(t *testing.T) {
+	eu := &core.ExecutionUnit{AnnotationKey: core.AnnotationKey{ID: "test", Capability: annotation.ExecutionUnitCapability}, DockerfilePath: "path"}
+	cases := []struct {
+		name    string
+		lambda  *LambdaFunction
+		vpc     bool
```
gordon-klotho, line 17: This appears to be unused in the test. I know we'll probably have this eventually, but the case struct should match with what is actually tested (so we don't forget to implement it).

gordon-klotho (approved):

pkg/provider/aws/execution_unit.go
jhsinger-klotho: yeah it should right now, but maybe not always if we allow people to combine units or something. Also i tried to keep the create params consistent if possible

pkg/provider/aws/resources/cloudwatch.go
jhsinger-klotho: hmmmm i didnt think there was, but ill take another look

pkg/provider/aws/resources/lambda_test.go
jhsinger-klotho: oh the vpc is from leftover when create wasnt just responsible for the name, good catch i will remove

pkg/provider/aws/resources/iam.go
jhsinger-klotho: yeah going to add this to the same ticket

pkg/provider/aws/resources/cloudwatch.go
DavidSeptimus-Klotho: Every time I've written it, I've ended up removing it before putting up a PR because I ended up not using it. At least the groundwork of renaming the old generic `GetResource` that returned a construct made it in at some point.

PR #596, opened 2023-05-17
orm rds expansion

jhsinger-klotho:

follows up to (lambda expansion pr)[https://github.com/klothoplatform/klotho/pull/595] and adds the expansion for networking resources and rds instance + subnet group.

Does the orm construct expansionto generate rds instance and necessary resources
### Standard checks


pkg/provider/aws/resources/region.go
```diff
@@ -43,7 +43,7 @@ func (region *Region) Id() core.ResourceId {
 	return core.ResourceId{
 		Provider: AWS_PROVIDER,
 		Type:     REGION_TYPE,
-		Name:     region.Name,
+		Name:     "region",
```
atorres-klo, line 46: Do we want to make these const somewhere?

atorres-klo (commented):
minor comment

pkg/infra/iac2/templates_compiler_test.go
```diff
@@ -322,7 +322,7 @@ func Test_handleIaCValue(t *testing.T) {
 			resourceVarNamesById: map[core.ResourceId]string{
 				{Provider: "aws", Type: "availability_zones", Name: "AvailabilityZones"}: "azs",
 			},
-			want: "availabilityZones.names[2]",
+			want: "azs.names[2]",
```
gordon-klotho, line 325: What changed to require this to change? I don't see any changes to templates_compiler so I'm not sure.

pkg/provider/aws/orm_test.go
```diff
@@ -1,14 +1,102 @@
...
+			aws := AWS{
+				Config: tt.config,
+			}
+			err := aws.expandOrm(dag, tt.orm)
+
+			if !assert.NoError(err) {
+				return
+			}
+			tt.want.Assert(t, dag)
+			fmt.Println(coretesting.ResoucesFromDAG(dag).GoString())
```
gordon-klotho, line 93: nit: leftover debugging

pkg/provider/aws/resources/ecr_test.go
```diff
@@ -119,6 +118,8 @@ func Test_ImageCreate(t *testing.T) {
 				return
 			}
 			tt.want.Assert(t, dag)
+			graphImage := dag.GetResourceByVertexId(image.Id().String())
```
gordon-klotho, line 121: nit: inconsistent where to use `ByVertexId` and when to use the regular one (taking in the `ResourceId`)

pkg/provider/aws/resources/rds.go
```diff
@@ -107,6 +107,85 @@ type (
 	}
 )
 
+type RdsInstanceCreateParams struct {
+	AppName string
+	Refs    []core.AnnotationKey
+	Name    string
+}
+
+// Create takes in an all necessary parameters to generate the RdsInstance name and ensure that the RdsInstance is correlated to the constructs which required its creation.
```
gordon-klotho, line 116: I'm not a huge fan of the copy/paste preamble. Feels like it should go somewhere common (eg a package-level doc or an interface).

pkg/provider/aws/resources/rds.go
```diff
@@ -107,6 +107,85 @@ type (
...
+//
+// This method will also create dependent resources which are necessary for functionality. Those resources are:
```
gordon-klotho, line 118: I'm also not sure the maintenance overhead of listing these resources is worth it. It's relatively trivial to look through the function to see what resources it creates. Especially when this is listing transitive dependencies which may change underneath it.

Might be valuable for complex "<this> set is created based on <x> or <that> set is created based on <y>" but IMO not in these cases.

pkg/provider/aws/resources/vpc.go
```diff
@@ -88,6 +88,230 @@ type (
 	}
 )
 
+type VpcCreateParams struct {
+	AppName string
+	Refs    []core.AnnotationKey
+}
+
+// Create takes in an all necessary parameters to generate the Vpc name and ensure that the Vpc is correlated to the constructs which required its creation.
+func (vpc *Vpc) Create(dag *core.ResourceGraph, params VpcCreateParams) error {
```
gordon-klotho, line 97: Where does import fit in this new structure? Or are we regressing on it for now.

pkg/provider/aws/resources/vpc.go
```diff
@@ -88,6 +88,230 @@ type (
...
+			Refs:    params.Refs,
+		}
+		err := igw.Create(dag, igwParams)
+		if err != nil {
+			return err
+		}
+		dag.AddDependency(rt, igw)
+	}
+
+	err = dag.CreateDependencies(subnet, map[string]any{
```
gordon-klotho, line 262: Shouldn't the VPC be created before all the gateways, routetables, etc? If frontloaded, could also save some creation time for cache hits.

gordon-klotho (approved):
Not sure if this is a blocker or not (ie, not this change to the branch, but from the branch to main), but this would regress on importing VPC & network resources.

jhsinger-klotho:
> Not sure if this is a blocker or not (ie, not this change to the branch, but from the branch to main), but this would regress on importing VPC & network resources.

Yeah i cut a card specifically for working back in imports, i didnt want to block on that just because it seems like something we can do in parallel to developing the remaining features but i dont think it should be too hard to work back in since what youve done is  ~95% of the work still in this new way

pkg/provider/aws/resources/vpc.go
jhsinger-klotho: the vpc wouldnt get created explicitly, but we should be adding caching to the createDependencies calls to speed things up, just figured we could add that in later

pkg/provider/aws/resources/vpc.go
jhsinger-klotho: regressed for now, but it should fit right in where we check if one is existing we can also ensure if theres an import we have that mapped

pkg/provider/aws/resources/rds.go
jhsinger-klotho: fine by me

pkg/provider/aws/resources/region.go
jhsinger-klotho: yeah i can make these constant, it will be easier to reference throughout

pkg/infra/iac2/templates_compiler_test.go
jhsinger-klotho: so it should always pull the string name in the map that corresponds to the resource id. To be honest, im not sure how this test was ever passing. If you look at the remainder of the tests they follow the same convention. So i dont have a good answer as to why the old way was working

pkg/provider/aws/resources/rds.go
jhsinger-klotho: ok i can delete these for now

PR #597, opened 2023-05-17
Copy edges

jhsinger-klotho:

this pr follows the [orm expansion pr](https://github.com/klothoplatform/klotho/pull/596).

In this pr we copy over edges between constructs into the resource graph as well as their respective edge data used for later processing

### Standard checks

### DIFF
--- a/pkg/provider/aws/plugin.go
+++ b/pkg/provider/aws/plugin.go
@@ -11,6 +11,62 @@ import (
 	"go.uber.org/zap"
 )
 
+// ExpandConstructs looks at all existing constructs in the construct graph and turns them into their respective AWS Resources
+func (a *AWS) ExpandConstructs(result *core.ConstructGraph, dag *core.ResourceGraph) (err error) {
+	log := zap.S()
+	constructIds, err := result.TopologicalSort()
+	if err != nil {
+		return
+	}
+	// We want to reverse the list so that we start at the leaf nodes. This allows us to check downstream dependencies each time and process them.
+	reverseInPlace(constructIds)
+	var merr multierr.Error
+	for _, id := range constructIds {
+		construct := result.GetConstruct(id)
+		log.Debugf("Converting construct with id, %s, to aws resources", construct.Id())
+		switch construct := construct.(type) {
+		case *core.ExecutionUnit:
+			merr.Append(a.expandExecutionUnit(dag, construct))
+		case *core.Orm:
+			merr.Append(a.expandOrm(dag, construct))
+		}
+	}
+	return merr.ErrOrNil()
+}
+
+func (a *AWS) CopyConstructEdgesToDag(result *core.ConstructGraph, dag *core.ResourceGraph) (err error) {
+	var merr multierr.Error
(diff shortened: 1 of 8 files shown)


pkg/provider/aws/plugin.go
```diff
@@ -11,6 +11,62 @@ import (
 	"go.uber.org/zap"
 )
 
+// ExpandConstructs looks at all existing constructs in the construct graph and turns them into their respective AWS Resources
+func (a *AWS) ExpandConstructs(result *core.ConstructGraph, dag *core.ResourceGraph) (err error) {
+	log := zap.S()
+	constructIds, err := result.TopologicalSort()
```
gordon-klotho, line 17: Can you topological sort on a graph that can contain cycles? Does order matter for the expanding?

pkg/provider/aws/plugin.go
```diff
@@ -11,6 +11,62 @@ import (
...
+	if err != nil {
+		return
+	}
+	// We want to reverse the list so that we start at the leaf nodes. This allows us to check downstream dependencies each time and process them.
```
gordon-klotho, line 21: Is this true? I thought this stage doesn't do any cross construct dependencies (as that's done in the `CopyConstructEdgesToDag` step)

pkg/provider/aws/plugin_test.go
```diff
@@ -4,14 +4,197 @@ import (
...
+			aws := AWS{
+				Config: tt.config,
+			}
+			err := aws.ExpandConstructs(result, dag)
+
+			if !assert.NoError(err) {
+				return
+			}
+			tt.want.Assert(t, dag)
+			fmt.Println(coretesting.ResoucesFromDAG(dag).GoString())
```
gordon-klotho, line 118: nit: leftover debug

pkg/provider/aws/provider.go
```diff
@@ -47,3 +49,19 @@ func (a *AWS) GetResourcesDirectlyTiedToConstruct(construct core.Construct) ([]c
 	resources, found := a.constructIdToResources[construct.Id()]
 	return resources, found
 }
+
+func (a *AWS) MapResourceToConstruct(resource core.Resource, construct core.Construct) error {
+	if a.constructIdToResource == nil {
+		a.constructIdToResource = make(map[string]core.Resource)
+	}
+	if res, ok := a.constructIdToResource[construct.Id()]; ok {
```
gordon-klotho, line 57: Why is this an error instead of just deduping it (silently)?

gordon-klotho (approved):

pkg/provider/aws/plugin.go
jhsinger-klotho: you cant sort on a graph that contains cycles, it will error out, so i will need to remove this. It shouldnt matter since the expansion should be independent of eachother and if they collide in resources, the resources Create method should handle that

pkg/provider/aws/plugin.go
jhsinger-klotho: yeah removed, good catch

PR #598, opened 2023-05-17
add necessary resource create functions for rds proxy expansion

jhsinger-klotho:

This pr adds more resources Create methods in preparation for edge expansion pr. This pr is a follow up to [copy edges pr](https://github.com/klothoplatform/klotho/pull/597)

### Standard checks

### DIFF
--- a/pkg/provider/aws/resources/rds.go
+++ b/pkg/provider/aws/resources/rds.go
@@ -186,6 +186,78 @@ func (subnetGroup *RdsSubnetGroup) Create(dag *core.ResourceGraph, params RdsSub
 	return nil
 }
 
+type RdsProxyCreateParams struct {
+	AppName string
+	Name    string
+	Refs    []core.AnnotationKey
+}
+
+// Create takes in an all necessary parameters to generate the RdsProxy name and ensure that the RdsProxy is correlated to the constructs which required its creation.
+//
+// This method will also create dependent resources which are necessary for functionality. Those resources are:
+//   - Subnets
+//   - SecurityGroups
+func (proxy *RdsProxy) Create(dag *core.ResourceGraph, params RdsProxyCreateParams) error {
+	proxy.Name = rdsSubnetSanitizer.Apply(fmt.Sprintf("%s-%s", params.AppName, params.Name))
+	proxy.ConstructsRef = params.Refs
+
+	existingProxy := dag.GetResource(proxy.Id())
+	if existingProxy != nil {
+		graphProxy := existingProxy.(*RdsProxy)
+		graphProxy.ConstructsRef = core.DedupeAnnotationKeys(append(graphProxy.KlothoConstructRef(), params.Refs...))
+	} else {
+		proxy.Subnets = make([]*Subnet, 2)
+		proxy.SecurityGroups = make([]*SecurityGroup, 1)
+		err := dag.CreateDependencies(proxy, map[string]any{
+			"SecurityGroups": []SecurityGroupCreateParams{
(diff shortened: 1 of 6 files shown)


pkg/provider/aws/resources/rds.go
```diff
@@ -186,6 +186,78 @@ func (subnetGroup *RdsSubnetGroup) Create(dag *core.ResourceGraph, params RdsSub
 	return nil
 }
 
+type RdsProxyCreateParams struct {
```
atorres-klo, line 189: Should this stuff live in rds_proxy.go for better separation (I know it's a bit of an unusual case)

pkg/provider/aws/resources/rds.go
```diff
@@ -186,6 +186,78 @@ func (subnetGroup *RdsSubnetGroup) Create(dag *core.ResourceGraph, params RdsSub
...
+//   - SecurityGroups
+func (proxy *RdsProxy) Create(dag *core.ResourceGraph, params RdsProxyCreateParams) error {
+	proxy.Name = rdsSubnetSanitizer.Apply(fmt.Sprintf("%s-%s", params.AppName, params.Name))
+	proxy.ConstructsRef = params.Refs
+
+	existingProxy := dag.GetResource(proxy.Id())
+	if existingProxy != nil {
+		graphProxy := existingProxy.(*RdsProxy)
+		graphProxy.ConstructsRef = core.DedupeAnnotationKeys(append(graphProxy.KlothoConstructRef(), params.Refs...))
+	} else {
```
gordon-klotho, line 208: nit: reduce rightward drift by returning in the if block.

pkg/provider/aws/resources/rds.go
```diff
@@ -186,6 +186,78 @@ func (subnetGroup *RdsSubnetGroup) Create(dag *core.ResourceGraph, params RdsSub
...
+		proxy.Subnets = make([]*Subnet, 2)
```
gordon-klotho, line 209: This is odd - can't this just be determined from the length of the `Subnets` params?

pkg/provider/aws/resources/secret_test.go
```diff
@@ -0,0 +1,130 @@
...
+			if tt.wantErr {
+				assert.Error(err)
+				return
+			}
+			if !assert.NoError(err) {
+				return
+			}
+			tt.want.Assert(t, dag)
+
+			fmt.Println(coretesting.ResoucesFromDAG(dag).GoString())
```
gordon-klotho, line 122: nit: leftover debug

gordon-klotho (approved):

pkg/provider/aws/resources/rds.go
jhsinger-klotho: we could do that and also could split out packages in the future by service. For now im going to defer just for times sake, but i do agree we will need to figure out some sort of organizational packaging as we expand our services and what we support

pkg/provider/aws/resources/rds.go
gordon-klotho: I meant these:
```go
			"Subnets": []SubnetCreateParams{
				{
					AppName: params.AppName,
					Refs:    params.Refs,
					AZ:      "0",
					Type:    PrivateSubnet,
				},
				{
					AppName: params.AppName,
					Refs:    params.Refs,
					AZ:      "1",
					Type:    PrivateSubnet,
				},
			},
```

I was just wondering why create the array before calling the `CreateDependencies`

PR #599, opened 2023-05-17
lambda to rds expansion

jhsinger-klotho:

follows the [prereq pr](https://github.com/klothoplatform/klotho/pull/598)

This is edge definitions and a test showing how the expansion works

### Standard checks

### DIFF
--- a/pkg/provider/aws/edges/edges.go
+++ b/pkg/provider/aws/edges/edges.go
@@ -1,0 +1,117 @@
+package edges
+
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/klothoplatform/klotho/pkg/core"
+	"github.com/klothoplatform/klotho/pkg/provider/aws/resources"
+)
+
+var KnowledgeBase = core.EdgeKB{
+	core.Edge{From: reflect.TypeOf(&resources.LambdaFunction{}), To: reflect.TypeOf(&resources.RdsInstance{})}: core.EdgeDetails{},
+	core.Edge{From: reflect.TypeOf(&resources.LambdaFunction{}), To: reflect.TypeOf(&resources.RdsProxy{})}:    core.EdgeDetails{},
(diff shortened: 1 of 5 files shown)


pkg/provider/aws/edges/edges.go
```diff
@@ -0,0 +1,117 @@
...
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/klothoplatform/klotho/pkg/core"
+	"github.com/klothoplatform/klotho/pkg/provider/aws/resources"
+)
+
+var KnowledgeBase = core.EdgeKB{
+	core.Edge{From: reflect.TypeOf(&resources.LambdaFunction{}), To: reflect.TypeOf(&resources.RdsInstance{})}: core.EdgeDetails{},
```
DavidSeptimus-Klotho, line 12: Given the number of these we need to write, a helper like this might be useful:

```go
func typeOf[T core.Resource]() reflect.Type {
    var t T  
    return reflect.ValueOf(t).Type()
}

// e.g.
typeOf[*core.RdsInstance]()
```

gordon-klotho: Good idea - could go a step further too:
```go
func NewEdge[Src Resource, Dest Resource]() Edge {
    return Edge{From: typeOf[Src](), To: typeOf[Dst]()}
}
```

pkg/provider/aws/edges/edges.go
```diff
@@ -0,0 +1,117 @@
+package edges
```
gordon-klotho, line 1: To go along with a previous PR's comment regarding a `pkg/knowledgebase` package, I'd prefer this in an `aws/knowledgebase` package. We have a lot of different edges so we should be wary of potentially ambiguous names especially when it comes to packages (compared to types, which are always package-prefixed).

gordon-klotho (approved):

pkg/provider/aws/edges/edges.go
jhsinger-klotho: changed the package name

PR #601, opened 2023-05-18
Configure edges

jhsinger-klotho:

follows the [node configuration pr](https://github.com/klothoplatform/klotho/pull/600)

This pr implements the configure method for edges where necessary and shows how they are used via unit tests
### Standard checks


pkg/provider/aws/edges/edges.go
```diff
@@ -28,6 +93,28 @@ var KnowledgeBase = core.EdgeKB{
...
+			if targetGroup.RdsInstance == nil {
+				targetGroup.RdsInstance = instance
+			} else if targetGroup.RdsInstance != instance {
+				return fmt.Errorf("target group, %s, has edge to instance, %s, but internal property is set to a different instance", targetGroup.Name, instance.Name)
+			}
+
+			secret := targetGroup.RdsProxy.Auths[0].SecretArn.Resource.(*resources.Secret)
+			fmt.Println(secret.Id())
+			fmt.Println(dag.GetDownstreamResources(secret))
+			fmt.Println(dag.GetUpstreamDependencies(secret))
```
gordon-klotho, line 109: leftover debug

pkg/provider/aws/edges/edges.go
```diff
@@ -28,6 +93,28 @@ var KnowledgeBase = core.EdgeKB{
...
+			for _, res := range dag.GetDownstreamResources(secret) {
+				fmt.Println(res.Id())
```
gordon-klotho, line 111: ditto

pkg/provider/aws/execution_unit.go
```diff
@@ -34,6 +34,30 @@ func (a *AWS) expandExecutionUnit(dag *core.ResourceGraph, unit *core.ExecutionU
 	return nil
 }
 
+func (a *AWS) getLambdaConfiguration(result *core.ConstructGraph, dag *core.ResourceGraph, refs []core.AnnotationKey) (resources.LambdaFunctionConfigureParams, error) {
+	if len(refs) != 1 {
+		return resources.LambdaFunctionConfigureParams{}, fmt.Errorf("lambda must only have one construct reference")
+	}
+	lambdaConfig := resources.LambdaFunctionConfigureParams{}
+	construct := result.GetConstruct(refs[0].ToId())
```
gordon-klotho, line 42: Isn't this use-case what `MapResourceDirectlyToConstruct` was built for?

pkg/provider/aws/plugin.go
```diff
@@ -69,10 +69,18 @@ func (a *AWS) CopyConstructEdgesToDag(result *core.ConstructGraph, dag *core.Res
 }
 
 // configureResources calls every resource's Configure method, for resources that exist in the graph
-func (a *AWS) configureResources(dag *core.ResourceGraph) (err error) {
+func (a *AWS) configureResources(result *core.ConstructGraph, dag *core.ResourceGraph) (err error) {
 	var merr multierr.Error
 	for _, resource := range dag.ListResources() {
-		merr.Append(dag.CallConfigure(resource, nil))
+		var configuration any
+		if lambda, ok := resource.(*resources.LambdaFunction); ok {
```
gordon-klotho, line 76: Just looking forward, should this be a type switch?

pkg/provider/aws/resources/lambda.go
```diff
@@ -84,13 +84,29 @@ func (lambda *LambdaFunction) Create(dag *core.ResourceGraph, params LambdaCreat
...
 type LambdaFunctionConfigureParams struct {
-	ConnectionPoolConfigurationInfo ConnectionPoolConfigurationInfo
+	Timeout              int
+	MemorySize           int
+	EnvironmentVariables core.EnvironmentVariables
 }
 
 // Configure sets the intristic characteristics of a vpc based on parameters passed in
 func (lambda *LambdaFunction) Configure(params LambdaFunctionConfigureParams) error {
 	lambda.Timeout = 180
```
gordon-klotho, line 94: How do we propagate these defaults back into config so that the one we write in the compiled directory is accurate?

gordon-klotho (approved):

pkg/provider/aws/execution_unit.go
jhsinger-klotho: thats to get the resource which the construct is tied to, but here its the opposite. We would have to loop over all the keys in the map (since the keys are the constructs) and check the ids of their value

pkg/provider/aws/plugin.go
jhsinger-klotho: yeah it likely should. This will eventually be done very different because we dont want to think about constructs but for now let me change this

pkg/provider/aws/resources/lambda.go
jhsinger-klotho: well my thought is that the ones in the compiled directory are what are setting these unless they are completely overwritten by the specific exec unit or default. In doing so it requires us to create these by doing constraints on the nodes configuration itself which is still something we will have to figure out

PR #622, opened 2023-05-22
address all comments from pr chains

jhsinger-klotho:

addressing the actionable comments to unblock.

Creates new knowledge base pkg outside of core so we have some distinction. The kb methods now are on the kb object itself.

Some small changes surrounding caching in the kb methods to reduce allocation.

remove debug logging

other various small fixes

### Standard checks


pkg/knowledge_base/edge_kb.go
```diff
@@ -149,23 +150,23 @@ func isValidForPath(kb EdgeKB, edge Edge, dest reflect.Type) bool {
...
 		if !ok && dep.Properties.Data != nil {
-			merr.Append(fmt.Errorf("edge properties for edge %s -> %s, do not satisfy edge data format", dep.Source.Id().String(), dep.Destination.Id().String()))
+			merr.Append(fmt.Errorf("edge properties for edge %s -> %s, do not satisfy edge data format", dep.Source.Id(), dep.Destination.Id()))
 		} else if dep.Properties.Data != nil {
 			edgeData = data
 		}
 		// Find all possible paths given the initial source and destination node
-		paths := FindPaths(kb, reflect.TypeOf(dep.Source), reflect.TypeOf(dep.Destination))
+		paths := kb.FindPaths(reflect.TypeOf(dep.Source), reflect.TypeOf(dep.Destination))
 		validPaths := [][]Edge{}
```
gordon-klotho, line 170: nit: `[]Path`

pkg/provider/aws/knowledgebase/edges.go
```diff
@@ -1,18 +1,19 @@
...
+	knowledgebase "github.com/klothoplatform/klotho/pkg/knowledge_base"
 	"github.com/klothoplatform/klotho/pkg/provider/aws/resources"
 )
 
-var KnowledgeBase = core.EdgeKB{
-	core.Edge{From: reflect.TypeOf(&resources.LambdaFunction{}), To: reflect.TypeOf(&resources.RdsInstance{})}: core.EdgeDetails{
-		ExpansionFunc: func(from, to core.Resource, dag *core.ResourceGraph, data core.EdgeData) error {
-			lambda := from.(*resources.LambdaFunction)
-			instance := to.(*resources.RdsInstance)
+var AwsKB = knowledgebase.EdgeKB{
```
gordon-klotho, line 12: For convenience (not sure how others feel about this), we could alias some of the commonly used types - especially because the package name is the same.
ie:
```go
type (
    EdgeKB = knowledgebase.EdgeKB
    Edge = knowledgebase.Edge
    EdgeDetails = knowledgebase.EdgeDetails
    EdgeData = knowledgebase.EdgeData
)
```

pkg/provider/aws/resources/rds.go
```diff
@@ -239,6 +223,7 @@ func (proxy *RdsProxy) Create(dag *core.ResourceGraph, params RdsProxyCreatePara
 	if existingProxy != nil {
 		graphProxy := existingProxy.(*RdsProxy)
 		graphProxy.ConstructsRef = core.DedupeAnnotationKeys(append(graphProxy.KlothoConstructRef(), params.Refs...))
+		return nil
 	} else {
```
gordon-klotho, line 227: The `else` isn't needed anymore due to the return on the previous line.

gordon-klotho (approved):

pkg/provider/aws/knowledgebase/edges.go
jhsinger-klotho: yeah i think thats a solid idea, i can do that as i work on api gw

PR #623, opened 2023-05-22
API GW resource methods

jhsinger-klotho:

Adds the resources for API GW and other methods which are necessary to get the sample apps working.

### Standard checks

- **Unit tests**: Any special considerations? unit tests are added for all the added methods

### DIFF
--- a/pkg/provider/aws/expose.go
+++ b/pkg/provider/aws/expose.go
@@ -16,6 +16,28 @@ import (
 
 const API_GATEWAY_EXECUTION_CHILD_RESOURCES_IAC_VALUE = "child_resources"
 
+// expandOrm takes in a single orm construct and expands the generic construct into a set of resource's based on the units configuration.
+func (a *AWS) expandExpose(dag *core.ResourceGraph, expose *core.Gateway) error {
+	switch a.Config.GetExpose(expose.ID).Type {
+	case ApiGateway:
+		stage, err := core.CreateResource[*resources.ApiStage](dag, resources.ApiStageCreateParams{
+			AppName: a.Config.AppName,
+			Refs:    []core.AnnotationKey{expose.AnnotationKey},
+			Name:    expose.ID,
+		})
+		if err != nil {
+			return err
+		}
+		err = a.MapResourceToConstruct(stage.RestApi, expose)
+		if err != nil {
+			return err
+		}
+	default:
+		return fmt.Errorf("unsupported orm type %s", a.Config.GetExpose(expose.ID).Type)
+	}
+	return nil
+}
+
 // GenerateExposeResources will create the necessary resources within AWS to support a Gateway construct and its dependencies.
 func (a *AWS) GenerateExposeResources(gateway *core.Gateway, result *core.ConstructGraph, dag *core.ResourceGraph) error {
 	err := a.CreateRestApi(gateway, result, dag)
--- a/pkg/provider/aws/resources/api_gateway.go
+++ b/pkg/provider/aws/resources/api_gateway.go
@@ -2,23 +2,25 @@ package resources
 
 import (
 	"fmt"
+	"regexp"
+	"strings"
 
 	"github.com/klothoplatform/klotho/pkg/core"
 	"github.com/klothoplatform/klotho/pkg/sanitization/aws"
 )
 
 const (
-	API_GATEWAY_REST_TYPE        = "rest_api"
-	API_GATEWAY_RESOURCE_TYPE    = "api_resource"
-	API_GATEWAY_METHOD_TYPE      = "api_method"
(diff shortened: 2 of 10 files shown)


pkg/provider/aws/expose.go
```diff
@@ -16,6 +16,28 @@ import (
...
+		})
+		if err != nil {
+			return err
+		}
+		err = a.MapResourceToConstruct(stage.RestApi, expose)
+		if err != nil {
+			return err
+		}
+	default:
+		return fmt.Errorf("unsupported orm type %s", a.Config.GetExpose(expose.ID).Type)
```
gordon-klotho, line 36:
```suggestion
		return fmt.Errorf("unsupported expose type %s", a.Config.GetExpose(expose.ID).Type)
```

pkg/provider/aws/resources/api_gateway.go
```diff
@@ -85,6 +87,278 @@ type (
...
+type RestApiCreateParams struct {
+	AppName string
+	Refs    []core.AnnotationKey
+	Name    string
+}
+
+func (api *RestApi) Create(dag *core.ResourceGraph, params RestApiCreateParams) error {
+
+	name := restApiSanitizer.Apply(fmt.Sprintf("%s-%s", params.AppName, params.Name))
+	if params.AppName == "" {
```
gordon-klotho, line 99: Should this ever happen?

pkg/provider/aws/resources/api_gateway.go
```diff
@@ -85,6 +87,278 @@ type (
...
+	method.Authorization = "None"
+	if method.RequestParameters == nil {
+		method.RequestParameters = make(map[string]bool)
+	}
+	if params.RequestParameters != nil {
+		method.RequestParameters = params.RequestParameters
+	}
+	if params.Authorization != "" {
+		method.Authorization = params.Authorization
+	}
```
gordon-klotho, line 287: nit: these could just assign to the params value, since it's just checking against the zero value anyway.

pkg/provider/aws/resources/ecr.go
```diff
@@ -84,12 +83,21 @@ func (image *EcrImage) Create(dag *core.ResourceGraph, params ImageCreateParams)
...
 // Configure sets the intristic characteristics of a vpc based on parameters passed in
 func (image *EcrImage) Configure(params EcrImageConfigureParams) error {
 	image.ExtraOptions = []string{"--platform", "linux/amd64", "--quiet"}
+	if params.Dockerfile == "" {
+		return fmt.Errorf("image %s must have dockerfile set", image.Name)
+	}
+	image.Dockerfile = params.Dockerfile
+	if params.Context == "" {
+		return fmt.Errorf("image %s must have context set", image.Name)
+	}
```
gordon-klotho, line 99: 👍 good validation / precondition checking

gordon-klotho (approved):

pkg/provider/aws/resources/api_gateway.go
jhsinger-klotho: well those are already set by defaults i think when we do expansion so thats why i do the check. I should likely move that into here though so its more clear, but the request parameters i should then append to the existing list because those we have context of from the expansion func

pkg/provider/aws/resources/api_gateway.go
jhsinger-klotho: yeah because in the expansion we have the api name we just pass that directly in as the name so this way we can get the correct thing. I guess we could reverse engineer the name (sounds complex but just string replacement lol) to get the app name which might be cleaner. I think my takeaway is that our resources should also be namespaced by app for ease to work with

gordon-klotho: > we just pass that directly in as the name so this way we can get the correct thing

Not too sure what this means, can you point me to where that happens?

pkg/provider/aws/resources/api_gateway.go
gordon-klotho: The `params` don't have the defaults in them? I guess I'm confused then, I thought they would.

pkg/provider/aws/resources/api_gateway.go
jhsinger-klotho: its in the following pr, during the expansion for the integration https://github.com/klothoplatform/klotho/pull/624/files#diff-ab62ae487c13aa3db6ea2836ee2df671058485d34b5ee115a0fd67813ec5bc37R469

pkg/provider/aws/resources/api_gateway.go
jhsinger-klotho: well these wouldnt, i think realistically these shouldnt even be configurable yet because we dont allow it, so i could remove it. Was a little forward looking

PR #625, opened 2023-05-23
add S3Bucket.Create, and factor out a test util

yuval-klotho:
The bucket create is pretty simple, because it doesn't have any
dependencies.

I realized that its test was almost identical to that of
LamdbaFunction.Create, so I factored out a helper that I hope will
provide a nice pattern for testing these as we write them. I backported
the LambdaFunction's test as a check that there really is commonality,
but kept the rest unchanged just in the interest of time (and smaller
diff).

### Standard checks

- **Unit tests**: added, plus a refactor
- **Docs**: n/a

### DIFF
--- a/pkg/provider/aws/resources/lambda_test.go
+++ b/pkg/provider/aws/resources/lambda_test.go
@@ -11,15 +11,11 @@ import (
 
 func Test_LambdaCreate(t *testing.T) {
 	eu := &core.ExecutionUnit{AnnotationKey: core.AnnotationKey{ID: "test", Capability: annotation.ExecutionUnitCapability}}
-	cases := []struct {
-		name    string
-		lambda  *LambdaFunction
-		want    coretesting.ResourcesExpectation
-		wantErr bool
-	}{
+
+	cases := []coretesting.CreateCase[LambdaCreateParams, *LambdaFunction]{
 		{
-			name: "nil lambda",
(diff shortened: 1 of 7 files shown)


pkg/provider/aws/resources/lambda_test.go
```diff
@@ -34,44 +30,26 @@ func Test_LambdaCreate(t *testing.T) {
...
-				return
-			}
-			tt.want.Assert(t, dag)
-
-			graphLambda := dag.GetResource(lambda.Id())
-			lambda = graphLambda.(*LambdaFunction)
 
-			assert.Equal(lambda.Name, "my-app-test")
-			assert.ElementsMatch(lambda.ConstructsRef, []core.AnnotationKey{eu.AnnotationKey})
+			tt.Run(t)
```
jhsinger-klotho, line 52: can we do other checks on the result in this format or no

jhsinger-klotho: oh i see now, nevermind, i think this looks good

yuval-klotho: Yeah, and I purposefully made the `Checks func` required (if `!WantError`), so that people don't accidentally forget. I figure even if they want a no-op check, that forces them to at least opt into it.

PR #626, opened 2023-05-23
fix nil-deref effor in Subnet.Id()

yuval-klotho:
The `TestAllTypesHaveIcons` test calls `Id()` on each resource, to get its type. This was failing for Subnets that hadn't been fully initialized with a VPC yet.

### Standard checks

- **Unit tests**: fixes it
- **Docs**: n/a

### DIFF
--- a/pkg/provider/aws/resources/vpc.go
+++ b/pkg/provider/aws/resources/vpc.go
@@ -631,14 +631,10 @@ func (subnet *Subnet) KlothoConstructRef() []core.AnnotationKey {
 
 // Id returns the id of the cloud resource
 func (subnet *Subnet) Id() core.ResourceId {
-	var namespace string
-	if subnet.Vpc != nil {
-		namespace = subnet.Vpc.Name
-	}
 	id := core.ResourceId{
 		Provider:  AWS_PROVIDER,
 		Type:      VPC_SUBNET_TYPE_PREFIX + subnet.Type,
-		Namespace: namespace,
+		Namespace: subnet.Vpc.Name,
 		Name:      subnet.Name,
 	}
 	return id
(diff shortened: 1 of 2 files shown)


pkg/provider/aws/resources/vpc.go
```diff
@@ -631,10 +631,14 @@ func (subnet *Subnet) KlothoConstructRef() []core.AnnotationKey {
 
 // Id returns the id of the cloud resource
 func (subnet *Subnet) Id() core.ResourceId {
+	var namespace string
+	if subnet.Vpc != nil {
+		namespace = subnet.Vpc.Name
+	}
```
jhsinger-klotho, line 637: i think we want to have a vpc always here, is it easy enough to make our tests comply

yuval-klotho: `TestAllTypesHaveIcons` does something similar to the `all_types_test.go` from ifc2: it just iterates through a blank set of resources, each created as `&Foo{}`.

We could just hard-code an empty VPC into the empty Subnet's creation ([link][1]):
```diff
-&Subnet{},
+&Subnet{Vpc: &Vpc{}}
```

Sounds like you'd prefer that?

To me, the new `Create(dag, params)` paradigm implies that resources are at least somewhat usable before they're fully hooked into the rest of the graph, and this change (as I currently have it) supports that. But if that's not what you had in mind, I can hack that empty VPC into the empty Subnet for testse.

[1]: https://github.com/klothoplatform/klotho/blob/main/pkg/provider/aws/resources/list_all.go#L53

jhsinger-klotho: well i think in our configure for a subnet well likely want to ensure its vpc isnt null. i guess this is the only thing we have namespaced so far so i think we might have to make the decision, when namespacing something, does that something (in this case the vpc) need to exist

yuval-klotho: I think the bigger question is, how fully does a resource need to be configured before you can grab any of its ID fields? For example, if you reflectively create a `Foo`, can you then call `Id().Provider` on it in order to get its provider? It sounds like your answer is no: you need to first do some work to configure it (which today is "no-op for most resources, set its VPC for Subnets" — but in the general case, the answer is therefore "you may have to do something").

jhsinger-klotho (approved):

Issue #603, opened 2023-05-18
Fs Resource Create and Edge Expansion


jhsinger-klotho:
This task is meant to satisfy the Fs construct's expansion.

To do so we will:

- Add Create Methods for Each Resource necessary
    - [x] S3Bucket
- Add Configure Method For Each Resource Necessary
    - [x] S3Bucket
- Add the expand method for an fs construct and our internal resource construct
- My initial thought is that this will connect an edge from the method to their corresponding lambda
- Add EdgeKB data
    - [x] LambdaFunction -> S3Bucket
    - [x] IamPolicy -> S3Bucket

DoD:

- All Methods above are written and have unit tests
- ts-cloudfs (or any cloudfs) sample app is able to pass tests [ts-cloudfs](https://github.com/klothoplatform/sample-apps/tree/main/ts-cloudfs)

yuval-klotho:
Similarly, removing the S3BucketPolicy — we need that for static files (#607), but not for FS

PR #627, opened 2023-05-23
add S3Bucket.Configure

yuval-klotho:
Trivial PR — but I think this is all I need for S3Bucket's `Configure`, right?

In support of #603 

### Standard checks

- **Unit tests**: none needed
- **Docs**: n/a

### DIFF
--- a/pkg/core/resources.go
+++ b/pkg/core/resources.go
@@ -31,6 +31,11 @@ type (
 		Create(dag *ResourceGraph, param K) error
 	}
 
+	ConfigurableResource[K any] interface {
+		Resource
+		Configure(params K) error
+	}
+
 	ResourceId struct {
 		Provider string
 		Type     string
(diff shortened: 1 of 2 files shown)


pkg/core/resources.go
```diff
@@ -31,6 +31,11 @@ type (
 		Create(dag *ResourceGraph, param K) error
 	}
 
+	ConfigurableResource[K any] interface {
+		Resource
+		Configure(params K) error
+	}
+
```
jhsinger-klotho, line 38: what do we do with this?

jhsinger-klotho (approved):
nothing blocking, just curious what the configurable resource interface is going to be used for

pkg/core/resources.go
yuval-klotho: I just find it helpful in the IDE — it helps me find all the other implementations of Configure, if I want to use those as a template.

PR #630, opened 2023-05-24
add eks resources

jhsinger-klotho:

this pr is a prereq to https://github.com/klothoplatform/klotho/pull/631, but tests dont pass yet due to changes in the next pr (unfortunately some cleanup caused failure). So when both are approved i will cancel this one and purely work off the other. This was meant to make it more readable and separation of resource methods vs expansion

Adding EKS Resource methods and edges in preparation for EKS expansion

Also added a testing utility for configure methods, similar to the one for create methods for ease of testing

### Standard checks

- **Unit tests**: Any special considerations? yup, each method should have a unit test

### DIFF
--- a/pkg/provider/aws/resources/eks.go
+++ b/pkg/provider/aws/resources/eks.go
@@ -8,9 +8,7 @@ import (
 	"reflect"
 	"strings"
 
-	"go.uber.org/zap"
-
-	"github.com/klothoplatform/klotho/pkg/config"
+	"github.com/klothoplatform/klotho/pkg/collectionutil"
 	"github.com/klothoplatform/klotho/pkg/core"
 	"github.com/klothoplatform/klotho/pkg/infra/kubernetes"
 	"github.com/klothoplatform/klotho/pkg/sanitization/aws"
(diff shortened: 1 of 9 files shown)


pkg/provider/aws/resources/eks.go
```diff
@@ -144,192 +142,261 @@ type (
...
-	dag.AddResource(clusterRole)
+	cluster.Name = clusterSanitizer.Apply(fmt.Sprintf("%s-%s", params.AppName, params.Name))
 
-	cluster := NewEksCluster(appName, clusterName, vpc, subnets, securityGroups, clusterRole)
-	cluster.ConstructsRef = references
-	dag.AddDependenciesReflect(cluster)
+	existingCluster := dag.GetResource(cluster.Id())
+	if existingCluster != nil {
+		graphCluster := existingCluster.(*EksCluster)
+		graphCluster.ConstructsRef = append(graphCluster.ConstructsRef, params.Refs...)
```
DavidSeptimus-Klotho, line 158: do we need to deduplicate refs here?

DavidSeptimus-Klotho (approved):

pkg/provider/aws/resources/eks.go
jhsinger-klotho: ah yeah good catch, will make that change

PR #634, opened 2023-05-24
edges for fs, and plugging it into plugin.go

yuval-klotho:
In service of (and maybe resolves?) #603. `ts-cloudfs` is working (manually tested), which is that ticket's DoD.

### Standard checks

- **Unit tests**: lmk if any are missing
- **Docs**: n/a


jhsinger-klotho (approved):
LGTM!

PR #635, opened 2023-05-24
expose eks resources

### LINKED ISSUE #606: Eks Expose Resource Create and Edge Expansion
This task is meant to satisfy the Fs construct's expansion.

To do so we will:

Note: Our kubernetes resources may not quite fit the create and configure structure (Mainly helm charts and manifests). This is because their configuration is not generic and has to be set beforehand.

   - Resources:
      - VpcLink
      - TargetGroup
      - Listener
      - LoadBalancer
- modify logic in copy edges to point the api methods to the helm chart
- Add EdgeKB data
   - Possible Edges
      - ApiIntegration -> VpcLinke
      - VpcLink -> Vpc
      - Listener -> LoadBalancer
      - TargetGroup -> Vpc
      - LoadBalancer -> Subnet & SG

DoD:

- go expose sample app can run on eks

jhsinger-klotho:
• Does it relate to or fix any issue? closes #606 

We expand api -> target group because we dont have an understanding of what exists within the helm chart. We have the target groups created as a part of the expansion of the helm chart itself (by values) and can use those to determine which routes (by ref to exec unit name) go to which target groups

### Standard checks

- **Unit tests**: Any special considerations? added a unit test for the expansion since its more complex


pkg/provider/aws/knowledgebase/api_gateway.go
```diff
@@ -109,6 +109,97 @@ var ApiGatewayKB = knowledgebase.Build(
 			return nil
 		},
 	},
+	knowledgebase.EdgeBuilder[*resources.ApiIntegration, *resources.LoadBalancer]{
+		Expand: func(integration *resources.ApiIntegration, lb *resources.LoadBalancer, dag *core.ResourceGraph, data knowledgebase.EdgeData) error {
+			// This isnt an instance of an expanded path, rather an existing edge so ignore
+			if integration.Name != "" {
+				return nil
+			}
+			restApi, ok := data.Source.(*resources.RestApi)
```
DavidSeptimus-Klotho, line 118: for my own understanding -- are the source and destination in the `EdgeData` the original, non-expanded edge's nodes?

pkg/provider/aws/knowledgebase/edge_expansion_test.go
```diff
@@ -177,6 +177,115 @@ func Test_ExpandEdges(t *testing.T) {
...
+					"aws:vpc_link:aws:load_balancer:my-app-tg",
+				},
+				Deps: []coretesting.StringDep{
+					{Source: "aws:api_integration:my-app-/-GET", Destination: "aws:api_method:my-app-/-GET"},
+					{Source: "aws:api_integration:my-app-/-GET", Destination: "aws:load_balancer:my-app-tg"},
+					{Source: "aws:api_integration:my-app-/-GET", Destination: "aws:rest_api:api"},
+					{Source: "aws:api_integration:my-app-/-GET", Destination: "aws:vpc_link:aws:load_balancer:my-app-tg"},
+					{Source: "aws:api_integration:my-app-/route/1-GET", Destination: "aws:api_method:my-app-/route/1-GET"},
+					{Source: "aws:api_integration:my-app-/route/1-GET", Destination: "aws:api_resource:my-app-/route/1"},
+					{Source: "aws:api_integration:my-app-/route/1-GET", Destination: "aws:load_balancer:my-app-tg"},
```
DavidSeptimus-Klotho, line 234: do we know where that trailing dash on `my-app-` is coming from? It would be nice if we could get rid of it.

pkg/provider/aws/resources/eks.go
```diff
@@ -387,6 +386,25 @@ func (nodeGroup *EksNodeGroup) Configure(params EksNodeGroupConfigureParams) err
 	return nil
 }
 
+func (cluster *EksCluster) SetUpDefaultNodeGroup(dag *core.ResourceGraph, appName string) error {
+	_, err := core.CreateResource[*EksNodeGroup](dag, EksNodeGroupCreateParams{
+		InstanceType: "t3.medium",
+		NetworkType:  PrivateSubnet,
+		Refs:         cluster.ConstructsRef,
+		AppName:      appName,
+		ClusterName:  strings.TrimLeft(cluster.Name, fmt.Sprintf("%s-", appName)),
```
DavidSeptimus-Klotho, line 395: Not really sure how I feel about this pattern, but it's hard to argue with shortening any of our names

DavidSeptimus-Klotho (approved):

pkg/provider/aws/knowledgebase/api_gateway.go
jhsinger-klotho: yup, i believe i put a godoc for that, but i will do another pass on godocs to help explain it all

pkg/provider/aws/knowledgebase/edge_expansion_test.go
jhsinger-klotho: its because we do appname-route. we could change that up if we want to though

pkg/provider/aws/resources/eks.go
jhsinger-klotho: i also dont like it, i think we should take the time to make an appname global to all ids so that we dont have to prepend it onto everything. Maybe thats a high value task

pkg/provider/aws/knowledgebase/api_gateway.go
```diff
@@ -109,6 +109,99 @@ var ApiGatewayKB = knowledgebase.Build(
...
+			refs := core.AnnotationKeySet{}
+			refs.AddAll(lb.ConstructsRef)
+			refs.AddAll(restApi.ConstructsRef)
```
gordon-klotho, line 121: nit:
```go
lb.ConstructsRef.CloneWith(restApi.ConstructsRef)
```
This is a common pattern, so there's a special method for it.

pkg/provider/aws/knowledgebase/api_gateway.go
```diff
@@ -109,6 +109,99 @@ var ApiGatewayKB = knowledgebase.Build(
...
+			listener.LoadBalancer.Scheme = "internal"
+			listener.DefaultActions = []*resources.LBAction{{TargetGroupArn: core.IaCValue{Resource: tg, Property: resources.ARN_IAC_VALUE}, Type: "forward"}}
+			listener.Port = 80
+			listener.Protocol = "TCP"
+
+			for _, route := range data.Routes {
+				var err error
+				integration, err = core.CreateResource[*resources.ApiIntegration](dag, resources.ApiIntegrationCreateParams{
+					AppName:    data.AppName,
+					Refs:       refs,
```
gordon-klotho, line 157: This should `Clone` the refs, so that each gets its own set.

pkg/provider/aws/knowledgebase/api_gateway.go
jhsinger-klotho: 👍 changed it to this, good callout, need to make my brain remember these patterns now

PR #636, opened 2023-05-25
Secrets Expansion

### LINKED ISSUE #609: secret and config resource create and expansion
Secrets should be fully configured from the RDS Work, so this is just to write the expand function for the secret construct as well as the configureNodes section.

DoD:
Any unit testing on the above work
Tests pass for [ts-secrets](https://github.com/klothoplatform/sample-apps/tree/main/ts-secrets) or relative sample app
Tests pass for [go-secrets](https://github.com/klothoplatform/sample-apps/tree/main/go-secrets)

DavidSeptimus-Klotho:
Resolves #609 

This PR implements the following:

- Expansion of `Secrets` and `Config` constructs
- Expansion of `LambdaFunction` -> `Secret` edges
- Supresses a longstanding error message that occurred when compiling with the `--verbose` flag (no functional impact)

This change was verified to work with the `ts-secrets` and `go-secrets` sample apps.

**ts-secrets:**

image
image

**go-secrets**

image
image


pkg/lang/javascript/plugin_persist.go
```diff
@@ -161,27 +161,27 @@ func (p *persister) handleFile(f *core.SourceFile, unit *core.ExecutionUnit) ([]
...
+			construct, transformErr = p.transformSecret(f, annot, pResult)
 			runtimeErr = p.runtime.AddSecretRuntimeFiles(unit)
 		case *core.Orm:
-			resource, transformErr = p.transformORM(unit, f, annot, pResult)
+			construct, transformErr = p.transformORM(unit, f, annot, pResult)
 			runtimeErr = p.runtime.AddOrmRuntimeFiles(unit)
 		case *core.RedisCluster:
-			resource, transformErr = p.transformRedis(unit, f, annot, pResult, keyType)
+			construct, transformErr = p.transformRedis(unit, f, annot, pResult, keyType)
 			runtimeErr = p.runtime.AddRedisClusterRuntimeFiles(unit)
```
jhsinger-klotho, line 182: good renaming, just curious as to what made you look at this plugin?

pkg/provider/aws/config_resources.go
```diff
@@ -1,18 +1,38 @@
...
+	mapped := false
+	for _, r := range dag.GetDownstreamResources(secretVersion) {
+		if secret, ok := r.(*resources.Secret); ok {
+			if err := a.MapResourceToConstruct(secret, construct); err != nil {
+				return err
+			}
+			mapped = true
+			break
+		}
+	}
```
jhsinger-klotho, line 33: cant you call secretVersion.Secret here rather than looking at the downstream resources?

pkg/provider/aws/knowledgebase/iam.go
```diff
@@ -18,9 +18,17 @@ var IamKB = knowledgebase.Build(
...
+			secretPolicyDoc := resources.CreateAllowPolicyDocument([]string{"secretsmanager:DescribeSecret", "secretsmanager:GetSecretValue"}, []core.IaCValue{{Resource: secret, Property: resources.ARN_IAC_VALUE}})
+			inlinePol := resources.NewIamInlinePolicy(fmt.Sprintf("%s-secretpolicy", secret.Name), core.DedupeAnnotationKeys(append(role.ConstructsRef, secret.ConstructsRef...)), secretPolicyDoc)
+			role.InlinePolicies = append(role.InlinePolicies, inlinePol)
+			return nil
+		},
+	},
 	knowledgebase.EdgeBuilder[*resources.IamPolicy, *resources.SecretVersion]{
 		Configure: func(policy *resources.IamPolicy, secretVersion *resources.SecretVersion, dag *core.ResourceGraph, data knowledgebase.EdgeData) error {
-			secretPolicyDoc := resources.CreateAllowPolicyDocument([]string{"secretsmanager:GetSecretValue"}, []core.IaCValue{{Resource: secretVersion.Secret, Property: resources.ARN_IAC_VALUE}})
+			secretPolicyDoc := resources.CreateAllowPolicyDocument([]string{"secretsmanager:DescribeSecret", "secretsmanager:GetSecretValue"}, []core.IaCValue{{Resource: secretVersion.Secret, Property: resources.ARN_IAC_VALUE}})
```
jhsinger-klotho, line 31: do we need this policy on both of these here? I know that the rds dep i believe is the secretversion, so that wouldnt need describe. Im assuming that config vs secrets uses inline vs policy? Describe doesnt hurt, but a thought for the future is how do we know read vs write when we want to have granular permissions

pkg/provider/aws/resources/secret.go
```diff
@@ -10,13 +10,12 @@ import (
...
 		Name          string
-		SecretName    string
 		ConstructsRef []core.AnnotationKey
 	}
 
 	SecretVersion struct {
-		SecretName    string
 		Secret        *Secret
+		DetectedPath  string
 		Path          string
```
jhsinger-klotho, line 19: whats the difference between these?

pkg/provider/aws/secret.go
```diff
@@ -2,42 +2,63 @@ package aws
...
+			if secret, ok := r.(*resources.Secret); ok {
+				if err := a.MapResourceToConstruct(secret, construct); err != nil {
+					return err
+				}
+				mapped = true
+				break
+			}
+		}
+		if !mapped {
+			return fmt.Errorf("unable to find secret resource for secret version, %s", secretVersion.Id())
```
jhsinger-klotho, line 34: how do we map multiple secrets to the same construct? i forgot to take that into account when i assumed it was 1:1 because our secrets capability is the only one which doesnt follow that.

pkg/provider/aws/knowledgebase/iam.go
gordon-klotho: I'm not sure we should be mapping anything directly to the SecretVersion. If the user updated the secret, that would invalidate all the policies to read it. The permissions on the Secret should suffice for everything we need to do.

pkg/infra/iac2/templates/secret_version/factory.ts
```diff
@@ -10,19 +10,21 @@ interface Args {
...
-                //TMPL {{- if eq .Type.Raw "string" }}
-                secretString: fs.readFileSync(args.Path, 'utf-8').toString(),
-                //TMPL {{ else }}
-                //TMPL secretBinary: fs.readFileSync({{ .Path.Parse }}, 'base64').toString(),
-                //TMPL {{ end }}
-            },
-            { protect: args.protect }
-        )
-    }
+function create(args: Args): aws.secretsmanager.SecretVersion | undefined {
```
gordon-klotho, line 13: Not sure if we need to reference the SecretVersion anywhere. This is mostly provided as a deploy-time convenience than it is for anything related to the topology.

pkg/provider/aws/resources/secret.go
DavidSeptimus-Klotho: `DetectedPath` is set on creation and doesn't change -- it's intended more for the `Secrets` construct use-case. `Path` is configurable and defaults to `DetectedPath` if it's not set in klotho config. This helps with `Secrets` construct expansion where correlating a specific `Secret` resource back to a specific secret path can be awkward.

pkg/provider/aws/secret.go
DavidSeptimus-Klotho: We iterate over the construct's `Secrets.Secrets` slic, create a `Secret` resource and populate the resource's `DetectedPath` value. Then we set each secret's `Path` field  then set the `Secret.Path` value as part of `configureResources()` to its `DetectedPath`. If the construct is `Config` instead of `Secrets`, we ignore the resource's `DetectedPath` field and instead retrieve it from klotho config.

pkg/infra/iac2/templates/secret_version/factory.ts
DavidSeptimus-Klotho: We don't currently, but I don't think that's a reason to specifically not create a variable in the generated  `index.ts` file.

pkg/provider/aws/knowledgebase/iam.go
DavidSeptimus-Klotho: @gordon-klotho I believe you're right. That edge was created as part of the RDS expansion change, but I can change it to `Secret` and verify that it still works.

pkg/infra/iac2/templates/secret_version/factory.ts
gordon-klotho: The reason it was as a void function (thus, not assignable to a variable) was because we'd have to sacrifice some readability to return the value from the factory function to make it assignable to a variable. Use-cases like these were why the void function support was added initially. IMO the `(() => {})()` pattern adds some cognitive overhead - one that I prefer to avoid if possible.

pkg/provider/aws/knowledgebase/iam.go
DavidSeptimus-Klotho: @jhsinger-klotho yeah, dealing with more granular permissions is something we'll need to deal with at some point.

jhsinger-klotho (approved):
I still find our secrets odd since they have their own unique behavior but for now good to leave as is

pkg/provider/aws/config_resources.go
DavidSeptimus-Klotho: Yeah, not sure how I missed that. _I think it was Copilot messing with my IDE's autocomplete._

pkg/lang/javascript/plugin_persist.go
DavidSeptimus-Klotho: I was just backtracking to understand the the how we handle secret detection.

pkg/provider/aws/secret.go
```diff
@@ -2,42 +2,59 @@ package aws
...
-		a.PolicyGenerator.AddInlinePolicyToUnit(unit.Id(), policy)
+		a.MapResourceDirectlyToConstruct(secretVersion.Secret, construct)
 	}
 	return nil
 }
+
+func (a *AWS) getSecretVersionConfiguration(secretVersion *resources.SecretVersion, result *core.ConstructGraph) (resources.SecretVersionConfigureParams, error) {
+	secretVersionConfig := resources.SecretVersionConfigureParams{}
+	if len(secretVersion.ConstructsRef) == 0 {
+		// this case may occur when a secret is created as part of edge expansion and is configured as part of that process
```
jhsinger-klotho, line 32: the version would still have refs about the rds and lambda/exec unit that caused its creation so wouldnt we error here?

DavidSeptimus-Klotho:
This change has been verified with the following samples:
- ts-sequelize (testing configuration of RDS proxy secrets)
- ts-secrets (modified to use 2 fs secrets annotations, one of them including multiple secrets)
- go-secrets (modified to use 2 config/secrets instead of one)

Issue #607, opened 2023-05-18
Static Unit and Cloudfront Expansion


jhsinger-klotho:
This task is meant to satisfy the Expose construct's expansion.

To do so we will:
- Add Create Methods for Each Resource necessary
- Add Configure Method For Each Resource Necessary
- Resources:
    - CloudfrontDistribution
    - OriginAccessIdentity
 - Add the expand method for an expose construct
 - Add the copyEdges logic
 - Add EdgeKB data
     - Possible Edges
        - CloudfrontDistribution -> S3Bucket
        - CloudfrontDistribution -> RestApi

DoD:
- All Methods above are written and have unit tests
- Deploy the [ts-static-unit](https://github.com/klothoplatform/sample-apps/tree/main/ts-static-unit) sample app and verify it works both behind a cdn and without a cdn
- [go-expose](https://github.com/klothoplatform/sample-apps/tree/main/go-expose) sample app is able to pass tests when the cdn id is set on the expose construct

PR #640, opened 2023-05-25
static units (no CDN or Cloudfront yet)

yuval-klotho:
In service of #607 

`ts-static-unit` succeeds, as long as you don't ask for a CDN.

> **Note**
> Relative to the version before node expansion (8d98439), this version removes the payloads bucket (`aws:s3:Bucket` `ystest--nternal-lotho-ayloads` when I ran it). Please verify that that's correct.

### Standard checks

- **Unit tests**: added
- **Docs**: n/a


pkg/provider/aws/resources/s3.go
```diff
@@ -102,14 +103,52 @@ func (bucket *S3Bucket) Configure(params S3BucketConfigureParams) error {
...
+//
+//func NewS3Object(bucket *S3Bucket, objectName string, key string, path string) *S3Object {
+//	return &S3Object{
+//		Name:          objectSanitizer.Apply(fmt.Sprintf("%s-%s", bucket.Name, objectName)),
+//		ConstructsRef: bucket.KlothoConstructRef(),
+//		Key:           key,
+//		FilePath:      path,
+//		Bucket:        bucket,
+//	}
+//}
```
jhsinger-klotho, line 115: can we delete these or do we want to wait until its fluly implemented?

pkg/provider/aws/resources/s3.go
```diff
@@ -102,14 +103,52 @@ func (bucket *S3Bucket) Configure(params S3BucketConfigureParams) error {
...
+	Key      string
+	FilePath string
+}
+
+func (object *S3Object) Create(dag *core.ResourceGraph, params S3ObjectCreateParams) error {
+	bucket, err := core.CreateResource[*S3Bucket](dag, S3BucketCreateParams{
+		AppName: params.AppName,
+		Refs:    core.AnnotationKeySetOf(params.Ref),
+		Name:    params.Ref.ID,
+	})
```
jhsinger-klotho, line 130: you should be able to use the create dependencies call here. Not going to make too big  of a difference since theres only one dep, but i think its a good pattern to use to have centralized logic

pkg/provider/aws/resources/s3.go
```diff
@@ -102,14 +103,52 @@ func (bucket *S3Bucket) Configure(params S3BucketConfigureParams) error {
...
+	dag.AddDependency(object, bucket)
+	return nil
+}
+
+type S3ObjectConfigureParams struct{}
+
+func (object *S3Object) Configure(params S3ObjectConfigureParams) error {
+	// nothing
+	return nil
 }
```
jhsinger-klotho, line 152: nit: dont have to write this if nothing happens

pkg/provider/aws/static_unit.go
```diff
@@ -4,20 +4,47 @@ import (
...
+			Ref:      unit.AnnotationKey,
+			Name:     filepath.Base(f.Path()),
+			Key:      f.Path(),
+			FilePath: filepath.Join(unit.ID, f.Path()),
+		})
+		if err != nil {
+			errs.Append(err)
+			continue
+		}
+		err = a.MapResourceToConstruct(object, unit)
```
jhsinger-klotho, line 40: is this using the map which only allows 1 resource to 1 construct? If so i would think we would want to map the bucket and not all the objects

pkg/provider/aws/static_unit.go
```diff
@@ -4,20 +4,47 @@ import (
...
+	return errs.ErrOrNil()
+}
+
+func (a *AWS) getStaticUnitObjectConfiguration() resources.S3ObjectConfigureParams {
+	return resources.S3ObjectConfigureParams{
+		// no fields; provided for consistency with other resources
 	}
-	a.MapResourceDirectlyToConstruct(bucket, unit)
-	return nil
 }
```
jhsinger-klotho, line 50: also might  not need this if it doesnt do anything i think

pkg/core/coretesting/expandable_resource.go
```diff
@@ -10,7 +10,7 @@ import (
 
 type CreateCase[P any, R core.ExpandableResource[P]] struct {
 	Name     string
-	Existing R
+	Existing []core.Resource
```
jhsinger-klotho, line 13: why change this to an array?

yuval-klotho: The more significant bit of the change is that it goes from an `R` to any `core.Resource`. I use this to test an S3Object.Create when there's already an existing S3Bucket. I thought I may need it as an array too, and though I didn't end up needing that, I figured I'd keep it since I'd already done it.

Do you prefer it as single?

pkg/provider/aws/resources/s3.go
yuval-klotho: oops, yeah I just forgot to delete

pkg/provider/aws/resources/s3.go
yuval-klotho: Hm, can I use it even though the S3Object doesn't yet have its final name (and thus string id)? The S3Object can't determine its name until the S3Bucket gets created and gives itself its name.

pkg/provider/aws/resources/s3.go
yuval-klotho: Are there existing patterns of it being omitted? Personally I like it being there — it makes it obvious/clear that it's empty by design, as opposed to someone forgetting.

pkg/provider/aws/static_unit.go
yuval-klotho: Oh, yes. That's a mistake — I had meant to call `MapResourceDirectlyToConstruct`.

I'm actually pretty unclear what the difference is between `MapResourceDirectlyToConstruct` and `MapResourceToConstruct`, now that you mention it. Why do we need the second one?

pkg/provider/aws/resources/s3.go
jhsinger-klotho: well so i think we want to try to avoid things like that, i would take in the param of BucketName and Name (not base the name of a reference in case we arent talking in terms of constructs) to generate the object and bucket name

jhsinger-klotho: this is what we do in the eks scenario to ensure the cluster gets created correctly and the fargate profile or node groups name correlates to the cluster

pkg/provider/aws/resources/s3.go
jhsinger-klotho: yeah we dont have many configure calls to be honest, we dont call it if it doesnt exist on the resource so it wont matter.

pkg/provider/aws/static_unit.go
jhsinger-klotho: well in this scenario i would think we just map the bucket to the static unit, so we shouldnt. I think david found a way to make secrets work as well, so we shouldnt need `MapResourceDirectlyToConstruct`, its just not deleted yet but was legacy

pkg/core/coretesting/expandable_resource.go
jhsinger-klotho: yeah i think so just because thats how all the tests are structured (even though they dont all use the testing utility yet)

pkg/provider/aws/resources/s3.go
yuval-klotho: ok

jhsinger-klotho (approved):

PR #642, opened 2023-05-26
Eks proxy

### LINKED ISSUE #605: Eks + persist Edge Expansion
This will follow #604 and eks expose.

We need to define edge data for the following connections
     - HelmChart -> S3Bucket
      - HelmChart -> RdsInstance
      - HelmChart -> RdsProxy
      - HelmChart -> Elasticache
      - HelmChart -> DynamoDB

DoD:
- Any applicable unit tests
- Each Sample app for the basic persistence testing, can pass with an eks exec unit

### LINKED ISSUE #611: eks proxy resource create and expansion
This task is meant to satisfy the eks proxy

To do so we will:
- Resources:
    - PrivateDnsNamespace
 - Add the expand method for an expose construct
 - Add the copyEdges logic
 - Add EdgeKB data
     - Possible Edges
        - IamRole -> PrivateDnsNamespace
        - Lambda -> PrivateDnsNamespace
        - HelmChart -> PrivateDnsNamespace
        - PrivateDnsNamespace -> HelmChart

ensure the cloudmap controller is installed on the cluster in this scenario

DoD:
- Deploy the [ts-microservices](https://github.com/klothoplatform/sample-apps-pro/tree/main/ts-microservices) in the pro repo and verify you can proxy TO an eks exec unit

jhsinger-klotho:
• Does it relate to or fix any issue? closes #605 and closes #611

- Does the work for eks proxy. We are able to fully do it from the lambda -> helm chart edge but not helm chart -> helm chart since we dont have the understanding of what we are modifying within the helm chart itself (we should eventually remove this when we break down the k8s objects) (we also may have other helm chart -> helm chart dependencies which do not mean proxy, and that would force us to use our knowledge of constructs in the KB)

- Create the edges for k8s exec unit -> data

Tested lambda -> eks and eks -> eks and eks -> ddb through the ts-microservices sample-apps-pro
```
(base) jordansinger@Jordans-MacBook-Pro ts-microservices % npm run integ-test

> microservices-sample-app@1.0.0 integ-test
> jest --group=integration

 PASS  test/happy_path.test.ts
  Happy Path Test
    ✓ Add User (524 ms)
    ✓ Get User (348 ms)

Test Suites: 1 passed, 1 total
Tests:       2 passed, 2 total
Snapshots:   0 total
Time:        1.48 s, estimated 8 s
Ran all test suites.
```

tested py-orm,  py-redis and py-cloudfs as well

```
(base) jordansinger@Jordans-MacBook-Pro py-orm % curl -X PUT  https://629dokyqi8.execute-api.us-west-2.amazonaws.com/stage
"Success"%
(base) jordansinger@Jordans-MacBook-Pro py-orm % curl https://629dokyqi8.execute-api.us-west-2.amazonaws.com/stage/patrick
```

### Standard checks


pkg/provider/aws/knowledgebase/iam.go
```diff
@@ -141,6 +141,10 @@ var IamKB = knowledgebase.Build(
 	},
 	knowledgebase.EdgeBuilder[*resources.IamRole, *resources.OpenIdConnectProvider]{
 		Configure: func(role *resources.IamRole, oidc *resources.OpenIdConnectProvider, dag *core.ResourceGraph, data knowledgebase.EdgeData) error {
+			if strings.Contains(role.Name, "alb-controller") {
```
gordon-klotho, line 144: Not sure how I feel about relying on the `Name` field programmatically. Probably fine for now, but maybe there's another way using dependencies or something that would be more robust.

pkg/provider/aws/plugin.go
```diff
@@ -23,7 +23,7 @@ func (a *AWS) ExpandConstructs(result *core.ConstructGraph, dag *core.ResourceGr
 			merr.Append(a.expandExpose(dag, construct))
 		case *core.Orm:
 			merr.Append(a.expandOrm(dag, construct))
-		case *core.Fs:
+		case *core.Fs, *core.InternalResource:
```
gordon-klotho, line 26: Are all `InternalResource`s Fs? Seems like this is too generically named or that there's a field or something on there that should indicate how to expand it.

pkg/provider/aws/plugin.go
```diff
@@ -52,21 +52,50 @@ func (a *AWS) CopyConstructEdgesToDag(result *core.ConstructGraph, dag *core.Res
...
-			switch dep.Destination.(type) {
+			if a.Config.GetExecutionUnit(construct.ID).Type == kubernetes.KubernetesType {
+				data.SourceRef = construct.AnnotationKey
+			}
+			switch dest := dep.Destination.(type) {
 			case *core.Orm:
 				data.Constraint = knowledgebase.EdgeConstraint{
 					NodeMustExist: []core.Resource{&resources.RdsProxy{}},
 				}
+				// Because we dont understand resource types, our edges cant depict what is truly valid and what isnt yet. Example
```
gordon-klotho, line 63: > Because we dont understand resource types

Not sure what this means - aren't resource types all over the knowledge base and resource graph?

pkg/provider/aws/plugin.go
```diff
@@ -96,7 +125,9 @@ func (a *AWS) CopyConstructEdgesToDag(result *core.ConstructGraph, dag *core.Res
 				data.Destination = targetTG
 			}
 		}
-		dag.AddDependencyWithData(sourceResource, targetResource, data)
+		if sourceResource != targetResource {
+			dag.AddDependencyWithData(sourceResource, targetResource, data)
+		}
```
gordon-klotho, line 130: This check (source != target) should be checked in the DAG code probably.

pkg/provider/aws/resources/cloud_map.go
```diff
@@ -19,12 +19,25 @@ type (
...
+
+func (namespace *PrivateDnsNamespace) Create(dag *core.ResourceGraph, params PrivateDnsNamespaceCreateParams) error {
+	namespace.Name = privateDnsNamespaceSanitizer.Apply(params.AppName)
+	namespace.ConstructsRef = params.Refs
+
+	existingNamespace, found := core.GetResource[*PrivateDnsNamespace](dag, namespace.Id())
+	if found {
+		existingNamespace.ConstructsRef.AddAll(params.Refs)
+	} else {
+		err := dag.CreateDependencies(namespace, map[string]any{"Vpc": params})
```
gordon-klotho, line 35: Are `params` really a valid configuration for VPC?

pkg/provider/aws/resources/ecr.go
```diff
@@ -37,7 +37,7 @@ type RepoCreateParams struct {
 
 func (repo *EcrRepository) Create(dag *core.ResourceGraph, params RepoCreateParams) error {
 	repo.Name = params.AppName
-	repo.ConstructsRef = params.Refs
+	repo.ConstructsRef = params.Refs.Clone()
```
gordon-klotho, line 40: We should decide whether cloning happens in `Create` or caller-side. There's reasonable arguments for either, so we should pick one and standardize.

pkg/provider/aws/resources/eks.go
```diff
@@ -628,24 +627,25 @@ func (cluster *EksCluster) InstallAlbController(references core.AnnotationKeySet
...
+		AppName:     appName,
+		ClusterName: strings.TrimLeft(cluster.Name, fmt.Sprintf("%s-", appName)),
+		Refs:        role.ConstructsRef.Clone(),
+	})
 	var aRef core.AnnotationKey
 	for ref := range references {
 		aRef = ref
 		break
 	}
+	dag.AddDependency(role, oidc)
```
gordon-klotho, line 647: I'm confused why the role would need to depend on these - doesn't seem like it references it.

gordon-klotho (commented):
Mostly good - not really anything blocking except I had a fair number of questions/comments so holding off on the approval just based on volume.

pkg/knowledge_base/edge_kb.go
```diff
@@ -58,6 +58,9 @@ type (
 		Destination core.Resource
 		// Routes refers to any api routes being satisfied by the edge
 		Routes []core.Route
+		// SourceRef denotes the source annotation being used during expansion or configuration
+		// This is a temporary field due to helm chart being the lowest level of kubernetes resource at the moment
```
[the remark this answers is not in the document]
jhsinger-klotho: because multiple exec units can exist within a helm chart so the .Source.Refs doesnt really give us the information as to what we are connecting when we get to the dependency here. Realistically we would want the dependency to be to the manifest within the helm chart so we know what exec unit its tied to

pkg/provider/aws/execution_unit.go
```diff
@@ -196,7 +197,7 @@ func (a *AWS) getNodeGroupConfiguration(result *core.ConstructGraph, dag *core.R
 		construct := result.GetConstruct(ref.ToId())
 		unit, ok := construct.(*core.ExecutionUnit)
 		if !ok {
-			return nodeGroupConfig, fmt.Errorf("node group must only have construct references to ExecutionUnits but got %T", construct)
+			continue
```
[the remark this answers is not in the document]
jhsinger-klotho: we want all the refs to be processed used so we can figure out the DiskSize required

pkg/provider/aws/knowledgebase/eks.go
```diff
@@ -86,9 +103,123 @@ var EksKB = knowledgebase.Build(
...
+	},
+	knowledgebase.EdgeBuilder[*kubernetes.HelmChart, *resources.RdsProxy]{
+		Configure: func(chart *kubernetes.HelmChart, proxy *resources.RdsProxy, dag *core.ResourceGraph, data knowledgebase.EdgeData) error {
+			role := GetRoleForUnit(chart, data.SourceRef)
+			if role == nil {
+				return fmt.Errorf("no role found for chart %s and source reference %s in HelmChart to ddb RdsProxy expansion", chart.Id(), data.SourceRef.ToId())
+			}
+			upstreamResources := dag.GetUpstreamResources(proxy)
+			for _, res := range upstreamResources {
+				if tg, ok := res.(*resources.RdsProxyTargetGroup); ok {
```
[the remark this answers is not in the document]
jhsinger-klotho: can change

pkg/provider/aws/knowledgebase/eks.go
```diff
@@ -86,9 +103,123 @@ var EksKB = knowledgebase.Build(
...
 )
+
+func GetRoleForUnit(chart *kubernetes.HelmChart, ref core.AnnotationKey) *resources.IamRole {
+	for _, val := range chart.Values {
+		if iacVal, ok := val.(core.IaCValue); ok {
+			if role, ok := iacVal.Resource.(*resources.IamRole); ok {
+				singleRef, ok := role.ConstructsRef.GetSingle()
+				if ok && singleRef == ref {
+					return role
+				}
```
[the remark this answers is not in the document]
jhsinger-klotho: here we also want to make sure that the role has a single ref i believe otherwise were not positive that its the right role, because something like the node group role could have many refs

pkg/provider/aws/knowledgebase/iam.go
jhsinger-klotho: yeah so this is another case of not having the helm chart broken down, realistically we would want to be able to go directly to the service account name in here, which we cant do today from our kubernetes setup. Definitely will need to revisit this

pkg/provider/aws/plugin.go
jhsinger-klotho: right now they are since its just the payload bucket. Open to other suggestions just didnt want to make things too complex

pkg/provider/aws/plugin.go
jhsinger-klotho: sorry, because we dont understand that helm chart and lambda are both compute resources is closer to what it means. Our KB should (and hopefully will soon) have the knowledge that it doesnt need to put compute between compute and data but right now has no idea because we dont classify any resource types so lambda -> helm chart -> rds is a valid path to it

pkg/provider/aws/plugin.go
jhsinger-klotho: makes sense, ill move it over

pkg/provider/aws/resources/eks.go
jhsinger-klotho: the roles assume role policy can reference it when its tied to a service account. Realistically we will want to pass in the service account name as edge data as well, we likely need to figure out how we want to structure edge data because that will address one of the comments above

pkg/provider/aws/plugin.go
gordon-klotho: Also, that seems like an AWS-specific detail (the payloads bucket). Is there a reason we don't just create the S3 bucket as part of this plugin? Doesn't make sense as a construct IMO. That said, may be beyond the scope of this change if it's too much, so feel free to disregard (maybe ticket for it though).

pkg/provider/aws/resources/cloud_map.go
gordon-klotho: Sure, or a comment - something so we don't get tripped up about it later.

pkg/provider/aws/resources/eks.go
gordon-klotho: I'd expect that dependency to be added when the reference is added - where does that happen and is there a reason the dependency couldn't be added there?

pkg/provider/aws/knowledgebase/eks.go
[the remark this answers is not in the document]
jhsinger-klotho: you mean something like 
```
func GetRoleForUnit(chart *kubernetes.HelmChart, ref core.AnnotationKey) *resources.IamRole {
	for key, val := range chart.Values {
		if kubernetes.GenerateRoleArnPlaceholder(ref.ID) == key {
			if iacVal, ok := val.(core.IaCValue); ok {
				if role, ok := iacVal.Resource.(*resources.IamRole); ok {
					return role
				}
			}
		}
	}
	return nil
}
```

pkg/provider/aws/knowledgebase/eks.go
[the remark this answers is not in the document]
jhsinger-klotho: a little confused actually what the change would be

pkg/provider/aws/plugin.go
jhsinger-klotho: im a little confused on the last message. Agreed we need to work on the internal payloads bucket, but what exactly do we want the ticket to be about?

pkg/provider/aws/knowledgebase/eks.go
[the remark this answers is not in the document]
gordon-klotho: Yeah, something like that.

pkg/provider/aws/knowledgebase/eks.go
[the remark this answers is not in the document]
gordon-klotho:
```go
tg, ok := res.(*resources.RdsProxyTargetGroup)
if !ok {
    continue
}
// the rest, de-indented now
```

pkg/provider/aws/plugin.go
jhsinger-klotho: yeah i think the next designs which we will hopefully be picking up next week will address this

pkg/provider/aws/resources/eks.go
jhsinger-klotho: this is where that all happens. when we install the ALB controller we know we need to create a service account and a role. In order to tie the 2 together we have to have the role depend on the oidc

jhsinger-klotho: this will be able to be cleaned up long term when we dont only work in terms of helm charts and manifests in our KB

pkg/provider/aws/knowledgebase/eks.go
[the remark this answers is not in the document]
jhsinger-klotho: 👍 will modify to this then

pkg/provider/aws/knowledgebase/lambda.go
```diff
@@ -153,6 +154,74 @@ var LambdaKB = knowledgebase.Build(
 			return nil
 		},
 	},
+	knowledgebase.EdgeBuilder[*resources.LambdaFunction, *kubernetes.HelmChart]{
+		Expand: func(lambda *resources.LambdaFunction, destination *kubernetes.HelmChart, dag *core.ResourceGraph, data knowledgebase.EdgeData) error {
+			if len(lambda.Subnets) == 0 {
+				lambda.Subnets = make([]*resources.Subnet, 2)
```
gordon-klotho, line 160: I'm still confused why we need this when `len(subparams["Subnets"]) == 2` - I'm sure I'll forget it at some point if it's required.

pkg/provider/aws/knowledgebase/lambda.go
```diff
@@ -153,6 +154,74 @@ var LambdaKB = knowledgebase.Build(
...
+					}
+				}
+				err := dag.CreateDependencies(lambda, subparams)
+				if err != nil {
+					return err
+				}
+			}
+			refs := core.AnnotationKeySet{}
+			refs.AddAll(lambda.ConstructsRef)
+			refs.AddAll(destination.ConstructRefs)
```
gordon-klotho, line 193: nit: `CloneWith`

pkg/provider/aws/plugin.go
```diff
@@ -74,23 +74,54 @@ func (a *AWS) copyConstructEdgeToDag(
 	dep graph.Edge[core.Construct],
 	dag *core.ResourceGraph) error {
 	data := knowledgebase.EdgeData{AppName: a.Config.AppName, Source: sourceResource, Destination: destinationResource}
+
 	switch construct := dep.Source.(type) {
 	case *core.ExecutionUnit:
-		switch dep.Destination.(type) {
+		if a.Config.GetExecutionUnit(construct.ID).Type == kubernetes.KubernetesType {
+			data.SourceRef = construct.AnnotationKey
+		}
```
gordon-klotho, line 82: Is the reason this is only set for k8s because it's only used for that scenario? If we really want to make this field be hyper-specific, IMO the name should reflect that.

pkg/provider/aws/plugin.go
```diff
@@ -74,23 +74,54 @@ func (a *AWS) copyConstructEdgeToDag(
...
+		switch dest := dep.Destination.(type) {
 		case *core.Orm:
 			data.Constraint = knowledgebase.EdgeConstraint{
 				NodeMustExist: []core.Resource{&resources.RdsProxy{}},
 			}
+			// Because we dont understand resource types, our edges cant depict what is truly valid and what isnt yet. Example
+			// 	found multiple paths which satisfy constraints for edge aws:lambda_function:lambda -> aws:rds_instance:rds.
+			// Paths: [[{*resources.LambdaFunction *resources.RdsProxy} {*resources.RdsProxyTargetGroup *resources.RdsProxy} {*resources.RdsProxyTargetGroup *resources.RdsInstance}]
+			// [{*resources.LambdaFunction *resources.RdsProxy} {*kubernetes.HelmChart *resources.RdsProxy} {*kubernetes.HelmChart *resources.RdsInstance}]]
+			// When we can classify lambda and helm as compute then we can understand that there is no need for the data flow in this manner
```
gordon-klotho, line 97: Yeah, this (and the constraint in Kv) is awkward. Do we have a ticket/task for designing the fix for it? Until then, even time we add an edge, it could have unintended consequences by adding a new path between otherwise unrelated things.

pkg/provider/aws/resources/ecr.go
gordon-klotho: Not necessarily memory allocation, but just for standardization and consistency - even if that standard is "clone on both sides to be sure".

pkg/provider/aws/resources/eks.go
gordon-klotho: Ah, the connection happens in `[*resources.IamRole, *resources.OpenIdConnectProvider].Configure` where it sets the policy doc. It's a bit odd to me that they're so separate - hard to track down.

pkg/provider/aws/knowledgebase/eks.go
```diff
@@ -86,9 +103,125 @@ var EksKB = knowledgebase.Build(
...
+			for _, env := range data.EnvironmentVariables {
+				addEnvVarToChart(chart, proxy, env)
+			}
+			return nil
+		},
+		ValidDestinations: []core.Resource{&resources.RdsInstance{}},
+	},
 )
+
+func GetRoleForUnit(chart *kubernetes.HelmChart, ref core.AnnotationKey) *resources.IamRole {
```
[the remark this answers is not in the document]
jhsinger-klotho: getting a role for a unit is aws specific and really something that we shouldnt have to do but is a byproduct of not understanding whats within the helm chart, i can rename but id prefer to not move it at the moment

pkg/provider/aws/knowledgebase/iam.go
```diff
@@ -171,6 +175,18 @@ var IamKB = knowledgebase.Build(
 			return nil
 		},
 	},
+	knowledgebase.EdgeBuilder[*resources.IamPolicy, *resources.PrivateDnsNamespace]{
+		Configure: func(policy *resources.IamPolicy, namespace *resources.PrivateDnsNamespace, dag *core.ResourceGraph, data knowledgebase.EdgeData) error {
+			if policy.Policy == nil {
+				policy.Policy = resources.CreateAllowPolicyDocument([]string{"servicediscovery:DiscoverInstances"}, []core.IaCValue{{Property: core.ALL_RESOURCES_IAC_VALUE}})
+				return nil
+			}
+			statement := resources.CreateAllowPolicyDocument([]string{"servicediscovery:DiscoverInstances"}, []core.IaCValue{{Property: core.ALL_RESOURCES_IAC_VALUE}}).Statement
```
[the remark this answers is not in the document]
jhsinger-klotho: the dupes dont hurt outside of character count here, but we do have a deduplicate method for statements. Ill make a helper function for this so that we can reduce footprint

pkg/provider/aws/plugin.go
jhsinger-klotho: right now its only set for k8s because thats all that needs it, however there could be other scenarios in which we need it in the near term, we just havent run into those.

The hope is that we dont though i just named it generic in case

pkg/provider/aws/plugin.go
jhsinger-klotho: this should fall into the design task for high level intents that yuval is working on, if not ill cut one and have some ideas about what we would need to do here

pkg/provider/aws/knowledgebase/eks.go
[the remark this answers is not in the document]
gordon-klotho: That's fine if it's AWS-specific that it wouldn't go in HelmChart. But the name makes it seem like it's a general purpose function (ie, that its signature might be `GetRoleForUnit(ref core.AnnotationKey) *resources.IamRole`).

pkg/provider/aws/knowledgebase/lambda.go
gordon-klotho: [`MakeSlice`](https://pkg.go.dev/reflect#MakeSlice) is probably what you're looking for

pkg/provider/aws/plugin.go
gordon-klotho: Maybe down the line, but this problem is more knowledge-base implementation specific that I'm not confident will be covered by that.

PR #647, opened 2023-05-26
CDN for S3 and Expose

### LINKED ISSUE #607: Static Unit and Cloudfront Expansion

### LINKED ISSUE #653: revisit StageName for Expose + Cloudfront
Similar to https://github.com/klothoplatform/klotho/issues/652, we currently hard-code the stage name to `"stage"`, because edge expansion happens before node configure (and the stage name only gets set in configure).

yuval-klotho:
- `ts-static-unit` works with CDN now.
- `go-expose` works with CDN

Resolves #607 and resolves #653. ~~See #653 for an outstanding issue.~~

### Standard checks

- **Unit tests**: updated
- **Docs**: n/a


pkg/provider/aws/knowledgebase/cloudfront.go
```diff
@@ -0,0 +1,112 @@
...
+		},
+	}
+	origin := &resources.CloudfrontOrigin{
+		S3OriginConfig: s3OriginConfig,
+		DomainName: core.IaCValue{
+			Resource: conn.bucket,
+			Property: resources.BUCKET_REGIONAL_DOMAIN_NAME_IAC_VALUE,
+		},
+		OriginId: conn.construct.ID,
+	}
```
jhsinger-klotho, line 73: just a helper method right? If so i like that, it will really only ever be used in one place but thats ok

pkg/provider/aws/resources/cloudfront.go
```diff
@@ -82,56 +83,26 @@ type (
...
-	s3OriginConfig := S3OriginConfig{
-		OriginAccessIdentity: core.IaCValue{
-			Resource: oai,
-			Property: CLOUDFRONT_ACCESS_IDENTITY_PATH_IAC_VALUE,
-		},
+func (oai *OriginAccessIdentity) Create(dag *core.ResourceGraph, params OriginAccessIdentityCreateParams) error {
+	oai.Name = params.Name
+	oai.ConstructsRef = params.Refs
+	if dag.GetResource(oai.Id()) != nil {
+		return fmt.Errorf(`Origin Access Identity with name "%s" already exists`, oai.Name)
```
jhsinger-klotho, line 95: do we want to fail here? What if someone had 2 cdns with the same bucket? I know we dont support that now but just a what if

pkg/provider/aws/resources/cloudfront.go
```diff
@@ -176,6 +147,41 @@ func NewCloudfrontDistribution(appName string, cdnId string) *CloudfrontDistribu
...
+			Cookies:     Cookies{Forward: "none"},
+		},
+		MinTtl:               0,
+		DefaultTtl:           3600,
+		MaxTtl:               86400,
+		ViewerProtocolPolicy: "allow-all",
+	}
+	distro.Restrictions = &Restrictions{
+		GeoRestriction: GeoRestriction{RestrictionType: "none"},
+	}
```
jhsinger-klotho, line 180: i would put all these configuration like things in the Configure function just so we can expose it in the future

jhsinger-klotho (commented):
looks very much on track, only real comment thats maybe blocking is just the separation of create and configure for the CDN distro. other than that looks good to me!

pkg/provider/aws/knowledgebase/cloudfront.go
yuval-klotho: Yup, exactly. It's mainly to keep the code clean and self-documenting, since it's doing a few things.

pkg/provider/aws/resources/cloudfront.go
yuval-klotho: It ended up being non-trivial to do that, so I filed a bug and am going to keep things as-is: https://github.com/klothoplatform/klotho/issues/652

jhsinger-klotho: im a little confused as to why its non-trivial. In the edge couldnt you just set the specific thing on the DefaultCacheBehavior? Then configure will set the remainder of the fields?

pkg/provider/aws/resources/cloudfront.go
jhsinger-klotho: the oai is bucket level so it should be reuseable across cdns if needed, thus not needing a new id. We dont support a use case that needs it yet so not blocking

yuval-klotho:
Note: I just merged in a couple extra commits that resolve #653, which I had originally filed as a followup to this ticket. Turns out the fix was to just sidestep the issue entirely (thanks, @gordon-klotho!)

pkg/provider/aws/knowledgebase/cloudfront.go
```diff
@@ -0,0 +1,156 @@
...
+			default:
+				var ids []string
+				for _, cons := range collectionutil.Keys(stage.ConstructsRef) {
+					ids = append(ids, cons.ID)
+				}
+				sort.Strings(ids)
+				return errors.Errorf(
+					`couldn't determine the id of the construct that created API stage "%s": expected just one construct, but found [%s]`,
+					stage.Id(),
+					strings.Join(ids, ", "))
```
gordon-klotho, line 60:
```suggestion
				return errors.Errorf(
					`couldn't determine the id of the construct that created API stage "%s": expected just one construct, but found %v`,
					stage.Id(),
					ids)
```
nit: %v will do the joining and bracketing for you for arrays

pkg/provider/aws/resources/cloudfront.go
```diff
@@ -82,98 +84,57 @@ type (
...
+func (oai *OriginAccessIdentity) Create(dag *core.ResourceGraph, params OriginAccessIdentityCreateParams) error {
+	oai.Name = params.Name
+	oai.ConstructsRef = params.Refs
+	if dag.GetResource(oai.Id()) != nil {
+		return fmt.Errorf(`an Origin Access Identity with name "%s" already exists`, oai.Name)
 	}
-	distribution.Origins = append(distribution.Origins, origin)
-	distribution.DefaultCacheBehavior.TargetOriginId = origin.OriginId
+
+	// This is technically a config, but it's always just this value, so it's fine (and convenient) to inline it here
```
gordon-klotho, line 99: What's inlined?

pkg/provider/aws/resources/cloudfront.go
```diff
@@ -82,98 +84,57 @@ type (
...
+func (distro *CloudfrontDistribution) Create(dag *core.ResourceGraph, params CloudfrontDistributionCreateParams) error {
+	distro.Name = cloudfrontDistributionSanitizer.Apply(fmt.Sprintf("%s-%s", params.AppName, params.CdnId))
+	distro.ConstructsRef = params.Refs
+
+	if dag.GetResource(distro.Id()) != nil {
+		return errors.Errorf(`duplicate Cloudfront distribution "%s" (internal error)`, distro.Id())
+	}
+	dag.AddResource(distro)
+
+	// Some defaults. Someday these may move to a Configure (and be configurable), but not yet.
```
gordon-klotho, line 120: Can you add _why_ they're not moved to Configure yet?

pkg/provider/aws/static_unit.go
```diff
@@ -29,15 +28,30 @@ func (a *AWS) expandStaticUnit(dag *core.ResourceGraph, unit *core.StaticUnit) e
...
+				Refs:    core.AnnotationKeySetOf(unit.AnnotationKey),
+			})
+			if err != nil {
+				return err
+			}
+			dag.AddDependency(distro, bucket)
+		}
+		return nil
+	default:
+		return errors.Errorf(`Found too many buckets for unit %s. This is an internal error.`, unit.Id())
```
gordon-klotho, line 55:
```suggestion
		var bucketIds []core.ResourceId
		for k := range createdBuckets {
			bucketIds = append(bucketIds, k)
		}
		return errors.Errorf(`Found too many buckets %v for unit %s. This is an internal error.`, bucketIds, unit.Id())
```
nit: just a little more info/context on the error

gordon-klotho (approved):
Only minor comments, nothing that couldn't be fixed in a fast-follow or later.

pkg/provider/aws/resources/cloudfront.go
jhsinger-klotho: this will be the same in expose also, the default cache would be nil. I would just add a method that sets the defaultTargetId for the cdn, if the default cache is nil create and set the field. If its not nil, just override the field.

Then in configure we would take have to just not set that field in the default cache behavior because we know its derived by edge expansion

jhsinger-klotho (approved):

PR #654, opened 2023-05-30
add post-compile hooks

yuval-klotho:
These are on a file extension basis. They let you do something like:

    klotho --set-option output.post-write-hooks.ts='npx prettier -w {}'

### Initially

~Out of the box, we don't~ Previously, we wouldn't do any formatting.

```
$ cat ~/.klotho/options.yaml
ui:
    disable-logo: true
```

```
$ klotho -c klotho.yaml
Adding resource internal:InternalKlothoPayloads
Adding resource execution_unit:main
searching for http imports in main.go
main.go: Found 2 route(s) on app 'r'
main.go: Found 3 route(s) on mounted router 'routes.ExtraRoutes'
Adding resource expose:app
```

The output is pretty ugly:

```typescript
// $ cat compiled/index.ts| tail -n 15
cachedMethods: [`HEAD`,`GET`],
targetOriginId: `app`,
forwardedValues: {queryString: true,
cookies: {forward: `none`,
},
},
defaultTtl: 3600,
maxTtl: 86400,
viewerProtocolPolicy: `allow-all`,
},
        restrictions: {geoRestriction: {restrictionType: `none`,
},
},
        defaultRootObject: ``,
    })
```

### Add a formatter

Now, we do apply formatting. Optionally, you can use `--set-option` to change it:

```
$ klotho --set-option output.post-write-hooks.ts='npx prettier -w {}'
$ cat ~/.klotho/options.yaml
ui:
    disable-logo: true
output:
    post-write-hooks:
        ts: npx prettier -w {}
```

Now, when we compile, we get a notification about the hook:

```
klotho -c klotho.yaml
Adding resource internal:InternalKlothoPayloads
Adding resource execution_unit:main
searching for http imports in main.go
main.go: Found 2 route(s) on app 'r'
main.go: Found 3 route(s) on mounted router 'routes.ExtraRoutes'
Adding resource expose:app
index.ts: running post-output hook: npx prettier -w index.ts        # <--- HERE
```

And the output is a lot nicer:

```typescript
// $ cat compiled/index.ts| tail -n 15
        `OPTIONS`,
        `PATCH`,
        `POST`,
        `PUT`,
      ],
      cachedMethods: [`HEAD`, `GET`],
      targetOriginId: `app`,
      forwardedValues: { queryString: true, cookies: { forward: `none` } },
      defaultTtl: 3600,
      maxTtl: 86400,
      viewerProtocolPolicy: `allow-all`,
    },
    restrictions: { geoRestriction: { restrictionType: `none` } },
    defaultRootObject: ``,
  });
```

### Negative path case

```
$ klotho --set-option output.post-write-hooks.ts='npx yuvalisblah -w {}'

$ klotho -c klotho.yaml
Adding resource internal:InternalKlothoPayloads
Adding resource execution_unit:main
searching for http imports in main.go
main.go: Found 2 route(s) on app 'r'
main.go: Found 3 route(s) on mounted router 'routes.ExtraRoutes'
Adding resource expose:app
index.ts: running post-output hook: npx yuvalisblah -w index.ts
index.ts: failed to apply post-output hook to index.ts: exit status 1

$ echo $?
0
```

The failure comes out as a warning, and does not cause the compilation to fail.

## Standard checks

- **Unit tests**: not added
- **Docs**: not documented

### DIFF
--- a/pkg/compiler/compilation_document.go
+++ b/pkg/compiler/compilation_document.go
@@ -2,12 +2,15 @@ package compiler
 
 import (
 	"encoding/json"
+	"fmt"
 	"os"
 	"os/exec"
 	"path"
 	"path/filepath"
+	"regexp"
 	"strings"
 
+	"github.com/google/shlex"
 	"github.com/klothoplatform/klotho/pkg/config"
 	"github.com/klothoplatform/klotho/pkg/core"
 	"github.com/klothoplatform/klotho/pkg/logging"
@@ -31,6 +34,8 @@ type (
 		PostWriteHooks map[string]string `yaml:"post-write-hooks,omitempty"`
 	}
 )
+
+var unquotedCharsRe = regexp.MustCompile(`^[\w.{}:>=<@/-]*$`)
 
 func (doc *CompilationDocument) OutputTo(dest string) error {
 	errs := make(chan error)
@@ -63,8 +68,8 @@ func (doc *CompilationDocument) OutputTo(dest string) error {
 			file.Close()
 
 			fileExt := filepath.Ext(path)
-			if strings.HasPrefix(fileExt, ".") {
-				fileExt = fileExt[1:]
+			if fileExt != "" {
+				fileExt = strings.TrimPrefix(fileExt, ".")
 				if hook, found := doc.OutputOptions.PostWriteHooks[fileExt]; found {
 					log := zap.S().With(logging.FileField(f))
 					hookErr := postCompileHook(dest, f, hook, log)
@@ -91,7 +96,10 @@ func (doc *CompilationDocument) OutputTo(dest string) error {
 }
 
 func postCompileHook(dir string, file core.File, hook string, log *zap.SugaredLogger) error {
-	hookSegments := strings.Split(hook, " ")
+	hookSegments, err := shlex.Split(hook)
+	if err != nil {
+		return err
+	}
 	if len(hookSegments) == 0 {
 		return errors.New(`empty formatter command`)
 	}
@@ -104,8 +112,16 @@ func postCompileHook(dir string, file core.File, hook string, log *zap.SugaredLo
 	}
 
 	cmd := exec.Command(hookSegments[0], args...)
-	log.Infof(`running post-output hook: %s`, strings.Join(cmd.Args, " "))
 	cmd.Dir = dir
+
+	quotedArgs := cmd.Args
+	for i, arg := range quotedArgs {
+		if !unquotedCharsRe.MatchString(arg) {
+			quotedArgs[i] = fmt.Sprintf(`%#v`, arg)
+		}
+	}
+	log.Infof(`running post-output hook: %s`, strings.Join(quotedArgs, " "))
+
 	return cmd.Run()
 }
 
--- a/pkg/compiler/compilation_document.go
+++ b/pkg/compiler/compilation_document.go
@@ -3,11 +3,14 @@ package compiler
 import (
 	"encoding/json"
 	"os"
+	"os/exec"
 	"path"
 	"path/filepath"
+	"strings"
 
 	"github.com/klothoplatform/klotho/pkg/config"
 	"github.com/klothoplatform/klotho/pkg/core"
+	"github.com/klothoplatform/klotho/pkg/logging"
 	"github.com/klothoplatform/klotho/pkg/multierr"
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
@@ -21,6 +24,11 @@ type (
 		Configuration    *config.Application
 		Resources        *core.ResourceGraph
 		OutputFiles      []core.File
+		OutputOptions    OutputOptions
+	}
+
+	OutputOptions struct {
+		PostWriteHooks map[string]string `yaml:"post-write-hooks,omitempty"`
 	}
 )
 
@@ -53,6 +61,22 @@ func (doc *CompilationDocument) OutputTo(dest string) error {
 			}
 			_, err = f.WriteTo(file)
 			file.Close()
+
+			fileExt := filepath.Ext(path)
+			if strings.HasPrefix(fileExt, ".") {
+				fileExt = fileExt[1:]
+				if hook, found := doc.OutputOptions.PostWriteHooks[fileExt]; found {
+					log := zap.S().With(logging.FileField(f))
+					hookErr := postCompileHook(dest, f, hook, log)
+					if hookErr != nil {
+						zap.S().With(zap.Error(hookErr), logging.FileField(f)).Warnf(
+							`failed to apply post-output hook to %s: %s`,
+							f.Path(),
+							hookErr.Error())
+					}
+				}
+			}
+
 			errs <- err
 		}(files[idx])
 	}
@@ -64,6 +88,25 @@ func (doc *CompilationDocument) OutputTo(dest string) error {
 		}
 	}
 	return nil
+}
+
+func postCompileHook(dir string, file core.File, hook string, log *zap.SugaredLogger) error {
+	hookSegments := strings.Split(hook, " ")
+	if len(hookSegments) == 0 {
+		return errors.New(`empty formatter command`)
+	}
+	var args []string
+	for _, arg := range hookSegments[1:] {
+		if arg == "{}" {
+			arg = file.Path()
+		}
+		args = append(args, arg)
+	}
+
+	cmd := exec.Command(hookSegments[0], args...)
+	log.Infof(`running post-output hook: %s`, strings.Join(cmd.Args, " "))
+	cmd.Dir = dir
+	return cmd.Run()
 }
 
 func (document *CompilationDocument) OutputResources() (resourceCounts map[string]int, err error) {
--- a/pkg/compiler/compilation_document.go
+++ b/pkg/compiler/compilation_document.go
@@ -2,12 +2,12 @@ package compiler
 
 import (
 	"encoding/json"
-	"fmt"
 	"os"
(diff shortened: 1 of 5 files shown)


pkg/compiler/compilation_document.go
```diff
@@ -66,6 +90,25 @@ func (doc *CompilationDocument) OutputTo(dest string) error {
 	return nil
 }
 
+func postCompileHook(dir string, file core.File, hook string, log *zap.SugaredLogger) error {
+	hookSegments := strings.Split(hook, " ")
```
gordon-klotho, line 94: Consider using [shlex](https://pkg.go.dev/github.com/google/shlex) to handle things like quotations.

pkg/compiler/compilation_document.go
```diff
@@ -66,6 +90,25 @@ func (doc *CompilationDocument) OutputTo(dest string) error {
...
+	var args []string
+	for _, arg := range hookSegments[1:] {
+		if arg == "{}" {
+			arg = file.Path()
+		}
+		args = append(args, arg)
+	}
+
+	cmd := exec.Command(hookSegments[0], args...)
+	log.Infof(`running post-output hook: %s`, strings.Join(cmd.Args, " "))
```
gordon-klotho, line 107: nit: edge-case, but might want to quote all the args if the intention is that the command is copy/paste-able for reproducing. This also might be more of a Debug log if it can run over user files (see general comment `(1)`).

pkg/compiler/compilation_document.go
```diff
@@ -53,6 +61,22 @@ func (doc *CompilationDocument) OutputTo(dest string) error {
 			}
 			_, err = f.WriteTo(file)
 			file.Close()
+
+			fileExt := filepath.Ext(path)
+			if strings.HasPrefix(fileExt, ".") {
```
gordon-klotho, line 66: nit: `fileExt` will always have a "." or be empty

gordon-klotho (commented):
I have some reservations about running a formatter over users' code(1), even if it's opt-in - it may degrade their ability to recognize the output. I also worry that this will cause us to not prioritize making sure the output files are formatted nicely for users without a post-compile formatter.

High-level comments aside, the code itself is good - nothing blocking.


(1) I know the CompilationDocument.OutputFiles are only IaC files at the moment, but the intention is/was to move away from exec units from having their own OutputTo and instead be handled at IaC layer.

pkg/compiler/compilation_document.go
yuval-klotho: Ah cool, thanks for that library. That's a good call, because then it also lets people pass in a command of `bash -c 'whatever {} && whatever-else {}'` if they want

pkg/compiler/compilation_document.go
yuval-klotho: That wasn't my intention, but it's a good idea.

pkg/compiler/compilation_document.go
yuval-klotho: As in, I could have just done `if fileExt != ""`? If so, yeah — but I find the fact that it starts with `"."` non-obvious (I have to re-look-it-up each time), so if I went down that path I would have gone with:

```go
if fileExt != "" {
	fileExt = fileExt[1:] // if Ext returns non-empty, it's always prefixed by "."
	...
```

and by then it feels a bit potato-potato to me.

gordon-klotho: I was thinking more along the lines of:
```go
if fileExt != "" {
    fileExt = strings.TrimPrefix(fileExt, ".")
```
note that TrimPrefix will no-op if the prefix isn't found.

either way, just an "nit" so feel free to disregard

pkg/compiler/compilation_document.go
```diff
@@ -104,8 +112,16 @@ func postCompileHook(dir string, file core.File, hook string, log *zap.SugaredLo
 	}
 
 	cmd := exec.Command(hookSegments[0], args...)
-	log.Infof(`running post-output hook: %s`, strings.Join(cmd.Args, " "))
 	cmd.Dir = dir
+
+	quotedArgs := cmd.Args
+	for i, arg := range quotedArgs {
+		if !unquotedCharsRe.MatchString(arg) {
+			quotedArgs[i] = fmt.Sprintf(`%#v`, arg)
```
gordon-klotho, line 120: What does `%#v` do with strings? I'd expect something like `%q` here, or just using `strconv.Quote`.

yuval-klotho: > %#v	a Go-syntax representation of the value

`%q` also works, I just didn't read that far down the docs. ;-) `strconv.Quote` is also fine, I just didn't know about it.

pkg/compiler/compilation_document.go
```diff
@@ -104,8 +112,16 @@ func postCompileHook(dir string, file core.File, hook string, log *zap.SugaredLo
 	}
 
 	cmd := exec.Command(hookSegments[0], args...)
-	log.Infof(`running post-output hook: %s`, strings.Join(cmd.Args, " "))
 	cmd.Dir = dir
+
+	quotedArgs := cmd.Args
+	for i, arg := range quotedArgs {
+		if !unquotedCharsRe.MatchString(arg) {
```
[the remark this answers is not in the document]
yuval-klotho: This is user-visible at INFO log level, so imo it's decently important to have it look nice.

Instead of:

```
running post-output hook: npx prettier -w index.ts
```

it would be:

```
running post-output hook: "npx" "prettier" "-w" "index.ts"
```

... which I find harder to read, by enough to warrant the regex.

I'm a fan of xkcd as much as the next guy, but I don't think regex is really _that_ scary, especially at the level of complexity of this one.

yuval-klotho:
Sorry, I missed this high-level comment above:

> I have some reservations about running a formatter over users' code(1), even if it's opt-in - it may degrade their ability to 
recognize the output.

Speaking for myself it, it makes it a lot _easier_ to read the output.

> I also worry that this will cause us to not prioritize making sure the output files are formatted nicely for users without a post-compile formatter.

I don't see that formatting as a core value-add for Klotho, so personally I'm very happy to offload it. Sure, it'd be even nicer if it were done totally "in house" in the Go code; but isn't that letting the perfect be the enemy of the good?
 
> (1) I know the CompilationDocument.OutputFiles are only IaC files at the moment, but the intention is/was to move away from exec units from having their own OutputTo and instead be handled at IaC layer.

If/when this happens, we can remove or rework this undocumented feature.

---

Overall, I think this makes Klotho easier to use for me. But there's not consensus on that, I can close the PR. Lmk.

yuval-klotho:
It just occurred to me:

> I have some reservations about running a formatter over users' code(1), even if it's opt-in - it may degrade their ability to recognize the output.

Is your concern that this would mangle the user's `.ts` files that they wrote?

This only affects our generated files. The user's files are in a separate struct (the `core.ExecutionUnit`'s `.Files`), and are unaffected by this option. Those are written in [`(*core.ExecutionUnit).OutputTo(string)`](https://github.com/klothoplatform/klotho/blob/main/pkg/core/exec_unit.go#L86)

gordon-klotho:
> Speaking for myself it, it makes it a lot _easier_ to read the output. If a user doesn't feel that's true for them, then doesn't the opt-in nature of this provide a lot of protection?

Yeah, I just worry about the signal of "if you want it to look nice, you have to go install XYZ and go out of your way to configure it like ABC".
 
> I don't see that formatting as a core value-add for Klotho, so personally I'm very happy to offload it. Sure, it'd be even nicer if it were done totally "in house" in the Go code; but isn't that letting the perfect be the enemy of the good?

I agree that we should offload what we can, but I see formatting, and other UX, as part of a core value-add.

> Is your concern that this would mangle the user's .ts files that they wrote (or .js files, if they added an option for that)?
> 
> This only affects our generated files. The user's files are in a separate struct (the core.ExecutionUnit's .Files), and are unaffected by this option. Those are written in [(*core.ExecutionUnit).OutputTo(string)](https://github.com/klothoplatform/klotho/blob/main/pkg/core/exec_unit.go#L86)

Yeah, that's my concern and what the footnote (1) is covering - that the intention of OutputFiles was to be all output files, and the existence of ExecutionUnit.OutputTo is a carryover from the previous design. We can revisit the desire to move to OutputFiles, given that we haven't spent any effort in standardizing, but I just won't want to accidentally change this behaviour in the future, and it'd be hard to catch.

The strongest concern and the closest (but not quite qualifying, to be clear) of a blocker is the UX around "you need to install <yet another tool> for Klotho to be useful". We've already heard concerns about Docker and pulumi. It's better than nothing, but if we could spend like 2 person-hours on fixing it in the generation code/template, I'm sure we could get it to an acceptable state.

yuval-klotho:
Okay, I will close this.

pkg/compiler/compilation_document.go
```diff
@@ -21,10 +28,21 @@ type (
...
+		PostWriteHooks map[string]string `yaml:"post-write-hooks,omitempty"`
 	}
 )
 
+var unquotedCharsRe = regexp.MustCompile(`^[\w.{}:>=<@/-]*$`)
+
 func (doc *CompilationDocument) OutputTo(dest string) error {
+
+	postWriteHooks := newDefaultPostWriteHooksMap()
+	collectionutil.Extend(doc.OutputOptions.PostWriteHooks).Into(postWriteHooks)
```
gordon-klotho, line 44: nit (opinion): Even though this is shorter, I don't think it's as clear as written out (and it's only 3 lines and probably the same number of characters overall).

gordon-klotho (approved):

PR #655, opened 2023-05-30
Improves s3 bucket sanitization

### LINKED ISSUE #641: camelCase and dash-case conversion errors in index.ts
When I ran `ts-static-unit`, I noticed this in the resulting `index.ts`:

```
const s3BucketYstestNternalLothoAyloads = new aws.s3.Bucket(
        `ystest--nternal-lotho-ayloads`,
        {
```

It looks like some code responsible for converting to camel case is mistakenly removing the first letter of every word. The same thing happens to the dashed resource name.

Fixing this would be a breaking change, since it changes the logical IDs for the Pulumi resources.

DavidSeptimus-Klotho:
Resolves #641

This PR improves sanitization for s3 bucket names and includes the following changes:
- Fixed the original issue in #641 by allowing uppercase letters and then converting the output to lowercase
- Added additional s3 bucket name validation rules

### DIFF
--- a/pkg/sanitization/aws/s3.go
+++ b/pkg/sanitization/aws/s3.go
@@ -6,19 +6,54 @@ import (
 	"github.com/klothoplatform/klotho/pkg/sanitization"
 )
 
-// EnvVarKeySanitizer returns a sanitized environment key when applied.
 var S3BucketSanitizer = sanitization.NewSanitizer(
 	[]sanitization.Rule{
-		// replace "-" or whitespace with "_"
+		// must not contain non-alphanumeric characters other than "-" or "."
+		// uppercase characters will be converted to lowercase
 		{
-			Pattern:     regexp.MustCompile(`[^a-z0-9.-]`),
+			Pattern:     regexp.MustCompile(`[^a-zA-Z\d.-]`),
 			Replacement: "-",
+			Lowercase:   true,
+		},
+		{
+			// must not start with "xn--"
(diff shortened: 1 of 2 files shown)


jhsinger-klotho (approved):

pkg/sanitization/aws/s3.go
```diff
@@ -6,19 +6,54 @@ import (
...
+			Pattern:     regexp.MustCompile(`^[^a-z\d]+|[^a-z\d]+$`),
+			Replacement: "",
+		},
+		{
+			// must not contain repeated periods
+			Pattern:     regexp.MustCompile(`\.\.+`),
+			Replacement: ".",
+		},
+		{
+			// must not be formatted as an IP address (for example, 192.168.5.4)
```
yuval-klotho, line 34: Just to double check, do we need to also check IPV6?

(If so, I'd recommend not doing that via regex, and just trying to parse it using some real library)

yuval-klotho (approved):

pkg/sanitization/aws/s3.go
gordon-klotho, line 34: does this only mean IPv4 addresses or does it include IPv6 as well?

DavidSeptimus-Klotho: IPv6 addresses (e.g. `2001:0db8:85a3:0000:0000:8a2e:0370:7334`) are separated by colons, so they're implicitly not allowed.

@yuval-klotho ^

PR #661, opened 2023-06-01
kms resource and creation

### LINKED ISSUE #660: kms resource and expansion

jhsinger-klotho:
• Does it relate to or fix any issue? closes #660 

Adding kms resources and their respective IaC. Not doing edges at this point since we are not positive on how we are taking in constraints, but everything should be in place outside of the edges in the KB

### Standard checks


pkg/provider/aws/resources/kms.go
```diff
@@ -0,0 +1,172 @@
...
+
+type KmsKeyConfigureParams struct {
+}
+
+func (key *KmsKey) Configure(params KmsKeyConfigureParams) error {
+	key.EnableKeyRotation = true
+	key.Enabled = true
+	key.MultiRegion = false
+	key.KeySpec = "SYMMETRIC_DEFAULT"
+	key.KeyUsage = "ENCRYPT_DECRYPT"
```
yuval-klotho, line 77: can we `const`-ify these two?

yuval-klotho (approved):

pkg/provider/aws/resources/kms.go
jhsinger-klotho: we could, but these fields could be configured in the future to have different values which is why i didnt do that for now

PR #666, opened 2023-06-06
common interface between Construct and Resource

yuval-klotho:
This PR introduces a new interface `BaseConstruct`, with just a single method `Id() ResourceId` (I'm keeping the discrepancy between "Construct" and "Resource" for now; if people prefer, I can either rename `ResourceId` or come up with a new name for `BaseConstruct`, like `AbstractResource`.)

* `Construct` and `Resource` both include `BaseConstruct`
* our `ConstructGraph` is now a graph of `BaseConstruct`s, not of `Construct`s. This means it can contain a heterogeneous mix of constructs and resources.
* In most places, this "just works". Some places really do need a Construct, so I used type checking there. We may be able to generalize those further in the future, but I didn't want to block other progress on that.

### Standard checks

- **Unit tests**: all passing
- **Docs**: n/a
- **Backwards compatibility**: I believe this should work fine; I manually tested with `ts-cloudfs` and it produced an equivalent `compiled/` dir.


pkg/config/config.go
```diff
@@ -42,9 +42,9 @@ type (
 	}
 
 	CloudResourceLink struct {
-		Source string `json:"source,omitempty" yaml:"source,omitempty" toml:"source,omitempty"`
-		Target string `json:"target,omitempty" yaml:"target,omitempty" toml:"target,omitempty"`
-		Type   string `json:"type,omitempty" yaml:"type,omitempty" toml:"type,omitempty"`
+		Source core.ResourceId `json:"source,omitempty" yaml:"source,omitempty" toml:"source,omitempty"`
+		Target core.ResourceId `json:"target,omitempty" yaml:"target,omitempty" toml:"target,omitempty"`
+		Type   string          `json:"type,omitempty" yaml:"type,omitempty" toml:"type,omitempty"`
```
jhsinger-klotho, line 47: i wonder if we should delete this, doesnt seem like we are going to use it, doesnt have to be done in this pr, just saw it

pkg/core/construct_graph.go
```diff
@@ -12,67 +12,74 @@ import (
 
 type (
 	ConstructGraph struct {
-		underlying *graph.Directed[Construct]
+		underlying *graph.Directed[BaseConstruct]
```
jhsinger-klotho, line 15: so do we need the construct graph now? 

When would something be put in here instead of the resource graph?

yuval-klotho: We don't need it, but it's a smaller diff (code-wise and mentally) to keep it. Given that we don't yet know what's going to replace it, I figured I'd rock the boat as little as possible.

pkg/config/config.go
```diff
@@ -42,9 +42,9 @@ type (
 	}
 
 	CloudResourceLink struct {
-		Source string `json:"source,omitempty" yaml:"source,omitempty" toml:"source,omitempty"`
-		Target string `json:"target,omitempty" yaml:"target,omitempty" toml:"target,omitempty"`
-		Type   string `json:"type,omitempty" yaml:"type,omitempty" toml:"type,omitempty"`
+		Source core.ResourceId `json:"source,omitempty" yaml:"source,omitempty" toml:"source,omitempty"`
```
gordon-klotho, line 45: Are Links used for anything right now? How do they fit in with the current direction?

pkg/compiler/compiler.go
```diff
@@ -89,11 +89,10 @@ func (c *Compiler) Compile() error {
 				return core.NewPluginError(p.Name(), err)
 			}
 		}
-		links, err := p.Translate(c.Document.Constructs, c.Document.Resources)
+		_, err := p.Translate(c.Document.Constructs, c.Document.Resources)
```
jhsinger-klotho, line 92: dont we need to change the plugin interface if we remove the links? Am i just missing that?

yuval-klotho: It returns a `core.CloudResourceLink interface` — not the `config.CloudResourceLink struct` that I've rm'ed. That said, it looks like nothing implements that interface, and we always just return `nil`, so I can clean that up too.

pkg/core/construct_graph.go
jhsinger-klotho: thats also an implementation aspect we are going to have to deal with because BaseConstructs can be cyclic and also can exist within the resource graph. It likely isnt an issue as we can just ensure that post expansion theres no cycles but we will have to allow cycles while constructs still exist within the resource graph/dag

jhsinger-klotho (approved):

pkg/core/construct_graph.go
jhsinger-klotho: wait wouldnt they both need to be of BaseConstruct?

PR #667, opened 2023-06-08
be able to load and export resource dag

jhsinger-klotho:
• Does it relate to or fix any issue? closes https://github.com/klothoplatform/klotho-pro/issues/88

The ability to load a BaseConstruct graph by calling all the providers load graph method.

Theres also a loadConstructs method for our higher level constructs

We also need to ensure that namespaced resources get loaded and added after their parent resource so we use reflection to call a load method (only subnets at the moment)

We will also output the graph as resources.yaml on compilation

### Standard checks

### DIFF
--- a/pkg/provider/aws/provider.go
+++ b/pkg/provider/aws/provider.go
@@ -47,8 +47,8 @@ func (a *AWS) MapResourceDirectlyToConstruct(resource core.Resource, construct c
 }
 
 func (a *AWS) GetResourcesDirectlyTiedToConstruct(construct core.BaseConstruct) ([]core.Resource, bool) {
-	resources, found := a.constructIdToResources[construct.Id()]
-	return resources, found
+	awsResources, found := a.constructIdToResources[construct.Id()]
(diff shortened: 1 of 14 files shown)


pkg/provider/aws/provider.go
```diff
@@ -49,3 +51,56 @@ func (a *AWS) GetResourcesDirectlyTiedToConstruct(construct core.BaseConstruct)
 	resources, found := a.constructIdToResources[construct.Id()]
 	return resources, found
 }
+
+func (a *AWS) LoadGraph(graph core.OutputGraph, dag *core.ConstructGraph) error {
```
yuval-klotho, line 55: It'd be nice to have a unit test for this, if only to see what the format is and do a quick smoke test.

yuval-klotho (approved):

pkg/provider/aws/provider.go
jhsinger-klotho: yeah i can add one for it

PR #670, opened 2023-06-09
start of engine

### LINKED ISSUE #669: define API for engine

jhsinger-klotho:
closes #669 

starts the definition of the engine by introducing constraints and parsing them from the input yaml file

### Standard checks

### DIFF
--- a/pkg/engine/constraints.go
+++ b/pkg/engine/constraints.go
@@ -1,0 +1,183 @@
+package engine
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"reflect"
+
+	"github.com/klothoplatform/klotho/pkg/core"
+	"gopkg.in/yaml.v3"
+	"k8s.io/utils/strings/slices"
+)
+
+type (
+
+	// Constraint is an interface detailing different intents that can be applied to a resource graph
+	Constraint interface {
+		// Scope returns where on the resource graph the constraint is applied
+		Scope() ConstraintScope
+	}
+
+	// BaseConstraint is the base struct for all constraints
+	// BaseConstraint is used in our parsing to determine the Scope of the constraint and what go struct it corresponds to
+	BaseConstraint struct {
+		Scope ConstraintScope `yaml:"scope"`
+	}
+
+	// Edge is a struct that represents how we take in data about an edge in the resource graph
+	Edge struct {
+		Source core.ResourceId `yaml:"source"`
+		Target core.ResourceId `yaml:"target"`
+	}
+
+	// ApplicationConstraint is a struct that represents constraints that can be applied on the entire resource graph
+	ApplicationConstraint struct {
+		Operator ConstraintOperator `yaml:"operator"`
+		Node     core.ResourceId    `yaml:"node"`
+		Edge     Edge               `yaml:"edge"`
+	}
+
+	// ConstructConstraint is a struct that represents constraints that can be applied on a specific construct in the resource graph
+	ConstructConstraint struct {
+		Operator   ConstraintOperator `yaml:"operator"`
+		Target     core.ResourceId    `yaml:"target"`
+		Type       string             `yaml:"type"`
+		Attributes map[string]any     `yaml:"attributes"`
+	}
+
+	// EdgeConstraint is a struct that represents constraints that can be applied on a specific edge in the resource graph
+	EdgeConstraint struct {
+		Operator ConstraintOperator `yaml:"operator"`
+		Target   Edge               `yaml:"target"`
+		Node     core.ResourceId    `yaml:"node"`
+	}
+
+	// NodeConstraint is a struct that represents constraints that can be applied on a specific node in the resource graph.
+	// NodeConstraints are used to control intrinsic properties of a node in the resource graph
+	NodeConstraint struct {
+		BaseConstraint
+		Target   core.ResourceId
+		Property string
+		Value    any
+	}
+
+	// ConstraintScope is an enum that represents the different scopes that a constraint can be applied to
+	ConstraintScope string
+	// ConstraintOperator is an enum that represents the different operators that can be applied to a constraint
+	ConstraintOperator string
+)
+
+const (
+	ApplicationConstraintScope ConstraintScope = "application"
+	ConstructConstraintScope   ConstraintScope = "construct"
+	EdgeConstraintScope        ConstraintScope = "edge"
+	NodeConstraintScope        ConstraintScope = "node"
+
+	MustExistConstraintOperator      ConstraintOperator = "must_exist"
+	MustContainConstraintOperator    ConstraintOperator = "must_contain"
+	MustNotExistConstraintOperator   ConstraintOperator = "must_not_exist"
+	MustNotContainConstraintOperator ConstraintOperator = "must_not_contain"
+	AddConstraintOperator            ConstraintOperator = "add"
+	EqualsConstraintOperator         ConstraintOperator = "equals"
+)
+
+func (b *ApplicationConstraint) Scope() ConstraintScope {
+	return ApplicationConstraintScope
+}
+
+func (b *ConstructConstraint) Scope() ConstraintScope {
+	return ConstructConstraintScope
+}
+
+func (b *EdgeConstraint) Scope() ConstraintScope {
+	return EdgeConstraintScope
+}
+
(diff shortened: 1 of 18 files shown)


pkg/engine/constraints.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	// NodeConstraint struct {
+	// 	BaseConstraint
+	// 	Target   core.ResourceId
+	// 	Property string
+	// 	Value    any
+	// }
+
+	ConstraintLevel    string
```
yuval-klotho, line 68: What kind of values will this be? Is it enum-like (e.g. `high`, `low`, `required`) or arbitrary?

Edit: oh, I see in the example below. I'd suggest calling this a scope, not level. Level is ambiguous, and I initially thought this was a _priority_ level.

pkg/engine/samples/constraints.yaml
```diff
@@ -0,0 +1,25 @@
+- level: application
+  operator: Add
+  node: klotho:execution_unit:my_compute
```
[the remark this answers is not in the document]
jhsinger-klotho: its the provider for constructs. Was done as a part of yuvals pr

pkg/engine/constraints.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/klothoplatform/klotho/pkg/core"
+	"github.com/klothoplatform/klotho/pkg/graph"
+	"gopkg.in/yaml.v3"
+)
+
+type (
+	Constraint interface {
+		Level() ConstraintLevel
+		IsSatisfied(dag *core.ResourceGraph) bool
+		CanApply() bool
```
yuval-klotho, line 16: What does `CanApply` mean? Does it mean "this is potentially relevant", or "this is potentially satisfiable"?

Looking at where it's used below, it seems like it's "this is potentially relevant". If so, do we really need this method? Why not just let `Apply(dag)` determine that? If the constraint isn't applicable, `Apply` can short-circuit return a no-error with no modification.

Similarly, I don't understand what `IsSatisfied` is for.

pkg/engine/constraints.go
```diff
@@ -0,0 +1,127 @@
...
+		Apply(dag *core.ResourceGraph) error
```
yuval-klotho, line 17: I think we may need more than just a binary error/no-error. The caller may want to know some details about _how_ the constraint was used, and especially if it wasn't used.

For example, let's say I have a graph `APIGW -> Lambda`, and I say "only use managed storage services." That constraint is a no-op: it's satisfiable by doing nothing, but we should probably provide some way for infracopilot to know that it turned into a no-op constraint, so that we can report that back to the user as a warning.

pkg/engine/constraints.go
```diff
@@ -0,0 +1,127 @@
...
+	}
```
yuval-klotho, line 18: Second comment on `Apply(dag)` (I'm replying on a second line so that we can keep the threads separate): this interface doesn't really let the user see if two constraints are individually possible, but impossible when combined. They'll see the error, but not the real reason.

For example, let's say I have a graph `APIGW -> Compute`, and I say "make the compute a serverless EC2". That's impossible, by definition; but I'm guessing we would turn this into two constraints: "compute must be serverless" and "compute must be EC2", and then apply them in some order. I think the user would get back an error saying either "compute_01 cannot be serverless because it has type EC2" or "compute_01 cannot be EC2 because it is serverless" (or maybe even "compute_01 cannot be EC2 because it is Lambda", depending on how specific `Apply(~)`'s actions get), but the error I'd really want is something like "serverless and EC2 are mutually exclusive".

pkg/engine/engine.go
```diff
@@ -0,0 +1,88 @@
...
+		return dag, err
+	}
+
+	return dag, nil
+}
+
+func ValidateAndApply(graph *core.ResourceGraph, constraints []Constraint) (bool, error) {
+	var joinedErr error
+	allSatisfied := true
+	for _, constraint := range constraints {
```
yuval-klotho, line 74: I think we should have a stable sorting of these — not just by level/scope (as the sort func above does), but even within each level/scope. I think a lot of constraints won't be commutative, so if we want "same graph + same constraints => same result" (which I think we should strive for), we need to make sure the constraints are applied in a consistent order.

pkg/engine/constraints.go
jhsinger-klotho: scope makes sense to me, definitely better naming

pkg/engine/constraints.go
jhsinger-klotho: so we could also do that, this part wont make sense yet as its me thinking about the engine and not the api portion.

But my thought was

IsSatisfied is just a method which checks to see if the dag satisfies the constraint
CanApply is a method that sees if the constraint is a type of constraint that we can satisfy by making changes to the dag. This might not be necessary since we should jsut be able to run Apply, but not important yet, its all implementation details behind the scene. Was just trying to start thinking about interfaces and stuff for our go layer

pkg/engine/constraints.go
jhsinger-klotho: yeah so i think in the engine we need to build a Decision matrix that can show what we did based on constraints

pkg/engine/constraints.go
jhsinger-klotho: yeah the interface isnt thought through yet, but also will not be publicly visible. Eventually if we ever wanted the extensibility then we would need to expose these, but that hasnt been thought through yet

yuval-klotho: I'm not suggesting the interface needs to be publicly visible, but I do think infracopilot needs a bit more insight than just "this one constraint failed".

pkg/engine/constraints.go
gordon-klotho: On that note, could you add godoc comments explaining what each function on the interface does and how we (initially) intend to use it?

jhsinger-klotho: yeah ill definitely add some and brush that up to make it clear. This pr when i move it out of draft will also just purely be constructs and parsing them from a yaml file so a lot of it will be clearer

pkg/engine/constraints.go
jhsinger-klotho: yeah agreed, ill get something set up to work that in so we can produce the reasoning on the klotho engine side

pkg/engine/constraints.go
```diff
@@ -0,0 +1,183 @@
...
+	return ConstructConstraintScope
+}
+
+func (b *EdgeConstraint) Scope() ConstraintScope {
+	return EdgeConstraintScope
+}
+
+// DecodeYAMLNode is a helper function that decodes a yaml node into a struct representing different constraints
+func DecodeYAMLNode[T Constraint](node *yaml.Node) (constraint T, err error) {
+	constraint = reflect.New(reflect.TypeOf(constraint).Elem()).Interface().(T)
```
gordon-klotho, line 99:
```suggestion
func DecodeYAMLNode[T interface{ Constraint; *I }, I any](node *yaml.Node) (constraint T, err error) {
	constraint = new(I)
```
A little nicer to use generics instead of reflection.

pkg/engine/constraints.go
```diff
@@ -0,0 +1,183 @@
...
+	err = extraFields(node, reflect.ValueOf(constraint))
+	if err != nil {
+		return constraint, err
+	}
+	err = node.Decode(constraint)
+	return constraint, err
+}
+
+// ParseConstraintsFromFile is a helper function that parses a yaml file into a map of constraints
+func ParseConstraintsFromFile(path string) (map[ConstraintScope][]Constraint, error) {
```
gordon-klotho, line 109: I'm not sure this is the structure we want. We may want to allow grouping and ordering from the caller side so that the constraints are evaluated in a specific, predictable order.

pkg/engine/constraints.go
jhsinger-klotho: can go ahead and do that

pkg/engine/constraints.go
jhsinger-klotho: lets chat about this because this is likely something we want to define. 

My thought is that we have levels to constraints and it must go in order of 
1. application
2. Construct
3. edge
4. node

if were saying we want even more ordering within that we can build it in, but are we saying well hold the order of the yaml file?

jhsinger-klotho:
closing in favor of #674

PR #672, opened 2023-06-12
refactor construct and resource interfaces

jhsinger-klotho:
This pr is intended to decouple the compiler UX from our code's core so that the engine doesnt rely on compiler concepts.

- The main concepts are changing the Resource interface to have references to BaseConstructs instead of AnnotationKeys
- Changing the Construct interface to have its id derived from its own Name and Type rather than from an annotationKey
   -  This allows us to have different ids for each struct rather than `persist` covering 5 different structs
 - AnnotationCapability was added to Construct for the time being until we can further decouple

The remainder of the changes are changes to get everything align to the new interfaces

```
	Construct interface {
		BaseConstruct

		// AnnotationCapability returns the annotation capability of the construct. This helps us tie the annotation types to the constructs for the time being
		AnnotationCapability() string
	}

	// BaseConstructRefSet is a set of BaseConstructs
	BaseConstructSet map[BaseConstruct]struct{}

	// Resource describes a resource at the provider, infrastructure level
	Resource interface {
		BaseConstruct
		// BaseConstructRef returns a set of BaseConstructs which caused the creation of this resource
		BaseConstructRef() BaseConstructSet
	}
```

### Standard checks


pkg/core/resources.go
```diff
@@ -88,20 +86,28 @@ const (
...
+func IsConstructOfAnnotationCapability(baseConstruct BaseConstruct, cap string) bool {
 	cons, ok := baseConstruct.(Construct)
 	if !ok {
 		return false
 	}
-	return cons.Provenance().Capability == cap
+	return cons.AnnotationCapability() == cap
+}
+
+func ListAllConstructs() []Construct {
```
gordon-klotho, line 97: This seems brittle and error prone, but I don't have a better suggestion right now.

pkg/core/resources.go
```diff
@@ -18,15 +18,19 @@ type (
...
 	Construct interface {
 		BaseConstruct
-		// Provenance returns the AnnotationKey that the construct was created by
-		Provenance() AnnotationKey
+
+		// AnnotationCapability returns the annotation capability of the construct. This helps us tie the annotation types to the constructs for the time being
+		AnnotationCapability() string
 	}
 
+	// BaseConstructsRefSet is a set of BaseConstructs
```
gordon-klotho, line 26:
```suggestion
```
nit: pretty clear what this is - a godoc which reiterates its name isn't necessary.

pkg/core/resources.go
```diff
@@ -18,15 +18,19 @@ type (
...
+	BaseConstructSet map[BaseConstruct]struct{}
+
 	// Resource describes a resource at the provider, infrastructure level
 	Resource interface {
 		BaseConstruct
-		// KlothoConstructRef returns AnnotationKey of the klotho resource the cloud resource is correlated to
-		KlothoConstructRef() AnnotationKeySet
```
gordon-klotho, line 29: Seems like AnnotationKeySet can be removed?

gordon-klotho (approved):
Hard to separate signal from the noise. My review is for `core` and `compiler` packages.

pkg/core/construct_graph_test.go
```diff
@@ -236,66 +236,66 @@ func Test_GetUpstreamDependencies(t *testing.T) {
...
-				g.AddVertex(c)
-			}
-			constructs := constructGraph.GetResourcesOfCapability(tt.capability)
-			if tt.want != nil && !assert.NotNil(constructs) {
-				return
-			}
-			assert.ElementsMatch(tt.want, constructs)
-		})
-	}
+// func Test_GetResourcesOfCapability(t *testing.T) {
```
gordon-klotho, line 239: uncomment or remove

pkg/core/resources.go
jhsinger-klotho: yeah i didnt know if we were going to want to keep it for the compiler side, which is why i left it in, but i could remove it if we want

pkg/core/resources.go
jhsinger-klotho: will remove

pkg/core/construct_graph_test.go
jhsinger-klotho: i can uncomment, i forgot i got this to work

pkg/core/construct_graph_test.go
```diff
@@ -317,7 +317,7 @@ func Test_LoadConstructsIntoGraph(t *testing.T) {
 					{Provider: AbstractConstructProvider, Type: "orm", Name: "test_orm"},
 					{Provider: AbstractConstructProvider, Type: "kv", Name: "test_kv"},
 					{Provider: AbstractConstructProvider, Type: "fs", Name: "test_fs"},
-					{Provider: AbstractConstructProvider, Type: "secret", Name: "test_secret"},
+					{Provider: AbstractConstructProvider, Type: "config", Name: "test_secret"},
```
gordon-klotho, line 320: There's non-config secrets (for Python and JS) right?

jhsinger-klotho: yeah, you can still pass in secrets, but we dont have a way to pass in the actual secrets paths yet since we only pass in id, which is why i use config.

Its something well need to likely support outside of infracopilot but we need to discuss how we want to do that for importing when we are importing more than just the id and the parsing gets much trickier

jhsinger-klotho: i guess that could be the attributes part of a constructConstraint?

pkg/provider/aws/execution_unit.go
```diff
@@ -154,23 +154,26 @@ func (a *AWS) handleHelmChartAwsValues(chart *kubernetes.HelmChart, unit *core.E
...
-	ref, oneRef := refs.GetSingle()
-	if !oneRef {
+func (a *AWS) getLambdaConfiguration(result *core.ConstructGraph, dag *core.ResourceGraph, refs core.BaseConstructSet) (resources.LambdaFunctionConfigureParams, error) {
+	if len(refs) > 1 {
 		return resources.LambdaFunctionConfigureParams{}, fmt.Errorf("lambda must only have one construct reference")
 	}
+	var ref core.BaseConstruct
+	for r := range refs {
+		ref = r
+	}
```
DavidSeptimus-Klotho, line 172: won't this result in a panic if there are 0 refs? (same comment applies in multiple places)

jhsinger-klotho: I dont think it panics, but we may have an error, the original code was if there [wasnt a single ref](https://github.com/klothoplatform/klotho/blob/3c7f578fe18977a913132ef57bb795736052162a/pkg/provider/aws/execution_unit.go#L167), so i need the first condition to check if there is one ref or 2.

jhsinger-klotho: oh we would panic, i see because ref would be nil and we would try to call nil on it

jhsinger-klotho: just pushed an update for this

gordon-klotho: It won't panic, but it will leave `ref` as the zero-value.

jhsinger-klotho: realistically this whole logic will be moved into the compiler ux and these will be converted into constraints, but not a reason to not have it work now

DavidSeptimus-Klotho (approved):

PR #676, opened 2023-06-13
Simplify LoadConstructIntoGraph edges

gordon-klotho:
No need to convert the Id to resource only to extract the Id again to give to AddDependency.

### Standard checks

### DIFF
--- a/pkg/core/construct_graph.go
+++ b/pkg/core/construct_graph.go
@@ -150,23 +150,10 @@ func LoadConstructsIntoGraph(input OutputGraph, graph *ConstructGraph) error {
 	}
 
 	for _, edge := range input.Edges {
-		shouldAdd := true
-		src, err := getConstructFromInputId(edge.Source)
-		if err != nil {
-			joinedErr = errors.Join(joinedErr, err)
-			shouldAdd = false
-		}
-		dst, err := getConstructFromInputId(edge.Destination)
-		if err != nil {
-			joinedErr = errors.Join(joinedErr, err)
-			shouldAdd = false
-		}
-		if shouldAdd {
-			graph.AddDependency(src.Id(), dst.Id())
-		}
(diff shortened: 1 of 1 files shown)


pkg/core/construct_graph.go
```diff
@@ -150,23 +150,10 @@ func LoadConstructsIntoGraph(input OutputGraph, graph *ConstructGraph) error {
...
-		}
-		dst, err := getConstructFromInputId(edge.Destination)
-		if err != nil {
-			joinedErr = errors.Join(joinedErr, err)
-			shouldAdd = false
-		}
-		if shouldAdd {
-			graph.AddDependency(src.Id(), dst.Id())
-		}
+		graph.AddDependency(edge.Source, edge.Destination)
```
jhsinger-klotho, line 153: do we care if the type isnt valid? Or will we just stick it in the graph (add dep creates resources as well).

The only reasoning being we should ensure each node in the graph is something we understand at some point based on the kb or provider we have loaded in.

Can defer on this for now if we want, because i think well do some restructuring that will make this easier just wanted to get your thoughts and then maybe cut a ticket?

jhsinger-klotho (approved):
one comment which i wont consider blocking since we are the only people using this functionality, but may be good to think about for the future

pkg/core/construct_graph.go
gordon-klotho: > do we care if the type isnt valid?

This would surface as node not existing. I think AddDependency with IDs needs to return an error for cases like node not existing (then we check it here), but that's beyond the scope of this change.

>  (add dep creates resources as well)

This one does not, because it's just the IDs it has nothing to add.

jhsinger-klotho: oh ok cool, then sounds good to me

PR #679, opened 2023-06-13
initial work

jhsinger-klotho:
### Standard checks


pkg/cli/klothomain.go
```diff
@@ -257,7 +260,7 @@ func (km KlothoMain) run(cmd *cobra.Command, args []string) (err error) {
 			cmd.SilenceUsage = true
 		},
 	}
-	defer analyticsClient.PanicHandler(&err, errHandler)
+	// defer analyticsClient.PanicHandler(&err, errHandler)
```
gordon-klotho, line 263: Do we still need to do this? I've been using -internalDebug and it's worked alright for showing panics to me. (also, make sure to uncomment before merge)

gordon-klotho (approved):
I didn't check all changes, but the engine and constraints package looks good.

PR #682, opened 2023-06-16
cleanup provider

jhsinger-klotho:

This is a piece of https://github.com/klothoplatform/klotho/pull/681/files

This cleans up the provider to fit the new engine makeup and only that (which means for the time being we have broken the compiler route, but this was already the case when switching to the new engine)

this combines all functionality of expanding constructs and the tests into a single file

for now im still leaving the config file just so we can build our binary

### Standard checks

- **Unit tests**: Any special considerations? added tests for each expansion route


pkg/provider/aws/resources/ecs.go
```diff
@@ -261,6 +267,14 @@ func (s *EcsService) Id() core.ResourceId {
 	}
 }
 
+func (td *EcsService) DeleteCriteria() core.DeleteCriteria {
+	return core.DeleteCriteria{
+		RequiresNoUpstream:     true,
+		RequiresNoDownstream:   true,
+		RequiresExplicitDelete: true,
+	}
```
DavidSeptimus-Klotho, line 275: `RequiresNoUpstreamOrDownstream`? There are a few others like this in other files.

DavidSeptimus-Klotho (approved):

pkg/provider/aws/resources/ecs.go
jhsinger-klotho: it would just default to false which is whats intended here because the combination of requires no upstream and requires no downstream will cover both scenarios

PR #683, opened 2023-06-16
core changes

jhsinger-klotho:

this pr follows https://github.com/klothoplatform/klotho/pull/682

This contains core changes to use the engine in the compiler flow and read in input graphs/output metadata

### Standard checks


pkg/cli/klothomain.go
```diff
@@ -410,31 +408,30 @@
...
-
+	provider, err := providers.GetProvider(&appCfg)
+	if err != nil {
+		return err
+	}
+	kb, err := providers.GetKnowledgeBase(&appCfg)
+	if err != nil {
+		return err
+	}
+	engine := engine.NewEngine(provider, kb)
```
DavidSeptimus-Klotho, line 419: nit: this var is shadowing the `engine` import.

pkg/core/resource_graph.go
```diff
@@ -114,6 +128,27 @@ func (rg *ResourceGraph) RemoveDependency(source ResourceId, target ResourceId)
...
+		err := rg.RemoveDependency(edge.Source.Id(), edge.Destination.Id())
+		if err != nil {
+			return err
+		}
+	}
+	return rg.RemoveResource(source)
+}
+
+func (rg *ResourceGraph) RemoveResource(resource Resource) error {
+	zap.S().Infof("Removing resource %s", resource.Id())
```
DavidSeptimus-Klotho, line 148: nit: this should probably be a debug message?

pkg/core/resources.go
```diff
@@ -25,10 +25,23 @@ type (
...
+	// Delete criteria is supposed to tell us when we are able to delete a resource based on its dependencies
+	DeleteCriteria struct {
+		// RequiresNoUpstream is a boolean that tells us if deletion relies on there being no upstream resources
+		RequiresNoUpstream bool
+		// RequiresNoDownstream is a boolean that tells us if deletion relies on there being no downstream resources
+		RequiresNoDownstream bool
+		// RequiresExplicitDelete is a boolean that tells us if deletion relies on the resource being explicitly deleted
+		RequiresExplicitDelete bool
+		// RequiresNoUpstreamOrDownstream is a boolean that tells us if deletion relies on there being no upstream or downstream resources
+		RequiresNoUpstreamOrDownstream bool
```
DavidSeptimus-Klotho, line 37: Is this shorthand for a combination of `RequiresNoUpstream` and `RequiresNoDownstream`? If so, will it require us to make extra checks whenever we want to evaluate those constraints (potentially leading to bugs when we forget to check both formulations)?

pkg/cli/klothomain.go
jhsinger-klotho: updating in my new pr

pkg/core/resource_graph.go
jhsinger-klotho: were going to have to rework our logging, i do think it should be info but we double print a lot so deferring for now

pkg/core/resources.go
jhsinger-klotho: its the or case, so like for rds proxy target group or for a load balancer, we view those as connection pieces so if theres nothing up OR downstream then we can remove. For the and case we use the 2 above.

Also this is all from the spike so i assume we will model this better in the future as we learn

PR #684, opened 2023-06-16
add reconciliation to the engine

jhsinger-klotho:
follows https://github.com/klothoplatform/klotho/pull/683

Creates a dag for data flow, used for another visualization 

Adds reconciliation to the engine so we can delete resources

### Standard checks


pkg/engine/dataflow.go
```diff
@@ -0,0 +1,84 @@
+package engine
+
+import (
+	"github.com/klothoplatform/klotho/pkg/collectionutil"
+	"github.com/klothoplatform/klotho/pkg/core"
+	"github.com/klothoplatform/klotho/pkg/provider/aws/resources"
+	"go.uber.org/zap"
+)
+
+func (e *Engine) GetDataFlowDag() *core.ResourceGraph {
```
DavidSeptimus-Klotho, line 10: This is neat.

pkg/knowledge_base/edge_kb.go
```diff
@@ -402,3 +403,41 @@ func (kb EdgeKB) findPathsInGraph(source, dest core.Resource, stack []graph.Edge
 	delete(visited, source)
 	return result
 }
+
+func (kb EdgeKB) GetTrueUpstream(source core.Resource, dag *core.ResourceGraph) []core.Resource {
```
DavidSeptimus-Klotho, line 407: Can you add docs for `GetTrueUpstream` and `GetTrueDownstream` to make their purpose clear?

pkg/visualizer/file.go
```diff
@@ -49,7 +50,7 @@ type (
 )
 
 func (f *File) Path() string {
-	return "topology.yaml"
+	return fmt.Sprintf("%stopology.yaml", f.PathPrefix)
```
DavidSeptimus-Klotho, line 53: nit: should we be `filepath.Join()` here instead of a string format?

pkg/visualizer/plugin.go
```diff
@@ -106,6 +106,46 @@ func (p Plugin) Translate(dag *core.ResourceGraph) ([]core.File, error) {
...
 }
 
+// Translate implements compiler.IaCPlugin - although it's not strictly an IaC plugin, it uses the same API
+func (p Plugin) Generate(dag *core.ResourceGraph, filenamePrefix string) ([]core.File, error) {
+	api := visApi{client: p.Client}
+
+	if validateTypes {
+		types := TypesChecker{DAG: dag}
+
+		if resp, err := api.request(http.MethodGet, `validate-types`, `application/text`, ``, types); err != nil {
```
DavidSeptimus-Klotho, line 116: We might want to disable this warning since the new visualizer doesn't have this capability just yet.

DavidSeptimus-Klotho (approved):

pkg/knowledge_base/edge_kb.go
jhsinger-klotho: yeah going to add docs to the new pr

pkg/visualizer/file.go
jhsinger-klotho: the path prefix is really just a filename prefix so i will rename

PR #685, opened 2023-06-19
Iacval refactor

jhsinger-klotho:

makes iac value an interface so that each provider can implement their own "provider value" so that we can parse yaml into their respective IRs

### Standard checks

### DIFF
--- a/pkg/core/resources.go
+++ b/pkg/core/resources.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/mitchellh/mapstructure"
 	"github.com/pkg/errors"
+	"gopkg.in/yaml.v3"
 )
 
 type (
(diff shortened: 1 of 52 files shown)


pkg/core/resources.go
```diff
@@ -26,7 +27,7 @@ type (
 	BaseConstructSet map[ResourceId]BaseConstruct
 
 	// Delete criteria is supposed to tell us when we are able to delete a resource based on its dependencies
```
DavidSeptimus-Klotho, line 29:
```suggestion
	// DeleteContext is supposed to tell us when we are able to delete a resource based on its dependencies
```

DavidSeptimus-Klotho (approved):

pkg/core/resources.go
jhsinger-klotho: updated in new pr

PR #687, opened 2023-06-20
Reconciler cleanup resources

jhsinger-klotho:

This code achieves cleaning up the rds proxy secret in the remove rds proxy case.

We have a new field on an edge to signal that the edge is linked in terms of deletion so that we know we can go delete resources

this pr also contains the finishing touches on parsing in the AwsResourceValue from the yaml files

### Standard checks


pkg/engine/loader.go
```diff
@@ -74,3 +86,75 @@ func (e *Engine) LoadConstraintsFromFile(path string) (map[constraints.Constrain
 	}
 	return constraints.ParseConstraintsFromFile(bytesArr)
 }
+
+func correctPointers(source core.BaseConstruct, resourceMap map[core.ResourceId]core.BaseConstruct) error {
```
gordon-klotho, line 90: Can you add a little documentation to this for what it does and why it's needed?

pkg/engine/loader.go
```diff
@@ -74,3 +86,75 @@ func (e *Engine) LoadConstraintsFromFile(path string) (map[constraints.Constrain
...
+			}
+
+		default:
+			loadNestedResourcesFromIds(source, fieldValue, resourceMap)
+		}
+	}
+	return nil
+}
+
+func loadNestedResourcesFromIds(source core.BaseConstruct, targetValue reflect.Value, resourceMap map[core.ResourceId]core.BaseConstruct) {
```
gordon-klotho, line 120: Does this one differ from the one in `core`?

pkg/engine/reconciler.go
```diff
@@ -17,17 +48,52 @@ func (e *Engine) deleteResource(resource core.Resource, explicit bool) bool {
 		log.Debug("Cannot delete resource as it was not explicitly requested")
 		return false
 	}
+	if !overrideExplicit {
+		explicit = false
+	}
```
gordon-klotho, line 53: What's the difference then between `explicit` and `overrideExplicit`?

pkg/engine/reconciler.go
```diff
@@ -71,10 +149,24 @@ func (e *Engine) deleteResource(resource core.Resource, explicit bool) bool {
...
+		for _, reflectResource := range reflectResources {
+			if res == reflectResource {
+				explicit = true
+				continue
+			}
+		}
+		e.deleteResource(res, explicit, false)
 	}
 	return true
 }
```
gordon-klotho, line 172: Definitely could use some unit tests on these to help wrap my head around what these reconciler functions are supposed to do. Some more godocs as well, where appropriate.

pkg/knowledge_base/edge_kb.go
```diff
@@ -31,6 +31,9 @@ type (
 		// ReverseDirection is specified when the data flow is in the opposite direction of the edge
 		// This is used in scenarios where we want to find paths, only allowing specific edges to be bidirectional
 		ReverseDirection bool
+		// DeletetionDependent is used to specify edges which should not influence the deletion criteria of a resource
+		// 0 specifies the are dependent while 1 specifies they are not dependent and will not influence the deletion criteria
```
gordon-klotho, line 35: Why `0`/`1` instead of `true`/`false`? If we want more than 2 cases in the future, would recommend having a separate "enum" type.

pkg/knowledge_base/edge_kb.go
```diff
@@ -404,6 +407,8 @@ func (kb EdgeKB) findPathsInGraph(source, dest core.Resource, stack []graph.Edge
 	return result
 }
 
+// GetTrueUpstream takes in a resource and returns all upstream resources which exist in the dag, if their edge does not specify the reverse direction flag.
```
gordon-klotho, line 410:
```suggestion
// GetTrueUpstream takes in a resource and returns all upstream resources which exist in the dag, if their edge does not specify the reverse direction flag in the knowledge base.
```
nit: I know this makes sense in context of being a method on EdgeKB, but this helps make it clear why this function is here as opposed to on the ResourceGraph.

pkg/provider/aws/resources/aws_resource_value.go
```diff
@@ -25,9 +31,52 @@ func (r *AwsResourceValue) SetResource(res core.Resource) {
...
+		return nil
+	}
+
+	typeToResource := make(map[string]core.Resource)
+	for _, res := range ListAll() {
+		typeToResource[res.Id().Type] = res
+	}
+	// Subnets are special because they have a type that is not the same as their resource type since it uses a characteristic of the subnet
+	typeToResource["subnet_private"] = &Subnet{}
+	typeToResource["subnet_public"] = &Subnet{}
```
gordon-klotho, line 55: This feels like it should be done in `ListAll`

jhsinger-klotho: yeah i think you may be right, probably easier to do that

pkg/engine/loader.go
jhsinger-klotho: yeah changed the name to explain the difference

jhsinger-klotho: and added some docs

PR #690, opened 2023-06-23
Improves parent/child relationship detection in the dataflow view

DavidSeptimus-Klotho:
This change improves the dataflow view's detection of parent/child relationships and handling of edge summarization.

### DIFF
--- a/pkg/engine/dataflow.go
+++ b/pkg/engine/dataflow.go
@@ -29,12 +29,15 @@ func (e *Engine) GetDataFlowDag() *core.ResourceGraph {
 		resources.EKS_CLUSTER_TYPE,
 	}
 
+	// Add relevant resources to the dataflow DAG
 	for _, resource := range e.Context.EndState.ListResources() {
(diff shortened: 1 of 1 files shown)


pkg/engine/dataflow.go
```diff
@@ -10,7 +10,7 @@ import (
 func (e *Engine) GetDataFlowDag() *core.ResourceGraph {
 	dataFlowDag := core.NewResourceGraph()
 	typesWeCareAbout := []string{
-		resources.LAMBDA_FUNCTION_TYPE,
+		resources.API_GATEWAY_REST_TYPE,
```
gordon-klotho, line 13: This is also on line 16

gordon-klotho (approved):
Not sure why LAMBDA was removed from `typesWeCareAbout` but the other changes are good

DavidSeptimus-Klotho:
> Not sure why LAMBDA was removed from `typesWeCareAbout` but the other changes are good

accidental. I've fixed that.

PR #691, opened 2023-06-23
Internal/dedupe api integration creates

gordon-klotho:
Fixes error seen by infracopilot

### Standard checks


pkg/provider/aws/construct_expansion.go
```diff
@@ -138,7 +138,7 @@ func (a *AWS) expandExpose(dag *core.ResourceGraph, expose *core.Gateway, constr
 	}
 	switch constructType {
 	case resources.API_GATEWAY_REST_TYPE:
-		stage, err := core.CreateResource[*resources.ApiStage](dag, resources.ApiStageCreateParams{
+		stage, err := core.CreateResource[*resources.RestApi](dag, resources.RestApiCreateParams{
```
DavidSeptimus-Klotho, line 141: nit: can you rename the variable too?

DavidSeptimus-Klotho (approved):

pkg/provider/aws/construct_expansion.go
```diff
@@ -138,15 +138,15 @@ func (a *AWS) expandExpose(dag *core.ResourceGraph, expose *core.Gateway, constr
 	}
 	switch constructType {
 	case resources.API_GATEWAY_REST_TYPE:
-		stage, err := core.CreateResource[*resources.ApiStage](dag, resources.ApiStageCreateParams{
+		api, err := core.CreateResource[*resources.RestApi](dag, resources.RestApiCreateParams{
```
jhsinger-klotho, line 141: i dont think we want this, we would never create the stage and deployment then

sample_construct_graphs/one_gateway_two_lambda.yaml
```diff
@@ -0,0 +1,22 @@
+constraints:
```
jhsinger-klotho, line 1: instead could you start a test folder where we can have inputs outputs? that ways we have samples we can run through the engine with expected outputs

PR #693, opened 2023-06-28
Allows lambdas to be assigned to a specific vpc

DavidSeptimus-Klotho:
Configures Security Groups and Subnets for a Lambda Function when a direct `lambda -> vpc` edge exists in the resource graph.

input:
```yaml
constraints: []
edges:
  - destination: aws:vpc:vpc_01
    source: aws:lambda_function:lambda_01
  - destination: aws:vpc:vpc_01
    source: aws:rds_instance:rds_01
resources:
  - aws:vpc:vpc_01
  - aws:rds_instance:rds_01
  - aws:lambda_function:lambda_01
```

output:

```yaml
infracopilot:
  provider: aws
  resources:
    vpc/vpc_01: # aws:vpc:vpc_01

    rds_instance/rds_01: # aws:rds_instance:rds_01
        parent: vpc/vpc_01


    lambda_function/lambda_01: # aws:lambda_function:lambda_01
        parent: vpc/vpc_01
```

diagram:


pkg/provider/aws/resources/lambda.go
```diff
@@ -103,6 +103,44 @@ func (lambda *LambdaFunction) MakeOperational(dag *core.ResourceGraph, appName s
...
 	}
+
+	downstreamVpcs := core.GetDownstreamResourcesOfType[*Vpc](dag, lambda)
+	upstreamVpcs := core.GetUpstreamResourcesOfType[*Vpc](dag, lambda)
+	if len(downstreamVpcs) > 1 {
+		return fmt.Errorf("lambda %s has more than one vpc downstream", lambda.Id())
+	}
+	if len(upstreamVpcs) > 0 && len(downstreamVpcs) == 1 {
+		return fmt.Errorf("lambda %s has an upstream vpc upstream and cannot be associated with a downstream vpc", lambda.Id())
+	}
```
jhsinger-klotho, line 114: i dont think we need to care about upstream vpcs here, not sure why we would have them but they shouldnt quite matter

pkg/provider/aws/resources/lambda.go
```diff
@@ -103,6 +103,44 @@ func (lambda *LambdaFunction) MakeOperational(dag *core.ResourceGraph, appName s
...
+		// remove the direct dependency since the engine will be confused
+		// by the presence of the following valid paths from lambda -> vpc:
+		//   - lambda -> subnet -> vpc
+		//   - lambda -> security group -> vpc
+		vpc := downstreamVpcs[0]
+		err = dag.RemoveDependency(lambda.Id(), vpc.Id())
+		if err != nil {
+			return err
+		}
+	}
```
jhsinger-klotho, line 142: just thinking, what if someone says i want a lambda in subnet 1, but the subnet 1 hasnt been made operational yet, so we dont know that it is in a vpc.

Maybe the correct answer is that we need to run make operational in topological order

PR #698, opened 2023-07-05
K8s

jhsinger-klotho:
closes https://github.com/klothoplatform/klotho-pro/issues/94

adds k8s resources and their corresponding eks edges

Adds an OperationalResourceError type to be returned on makeoperational functions, which the engine can attempt to reconcile

### Standard checks


pkg/core/errors.go
```diff
@@ -26,12 +26,42 @@ type (
...
+	}
+}
+
+func (err *OperationalResourceError) Error() string {
+	return fmt.Sprintf("error in making resource %s operational: %v", err.Resource.Id(), err.Cause)
+}
+
+func (err *OperationalResourceError) Format(s fmt.State, verb rune) {
+	if formatter, ok := err.Cause.(fmt.Formatter); ok {
+		formatter.Format(s, verb)
```
gordon-klotho, line 55: nit: shouldn't this also put in `"error in making resource %s operational"`?

pkg/engine/engine.go
```diff
@@ -279,11 +285,21 @@ func (e *Engine) ExpandConstructsAndCopyEdges() error {
 					break
 				}
 			}
-			mappedResources, err := e.Provider.ExpandConstruct(construct, e.Context.WorkingState, e.Context.EndState, constructType, attributes)
-			if err != nil {
-				joinedErr = errors.Join(joinedErr, fmt.Errorf("unable to expand construct %s, %s", res.Id(), err.Error()))
+			var expandError error
+			for _, provider := range e.Providers {
```
gordon-klotho, line 289: Should this iterate in a deterministic order?

pkg/knowledge_base/edge_kb.go
```diff
@@ -78,6 +79,20 @@ const (
 	Downstream Reuse = "downstream"
 )
 
+func MergeKBs(kbsToUse []EdgeKB) (EdgeKB, error) {
+	kb := EdgeKB{}
+	var err error
+	for _, currKb := range kbsToUse {
+		for edge, detail := range currKb {
+			if _, found := kb[edge]; found {
+				err = errors.Join(err, fmt.Errorf("edge for %s -> %s is already defined in the knowledge base", edge.Source, edge.Destination))
```
gordon-klotho, line 88: Do `EdgeKB` have a name? Would be nice when we have more than just 2 to have that in the error message.

gordon-klotho (approved):
I didn't look too closely at the k8s specific stuff. The general core/engine stuff lgtm

pkg/knowledge_base/edge_kb.go
jhsinger-klotho: yeah ill add to error message

jhsinger-klotho: they dont have names actually

gordon-klotho: Yeah, I figured - when we add the next one (so there's > 2) we should add a name so we can get that in the error.

PR #700, opened 2023-07-06
EKS Dataflow Visualization

DavidSeptimus-Klotho:
This PR adds support for Kubernetes Pods, Deployments, and Helm Charts to the dataflow view.

Additional changes include fixes that should eliminate intermittent inclusion of resources we don't care about in the dataflow view as well as some intermittent parent/child relationship issues.

Also included are new Klotho CLI flags that were mistakenly dropped in a previous PR.

### DIFF
--- a/pkg/cli/klothomain.go
+++ b/pkg/cli/klothomain.go
@@ -2,8 +2,10 @@ package cli
 
 import (
 	"fmt"
+	"github.com/klothoplatform/klotho/pkg/filter"
 	"os"
 	"regexp"
+	"strings"
 	"time"
 
 	"github.com/fatih/color"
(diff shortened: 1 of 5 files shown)


pkg/lang/javascript/aws_runtime/_emitter.ts
```diff
@@ -22,11 +22,7 @@ export class Emitter extends events.EventEmitter {
     private client: SNSClient
     private s3: S3Client
 
-    constructor(
-        private path: string,
-        private name: string,
-        private id: string
-    ) {
+    constructor(private path: string, private name: string, private id: string) {
```
gordon-klotho, line 25: I think you need to update your prettier - I believe Jordan ran into this failing the check in GH action earlier.

gordon-klotho (approved):

PR #701, opened 2023-07-17
Remove aws value and change IaCValue to struct and use Ids

gordon-klotho:
### Standard checks


jhsinger-klotho (dismissed):

jhsinger-klotho:
unit tests, etc then will approve

PR #707, opened 2023-07-27
fix replacement operation

jhsinger-klotho:
### Standard checks

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -170,6 +170,7 @@ func (e *Engine) Run() (*core.ResourceGraph, error) {
 		return nil, fmt.Errorf(errorString)
 	}
 	zap.S().Debugf("found %d valid graphs", numValidGraphs)
+
 	return e.Context.Solution, nil
 }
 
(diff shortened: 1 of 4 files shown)


pkg/engine/engine.go
```diff
@@ -388,23 +389,47 @@ func (e *Engine) ApplyApplicationConstraint(constraint *constraints.ApplicationC
 		if construct == nil {
 			return fmt.Errorf("construct, %s, does not exist", construct.Id())
 		}
-		new, err := e.getConstructFromInputId(constraint.ReplacementNode)
-		if err != nil {
-			return err
+		var new core.BaseConstruct
```
gordon-klotho, line 392: `new` is a keyword in Go, probably best to pick a different name (or add more to it)

pkg/engine/engine.go
```diff
@@ -388,23 +389,47 @@ func (e *Engine) ApplyApplicationConstraint(constraint *constraints.ApplicationC
...
+			if err != nil {
+				return err
+			}
+		} else {
+			provider := e.Providers[constraint.ReplacementNode.Provider]
+			new, err = provider.CreateResourceFromId(constraint.ReplacementNode, e.Context.InitialState)
+			if err != nil {
+				return err
+			}
 		}
```
gordon-klotho, line 405: This seems like it could be common logic in `e.getConstructFromInputId`

pkg/engine/engine.go
```diff
@@ -388,23 +389,47 @@ func (e *Engine) ApplyApplicationConstraint(constraint *constraints.ApplicationC
...
 		for _, up := range upstream {
 			_ = e.deleteConstruct(up, false, false)
 		}
-		downstream := e.Context.WorkingState.GetDownstreamConstructs(construct)
 		for _, down := range downstream {
 			_ = e.deleteConstruct(down, false, false)
 		}
+		e.Context.WorkingState.AddConstruct(new)
+		for _, up := range upstream {
+			if e.Context.WorkingState.GetConstruct(up.Id()) == nil {
```
gordon-klotho, line 421: Why only constructs? Shouldn't all dependencies get moved over?

PR #730, opened 2023-09-05
add decisions

jhsinger-klotho:
ability to record decisions and engine failures for output

### Standard checks

### DIFF
--- a/pkg/engine/decisions.go
+++ b/pkg/engine/decisions.go
@@ -2,6 +2,7 @@ package engine
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/klothoplatform/klotho/pkg/construct"
 	"github.com/klothoplatform/klotho/pkg/engine/constraints"
@@ -29,9 +30,18 @@ type (
 	DecisionResult struct {
 		Resource construct.Resource
 		Edge     *graph.Edge[construct.Resource]
-		Config   *knowledgebase.Configuration
+		Config   *knowledgebase.ConfigurationRule
 	}
 	Level string
+
+	OutputDecision struct {
+		Cause          Cause
+		Resources      []construct.ResourceId
+		Edges          []construct.OutputEdge
+		Config         []knowledgebase.ConfigurationRule
+		CauseMessage   string
+		DisplayMessage string
+	}
 )
 
 const (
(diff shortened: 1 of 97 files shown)


pkg/engine/decisions.go
```diff
@@ -0,0 +1,226 @@
...
+	DecisionResult struct {
+		Resource construct.Resource
+		Edge     *graph.Edge[construct.Resource]
+		Config   *knowledgebase.Configuration
+	}
+	Level string
+)
+
+const (
+	// LevelError is the level of a decision when an error occurs
```
gordon-klotho, line 38: Not really a useful godoc since this is apparent from the name. Instead, maybe describe what an "error" means and what circumstances to use that instead of one of the other levels.

pkg/engine/decisions.go
```diff
@@ -0,0 +1,226 @@
...
+
+func (context *SolveContext) recordDecision(decision Decision) {
+	// Right now this is a wrapper around append, but we may want to do more in the future
+	//
+	// For example, we may want to check if there are duplicate decisions in the stack/history of decisions. If so we can likely determine that we are looping or going to fail and can shortcuircit the operations which are in a loop.
+	context.Decisions = append(context.Decisions, decision)
+}
+
+// Here we want to validate that each decision is crafted correctly and that there is a valid cause, action and result pairing
+func (d Decision) Validate() error {
```
gordon-klotho, line 62: Personally, I don't see the use in validating structs that we're creating internally (ie, not user-input based). Unit tests should catch any invalid decisions.

pkg/engine/decisions.go
```diff
@@ -0,0 +1,226 @@
...
+		Action Action
+		Cause  *Cause
+	}
+	Action string
+	Cause  struct {
+		EdgeExpansion         *graph.Edge[construct.Resource]
+		EdgeConfiguration     *graph.Edge[construct.Resource]
+		OperationalResource   construct.Resource
+		ResourceConfiguration construct.Resource
+		ConstructExpansion    construct.BaseConstruct
```
gordon-klotho, line 26: These feel weird being actual Resource/Constructs instead of Ids.

pkg/engine/decisions.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+}
+
+// func (d Decision) MarshalJSON() ([]byte, error) {
+// 	fmt.Println(`{"level":"` + string(d.Level) + `","result":"` + d.Result.String() + `","action":"` + string(d.Action) + `","cause":"` + d.Cause.String() + `"}`)
+// 	return []byte(`{"level":"` + string(d.Level) + `","result":"` + d.Result.String() + `","action":"` + string(d.Action) + `","cause":"` + d.Cause.String() + `"}`), nil
+// }
+
+func (r DecisionResult) MarshalJSON() ([]byte, error) {
+	if r.Config != nil {
```
gordon-klotho, line 185: Use an internal type for these, should make it much cleaner. Though if the actual type used `ResourceId`s then this may not even be necessary.

PR #731, opened 2023-09-07
Support high-level configuration via templates

gordon-klotho:
### Standard checks


pkg/provider/aws/resources/templates/api_integration.yaml
```diff
@@ -18,8 +18,27 @@ rules:
...
+        value: '{{ .Value }}'
+      - object: '{{ self }}'
+        property: RequestParameters
+        value: |
+          {{ $params := split "/" .Value | filterMatch "^:\\w+$" }}
+          {{ zipToMap
+            ($params | mapString ":(.*)" "integration.request.path.$1")
+            ($params | mapString ":(.*)" "method.request.path.$1")
+          }}
+
```
jhsinger-klotho, line 40: how are we going to handle a configuration if it changes based on what is upstream of it? Before we did edge config local to the edges so it only happened based on what existed in the graph, where as now i dont see how that is possible without having a bunch of conditional statements in here. For example, the environment variables one im not quite sure how we are going to accomplish

PR #777, opened 2023-11-21
use consumption to influence topology

jhsinger-klotho:
uses the consumption to influence topology yaml for this point in time

### Standard checks

### DIFF
--- a/pkg/knowledge_base2/emitter.go
+++ b/pkg/knowledge_base2/emitter.go
@@ -226,9 +226,21 @@ func (c *ConsumptionObject) Emit(ctx DynamicValueContext, resource construct.Res
 }
 
 func (c *ConsumptionObject) Consume(val any, ctx DynamicValueContext, resource *construct.Resource) error {
-	err := resource.SetProperty(c.PropertyPath, val)
+	rt, err := ctx.KnowledgeBase.GetResourceTemplate(resource.ID)
 	if err != nil {
 		return err
+	}
+	prop := rt.GetProperty(c.PropertyPath)
+	if prop.IsPropertyTypeScalar() {
+		err := resource.SetProperty(c.PropertyPath, val)
+		if err != nil {
+			return err
+		}
+	} else {
+		err := resource.AppendProperty(c.PropertyPath, val)
+		if err != nil {
+			return err
+		}
 	}
 	return nil
 }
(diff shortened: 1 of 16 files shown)


pkg/knowledge_base2/emitter.go
```diff
@@ -109,9 +185,16 @@ func (c *ConsumptionObject) Convert(value any, res construct.ResourceId, ctx Dyn
...
+	// We convert here just to make sure it gets translated to the right type of input
+	// We will convert again when consuming to ensure strings/etc are converted to their respective struct
+	// if they match a property ref/id/etc
 	val, err := TransformToPropertyValue(res, c.PropertyPath, bstr, ctx, DynamicValueData{Resource: res})
-	if err == nil {
-		return val, nil
+	if err != nil {
+		return val, err
+	}
+	val, err = TransformToPropertyValue(res, c.PropertyPath, val, ctx, DynamicValueData{Resource: res})
```
gordon-klotho, line 195: The double convert is very weird, why doesn't `TransformtToPropertyValue` do this? I think it would violate the contract of what that function should do if it doesn't return a fully valid, transformed value.

What's the use-case for why this won't work in one shot with the current implementation?

pkg/knowledge_base2/property_types.go
```diff
@@ -545,3 +547,101 @@ func (r *ResourcePropertyType) ZeroValue() any {
 func (p *PropertyRefPropertyType) ZeroValue() any {
 	return construct.PropertyRef{}
 }
+
+func (m *MapPropertyType) Contains(value any, contains any) bool {
```
gordon-klotho, line 551: This is more of a subset than contains. Needs documentation too given that `func(any, any)` doesn't really give any signature-related hints to what should be passed in or valid types.

gordon-klotho (approved):

pkg/knowledge_base2/emitter.go
jhsinger-klotho: when we pass it a json string, it doesnt work in one shot. Thats what im running into now, because it converts the json string to map(string,string) but it doesnt evaluate the value to see if its a property ref or resource id. We can likely fix this function

gordon-klotho: Yeah, looks like it isn't doing the right thing in `MapPropertyType.Parse`. After decoding the string, it immediately returns instead of checking the sub properties.

PR #780, opened 2023-11-21
fix api gw dependencies

jhsinger-klotho:
fixes api gw deps

### Standard checks

### DIFF
--- a/pkg/engine2/operational_rule/operational_rule.go
+++ b/pkg/engine2/operational_rule/operational_rule.go
@@ -21,10 +21,6 @@ type (
 )
 
 func (ctx OperationalRuleContext) HandleOperationalRule(rule knowledgebase.OperationalRule) error {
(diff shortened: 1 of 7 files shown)


pkg/engine2/operational_rule/operational_rule.go
```diff
@@ -22,6 +22,9 @@ type (
 
 func (ctx OperationalRuleContext) HandleOperationalRule(rule knowledgebase.OperationalRule) error {
 
+	if ctx.Data.Edge != nil && ctx.Data.Edge.Source.String() == "aws:api_deployment:rest_api_0:api_deployment-0" {
+		fmt.Println("here")
+	}
```
gordon-klotho, line 27: Debug

gordon-klotho (approved):
Besides the leftover debug print

PR #784, opened 2023-11-27
Efs

jhsinger-klotho:
- fixes efs templates to work with lambda & ecs (eks still has remaining issue in installing eks add on for efs csi driver)

- fix bug in dependency capture to find config fields that are list indices

- fix spread placer due to AZ spread logic being inaccurate (only look at direct deps)

- introduce negative weights returned for candidate validity to influence edge weights 

ecs &lambda test results
```
============================= test session starts ==============================
platform darwin -- Python 3.10.11, pytest-7.4.3, pluggy-1.3.0
rootdir: /Users/jordansinger/workspace/engine-e2e-tests/output/aws-3b6en6
asyncio: mode=strict
created: 1/1 worker
1 worker [1 item]

.                                                                        [100%]
- generated xml file: /var/folders/n_/bbppnfk179d6l4vz4qmz0chh0000gn/T/tmpucara2jb -
============================== 1 passed in 0.91s ===============================
```

### Standard checks


pkg/engine2/operational_rule/resource_placer.go
```diff
@@ -219,12 +219,14 @@ func (ctx OperationalRuleContext) findNumConnectionsToTypeForAvailableResources(
 	for _, availableResource := range availableResources {
 		var err error
 		var connections []construct.ResourceId
+		// We will look to see what direct dependencies are already existing in the same direction as the rule
+		// if we dont only look at direct, we risk getting incorrect results if the resource can have non functional connections
```
gordon-klotho, line 223: Might not be necessary for a comment, but just for my knowledge, what "incorrect results" did this have that warrants the change?

pkg/engine2/path_selection/candidate_validity.go
```diff
@@ -247,31 +249,40 @@ func (d downstreamChecker) makeValid(resource, operationResource *construct.Reso
...
+				}
 			}
 			if p.IsPropertyTypeScalar() {
-				res.SetProperty(property, downstream)
+				currRes.SetProperty(property, downstream)
 				return true, errs
 			} else {
-				res.AppendProperty(property, downstream)
+				currRes.AppendProperty(property, downstream)
 				return true, errs
```
gordon-klotho, line 286: Feels like this could be simplified by adding `Append` to scalar types that just do `Set`.

pkg/engine2/path_selection/path_expansion.go
```diff
@@ -274,6 +274,20 @@ func ExpandPath(
 	for i, node := range nonBoundaryResources {
 		candidates[i] = make(map[construct.ResourceId]int)
 		candidates[i][node] = 0
+		resource, err := input.TempGraph.Vertex(node)
+		if err != nil {
+			errs = errors.Join(errs, err)
+			continue
+		}
+		// we know phantoms are always valid, but we want to ensure we make them valid based on src and target validity checks
```
gordon-klotho, line 282: Not sure what this means, the 2nd part seems to invalidate the 1st part.

pkg/engine2/path_selection/path_expansion.go
```diff
@@ -274,6 +274,20 @@ func ExpandPath(
...
+		valid, err := checkCandidatesValidity(ctx, resource, path, input.Classification)
+		if err != nil {
+			errs = errors.Join(errs, err)
+			continue
+		}
+		if !valid {
+			candidates[i][node] = -1000
```
gordon-klotho, line 289: If it's not valid, shouldn't it just be removed? Not-valid should be the same as not-preferred.

pkg/engine2/path_selection/path_selection.go
```diff
@@ -194,14 +193,30 @@ func calculateEdgeWeight(
 	// This will achieve priority for existing resources over newly created ones
 	weight := 0
 	if kb.GetFunctionality(source) != knowledgebase.Unknown && !source.Matches(dep.Source) {
-		weight += (FUNCTIONAL_WEIGHT / divideSourceBy)
+		if divideSourceBy > 0 {
+			weight += (FUNCTIONAL_WEIGHT / divideSourceBy)
+		} else if divideSourceBy < 0 {
+			weight += (FUNCTIONAL_WEIGHT * divideSourceBy * -1)
```
gordon-klotho, line 199:
```suggestion
			weight -= (FUNCTIONAL_WEIGHT * divideSourceBy)
```
nit

pkg/templates/aws/resources/rds_instance.yaml
```diff
@@ -78,6 +78,8 @@ consumption:
       value:
         '{{ .Self.Name }}_RDS_ENDPOINT': '{{ fieldRef "Endpoint" .Self }}'
         '{{ .Self.Name }}_RDS_CONNECTION_ARN': '{{ fieldRef "RdsConnectionArn" .Self }}'
+        '{{ .Self.Name }}_RDS_USERNAME': '{{ fieldRef "Username" .Self }}'
+        '{{ .Self.Name }}_RDS_PASSWORD': '{{ fieldRef "Password" .Self }}'
```
gordon-klotho, line 82: note: `.Self.Name` may not be a valid environment variable name

pkg/engine2/operational_rule/resource_placer.go
jhsinger-klotho: subnets in AZs, sometimes with EFS since the mount target isnt functional we see multiple subnets upstream of both AZs even though their direct dep and property are right

pkg/engine2/path_selection/candidate_validity.go
```diff
@@ -132,8 +128,13 @@ func checkAsTargetValidity(
 				resources, err = solution_context.GetResourcesFromPropertyReference(ctx,
 					resource.ID, ps.PropertyReference)
 				if err != nil {
+					// dont return error because it just means that the property isnt set and we can make the
+					// resource valid
+					zap.S().Debug("no resource available from resource %s from property ref %s", resource.ID, ps.PropertyReference)
```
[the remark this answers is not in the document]
jhsinger-klotho: dont think it needs to but it could for debugability. I just dont want it to cause compiler failures in the future if we fail on error

pkg/engine2/path_selection/candidate_validity.go
jhsinger-klotho: yeah could add that as well if we want

pkg/engine2/path_selection/path_expansion.go
jhsinger-klotho: will reword. all phantoms are not inherently valid, but can be made valid. So we want to make sure they are weighted less than a valid reusable candidates. Probably should rename the method since its not actually marking things as invalid yet, just as helping determine correct weights

PR #792, opened 2023-11-27
add alb support

jhsinger-klotho:
adds support for configuring load balancers as application load balancers

- ensures we remove dependencies when clearing properties
- polls the ready vertices in operational eval in case the ready vertices in teh list are deleted or mutated during an earlier evaluation

### Standard checks

### DIFF
--- a/pkg/templates/aws/edges/listener_certificate-acm_certificate.yaml
+++ b/pkg/templates/aws/edges/listener_certificate-acm_certificate.yaml
@@ -1,0 +1,11 @@
+source: aws:listener_certificate
+target: aws:acm_certificate
+operational_rules:
+  - configuration_rules:
+      - resource: '{{ .Source }}'
+        configuration:
+          field: Certificate
+          value: '{{ .Target }}'
+          
+unique:
+  source: true
--- a/pkg/templates/aws/edges/listener_certificate-acm_certificate.yaml
+++ b/pkg/templates/aws/edges/listener_certificate-acm_certificate.yaml
@@ -1,11 +1,5 @@
 source: aws:listener_certificate
 target: aws:acm_certificate
-operational_rules:
-  - configuration_rules:
-      - resource: '{{ .Source }}'
-        configuration:
-          field: Certificate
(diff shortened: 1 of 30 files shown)


pkg/templates/aws/resources/load_balancer_listener_rule.yaml
```diff
@@ -0,0 +1,31 @@
...
+    type: list
+    properties:
+      PathPattern:
+        type: list(string)
+        default_value: path-pattern
+      HttpRequestMethod:
+        type: list(string)
+  Listener:
+    type: resource(aws:load_balancer_listener)
+    default_value: '{{ upstream "aws:load_balancer" .Self }}'
```
DavidSeptimus-Klotho, line 24: Wouldn't this make the default value the upstream load balancer rather than a listener?

pkg/templates/aws/edges/listener_certificate-acm_certificate.yaml
```diff
@@ -0,0 +1,11 @@
+source: aws:listener_certificate
+target: aws:acm_certificate
+operational_rules:
+  - configuration_rules:
+      - resource: '{{ .Source }}'
+        configuration:
+          field: Certificate
+          value: '{{ .Target }}'
```
gordon-klotho, line 8: Is it weird to have this as a edge config rule instead of an operational rule?

gordon-klotho (approved):
Didn't look too closely at templates, tests should verify those work. Code looks good.

pkg/templates/aws/resources/load_balancer_listener_rule.yaml
jhsinger-klotho: yeah, i didnt think i included this file since its not used yet, miss on my part

pkg/templates/aws/edges/listener_certificate-acm_certificate.yaml
jhsinger-klotho: can move it over

PR #798, opened 2023-11-30
Properties

jhsinger-klotho:
starts building internal types for properties
- a property struct for each type of property we support
- a central interface in the kb used elsewhere

A reader for templates so we can convert them from expected yaml to internal types (WIP)

splits out operational rule from property rule (will further split this out into edge rule in a following change)

### Standard checks


pkg/construct2/resource_id.go
```diff
@@ -67,6 +67,15 @@ func (l *ResourceList) UnmarshalText(b []byte) error {
 	return fmt.Errorf("could not unmarshal resource list: %s", string(b))
 }
 
+func (l *ResourceList) MatchesAny(id ResourceId) bool {
+	for _, rid := range *l {
```
gordon-klotho, line 71:
```suggestion
func (l ResourceList) MatchesAny(id ResourceId) bool {
	for _, rid := range l {
```

pkg/knowledge_base2/properties/property.go
```diff
@@ -0,0 +1,74 @@
...
+	PropertyValidityCheck struct {
+		template *template.Template
+	}
+	ValidityCheckData struct {
+		Properties construct.Properties
+		Value      any
+	}
+)
+
+func ParsePropertyRef(value any, ctx knowledgebase.DynamicContext, data knowledgebase.DynamicValueData) (any, error) {
```
gordon-klotho, line 27: Shouldn't this always return a `PropertyRef`, not `any`?

pkg/knowledge_base2/properties/resource_property.go
```diff
@@ -0,0 +1,164 @@
...
+	knowledgebase "github.com/klothoplatform/klotho/pkg/knowledge_base2"
+)
+
+type (
+	ResourceProperty struct {
+		DefaultValue *construct.ResourceId  `json:"default_value" yaml:"default_value"`
+		AllowedTypes construct.ResourceList `yaml:"allowed_types"`
+		Namespace    bool                   `yaml:"namespace"`
+		SharedPropertyFields
+		*knowledgebase.PropertyDetails
```
gordon-klotho, line 17: Embedded fields that are pointers are a little weird and could more easily lead to nil-pointer access. Is there a reason this needs to be a pointer?

pkg/knowledge_base2/properties/resource_property.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+)
+
+func (r *ResourceProperty) SetProperty(resource *construct.Resource, value any) error {
+	if val, ok := value.(construct.ResourceId); ok {
```
gordon-klotho, line 22: nit: prefer a type switch. Also maybe support `*construct.Resource` even though we don't currently use it.

pkg/knowledge_base2/properties/resource_property.go
```diff
@@ -0,0 +1,164 @@
...
+
+func (r *ResourceProperty) RemoveProperty(resource *construct.Resource, value any) error {
+	delete(resource.Properties, r.Path)
+	return nil
+}
+
+func (r *ResourceProperty) Details() *knowledgebase.PropertyDetails {
+	return r.PropertyDetails
+}
+func (r *ResourceProperty) Clone() knowledgebase.Property {
```
gordon-klotho, line 42: If `PropertyDetails` isn't a pointer, you don't really need `Clone`, because everything is by-value.
This would clone the object:
```go
var original ResourceProperty
cloned := original
```

pkg/knowledge_base2/resource_template.go
```diff
@@ -38,33 +34,42 @@ type (
...
+		// Path is the path to the property in the resource
 		Path string `json:"-" yaml:"-"`
 	}
 
+	Property interface {
+		SetProperty(resource *construct.Resource, value any) error
+		AppendProperty(resource *construct.Resource, value any) error
+		RemoveProperty(resource *construct.Resource, value any) error
+		Details() *PropertyDetails
+		Clone() Property
```
gordon-klotho, line 61: I feel like we should be able to get away w/o `Clone`, but that could be a followup.

pkg/knowledge_base2/resource_template.go
```diff
@@ -38,33 +34,42 @@ type (
...
+		OperationalRule *PropertyRule `json:"operational_rule" yaml:"operational_rule"`
+		// Path is the path to the property in the resource
 		Path string `json:"-" yaml:"-"`
 	}
 
+	Property interface {
+		SetProperty(resource *construct.Resource, value any) error
+		AppendProperty(resource *construct.Resource, value any) error
+		RemoveProperty(resource *construct.Resource, value any) error
+		Details() *PropertyDetails
```
gordon-klotho, line 60: This returning a pointer makes me think that this is mutable. Is that intentional?

pkg/knowledge_base2/resource_template.go
```diff
@@ -38,33 +34,42 @@ type (
...
+		Clone() Property
+		Type() string
+		GetDefaultValue(ctx DynamicValueContext, data DynamicValueData) (any, error)
+		Validate(value any, properties construct.Properties) error
+		SubProperties() map[string]Property
```
gordon-klotho, line 65: This can return a `Properties` now that this type's not tied to the yaml parsing, right?

pkg/knowledge_base2/reader/properties.go
```diff
@@ -0,0 +1,227 @@
...
+		itemProp, err := getPropertyType(val)
+		if err != nil {
+			return nil, err
+		}
+		return &properties.SetProperty{ItemProperty: itemProp}, nil
+	},
+	"any": func(val string) (knowledgebase.Property, error) { return &properties.AnyProperty{}, nil },
+}
+
+func getPropertyType(ptype string) (knowledgebase.Property, error) {
```
gordon-klotho, line 205: This doesn't seem necessary, can just call `InitializeProperty` instead, right?

pkg/knowledge_base2/properties/property.go
jhsinger-klotho: yeah changing this

pkg/knowledge_base2/properties/resource_property.go
jhsinger-klotho: flipped we chatted about this

pkg/knowledge_base2/properties/resource_property.go
jhsinger-klotho: yeah, im first going for parity, then we can beef these up and make them more robust

pkg/knowledge_base2/properties/resource_property.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	return fmt.Errorf("invalid resource value %v", value)
+}
+
+func (r *ResourceProperty) AppendProperty(resource *construct.Resource, value any) error {
+	return r.SetProperty(resource, value)
+}
+
+func (r *ResourceProperty) RemoveProperty(resource *construct.Resource, value any) error {
+	delete(resource.Properties, r.Path)
```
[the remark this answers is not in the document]
jhsinger-klotho: yeah for things like sets/lists you could remove an item

will change the method for these though like you mentioned it might not work for nested ones

pkg/knowledge_base2/properties/resource_property.go
jhsinger-klotho: in the case where theyre pointers though you want to dereference and then do that?

pkg/knowledge_base2/properties/resource_property.go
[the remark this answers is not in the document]
gordon-klotho: Isn't that what `resource.RemoveProperty(path, optionalKeyOrElement)` does already?

pkg/knowledge_base2/operational_rule.go
```diff
@@ -16,6 +16,23 @@ type (
...
 
+	EdgeRule struct {
+		If                 string                `json:"if" yaml:"if"`
+		Steps              []EdgeOperationalStep `json:"steps" yaml:"steps"`
+		ConfigurationRules []ConfigurationRule   `json:"configuration_rules" yaml:"configuration_rules"`
+	}
+
+	PropertyRule struct {
+		If    string           `json:"if" yaml:"if"`
+		Step  *OperationalStep `json:"step" yaml:"step"`
```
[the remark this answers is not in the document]
jhsinger-klotho: yeah removing the resource is coming in the next refactor where we introduce edgeStep thats just resource + operational step.

i was doing a pointer to tell if it exists or not, rather than checking against OperationalStep{} each time

pkg/knowledge_base2/resource_template.go
jhsinger-klotho: i can try to remove it

pkg/knowledge_base2/resource_template.go
```diff
@@ -38,33 +34,42 @@ type (
...
-
-		Properties map[string]*Property `json:"properties" yaml:"properties"`
-
+		// OperationalRule defines a rule that is executed at runtime to determine the value of the property
+		OperationalRule *PropertyRule `json:"operational_rule" yaml:"operational_rule"`
+		// Path is the path to the property in the resource
 		Path string `json:"-" yaml:"-"`
 	}
 
+	Property interface {
```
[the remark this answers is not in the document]
jhsinger-klotho: yup will add

pkg/knowledge_base2/resource_template.go
```diff
@@ -38,33 +34,42 @@ type (
...
+		RemoveProperty(resource *construct.Resource, value any) error
+		Details() *PropertyDetails
+		Clone() Property
+		Type() string
+		GetDefaultValue(ctx DynamicValueContext, data DynamicValueData) (any, error)
+		Validate(value any, properties construct.Properties) error
+		SubProperties() map[string]Property
+		Parse(value any, ctx DynamicContext, data DynamicValueData) (any, error)
+		ZeroValue() any
+		Contains(value any, contains any) bool
```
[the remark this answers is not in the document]
jhsinger-klotho: yeah but the issue is that when you call it on the resource, we dont get the property from the KB and therefore dont necessarily know what logic to carry out. 

I agree it kind of makes more sense on the resource, but because the property struct understands more about what can exist within the property, we need to be able to reference that and cant pass the kb to the construct package

pkg/knowledge_base2/reader/properties.go
jhsinger-klotho: then go complains about a possible endless loop, so this was my workaround. Is there a way for go to like methods calling eachother?

pkg/knowledge_base2/resource_template.go
```diff
@@ -350,7 +255,16 @@ func (template ResourceTemplate) GetProperty(name string) *Property {
 
 var ErrStopWalk = errors.New("stop walk")
 
-func (tmpl ResourceTemplate) LoopProperties(res *construct.Resource, addProp func(*Property) error) error {
+// ReplacePath runs a simple [strings.ReplaceAll] on the path of the property and all of its sub properties.
+// NOTE: this mutates the property, so make sure to [Property.Clone] it first if you don't want that.
+func ReplacePath(p Property, original, replacement string) {
```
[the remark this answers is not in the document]
jhsinger-klotho: Property is the interface so the underlying struct should be a pointer but could modify this

pkg/knowledge_base2/properties/resource_property.go
gordon-klotho: yeah, you could
```go
var original *ResourceProperty
cloned := *original
use(&cloned)
```

pkg/knowledge_base2/operational_rule.go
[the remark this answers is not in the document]
gordon-klotho: I wouldn't check against `OperationalStep{}` because that allocates. I'd just do the length check in my original comment. Generally I try to avoid pointers unless there's a good reason to, like if copying the value should share the field. I think we _don't_ want that behaviour so that'd be a good reason not to use a pointer.

This is a nit though, non-blocking

pkg/knowledge_base2/reader/properties.go
gordon-klotho: You can have recursive functions, or mutually recursive functions in Go. What was the specific error you were getting?

pkg/knowledge_base2/resource_template.go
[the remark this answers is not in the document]
gordon-klotho: Ah, good point. It also uses `.Details()` which is a pointer that it modifies, so I'm double incorrect 😄

pkg/knowledge_base2/properties/resource_property.go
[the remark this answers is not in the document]
jhsinger-klotho: i guess it depends on if we want validation around the value theyre passing in or anything. Likely should be fine if its all already handled

pkg/knowledge_base2/properties/resource_property.go
jhsinger-klotho: Yeah but what about the case where we have to clone a property without knowing its type? I dont think theres a proper way to do that. Are you proposing still having a clone method, just doing it that way?

pkg/knowledge_base2/resource_template.go
jhsinger-klotho: it is an engine operation in the case where we want to inspect properties nested within arrays based on the way we currently do it

pkg/knowledge_base2/reader/properties.go
```diff
@@ -0,0 +1,267 @@
...
+
+		AllowedTypes construct.ResourceList `yaml:"allowed_types"`
+
+		SanitizeTmpl  string   `yaml:"sanitize"`
+		AllowedValues []string `yaml:"allowed_values"`
+
+		KeyProperty   knowledgebase.Property `yaml:"key_property"`
+		ValueProperty knowledgebase.Property `yaml:"value_property"`
+
+		ItemProperty knowledgebase.Property `yaml:"item_property"`
```
gordon-klotho, line 51: What are these for? They seem like they're computed but the yaml tag suggests otherwise. (check other fields too)

pkg/construct2/properties.go
```diff
@@ -432,6 +432,23 @@ func (i *mapValuePathItem) Remove(value any) (err error) {
 		arr = arr.Elem()
 	}
 	if arr.Kind() != reflect.Slice && arr.Kind() != reflect.Array {
+		if hs, ok := arr.Interface().(set.HashedSet[string, any]); ok {
+			if hs.Contains(value) {
+				removed := hs.Remove(value)
+				if !removed {
```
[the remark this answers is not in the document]
jhsinger-klotho: why does Remove return a bool then? i assume false would mean it failed to remove it

pkg/engine2/operational_eval/vertex_property.go
```diff
@@ -35,21 +35,21 @@ func (prop *propertyVertex) Dependencies(eval *Evaluator) (graphChanges, error)
 
 	// Template can be nil when checking for dependencies from a propertyVertex when adding an edge template
 	if prop.Template != nil {
-		propCtx.ExecuteValue(prop.Template.DefaultValue, resData)
-
-		if opRule := prop.Template.OperationalRule; opRule != nil {
-			if err := propCtx.ExecuteOpRule(resData, *opRule); err != nil {
+		_, _ = prop.Template.GetDefaultValue(propCtx.inner, resData)
```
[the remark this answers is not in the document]
jhsinger-klotho: ah it was because i was using DynamicContextValue and not DynamicContext interface in the method signature. Switched it

pkg/engine2/operational_eval/vertex_property.go
```diff
@@ -160,11 +160,16 @@ func (v *propertyVertex) evaluateConstraints(sol solution_context.SolutionContex
 		return fmt.Errorf("could not get current value for %s: %w", v.Ref, err)
 	}
 
-	if currentValue == nil && setConstraint.Operator == "" && v.Template != nil && v.Template.DefaultValue != nil {
+	ctx := solution_context.DynamicCtx(sol)
+	defaultVal, err := v.Template.GetDefaultValue(ctx, dynData)
```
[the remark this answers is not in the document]
jhsinger-klotho: not currently, we could add a method for that, although it wont actually speed up the performance because the first check in the computation is just does the field == nil so it would essentially be doing the same thing

pkg/engine2/operational_eval/vertex_property.go
```diff
@@ -282,21 +281,26 @@ func (v *propertyVertex) Ready(eval *Evaluator) (ReadyPriority, error) {
...
 		return NotReadyMax, nil
 	}
-	if v.Template.OperationalRule != nil {
+	if v.Template.Details().OperationalRule != nil {
 		// operational rules should run as soon as possible to create any resources they need
 		return ReadyNow, nil
 	}
-	if strings.Contains(v.Template.Type, "list") || strings.Contains(v.Template.Type, "set") {
+	ptype := v.Template.Type()
+	if strings.Contains(ptype, "list") || strings.Contains(ptype, "set") {
```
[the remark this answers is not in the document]
jhsinger-klotho: 👍 thought i fixed this, but was a different line to where i made that change

pkg/engine2/operational_rule/operational_step.go
```diff
@@ -254,65 +254,68 @@ func (ctx OperationalRuleContext) SetField(resource, fieldResource *construct.Re
...
-				if !ok {
-					continue
-				}
-				if !currResId.IsZero() && currResId == fieldResource.ID {
-					return nil
-				}
+		return nil
+	}
+
+	removeCurrentValue := func(res any) error {
```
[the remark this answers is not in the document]
jhsinger-klotho: yeah, made the change

pkg/knowledge_base2/reader/embed.go
```diff
@@ -1,23 +1,33 @@
...
 	if err != nil {
 		return nil, err
 	}
-	kb.models = kbModels
-	templates, err := TemplatesFromFs(resources, kbModels)
+	kbModels := map[string]*knowledgebase.Model{}
+	for name, model := range readerModels {
+		kbModel, err := model.Convert()
+		if err != nil {
+			errs = errors.Join(errs, fmt.Errorf("error converting model %s: %w", name, err))
```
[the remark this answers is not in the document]
jhsinger-klotho: errs is returned at the end, but can add a check for after this since we know templates will fail if models fail

pkg/knowledge_base2/reader/properties.go
jhsinger-klotho: if you want your item/key/value to be able to have validation along with it (the strings in my list need to match a regex) these would be where you would define that

pkg/templates/aws/resources/api_deployment.yaml
```diff
@@ -6,10 +6,10 @@ properties:
     type: resource(aws:rest_api)
     namespace: true
     operational_rule:
-      steps:
-        - direction: downstream
-          resources:
-            - aws:rest_api
+      step:
```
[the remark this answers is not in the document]
jhsinger-klotho: agreed, but i was going to refactor these further after this so we could split out edge/resource things

pkg/knowledge_base2/reader/properties.go
```diff
@@ -0,0 +1,267 @@
...
+	return propertyType, nil
+}
+
+func setChildPaths(property *Property, currPath string) {
+	for name, child := range property.Properties {
+		child.Name = name
+		path := currPath + "." + name
+		child.Path = path
+		setChildPaths(child, path)
+		property.Properties[name] = child
```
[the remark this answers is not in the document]
jhsinger-klotho: 👍

pkg/knowledge_base2/resource_template.go
```diff
@@ -19,59 +15,101 @@ import (
...
+		// Details returns the property details for the property
+		Details() *PropertyDetails
+		// Clone returns a clone of the property
+		Clone() Property
+		// Type returns the string representation of the type of the property, as it should appear in the resource template
+		Type() string
+		// GetDefaultValue returns the default value for the property, pertaining to the specific data being passed in for execution
+		GetDefaultValue(ctx DynamicValueContext, data DynamicValueData) (any, error)
+		// Validate ensures the value is valid for the property and returns an error if it is not
+		Validate(value any, properties construct.Properties) error
```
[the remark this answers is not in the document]
jhsinger-klotho: yeah it should likely be the resource and then value.

pkg/construct2/properties.go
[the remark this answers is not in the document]
gordon-klotho: It does, but failed to remove only happens if it doesn't contain. It returns a bool (many set/map operations do this) specifically to simplify this types of conditions.

pkg/knowledge_base2/reader/properties.go
gordon-klotho: Doesn't seem like these should be full `knowledgebase.Property` then, since it's just validation and none of the other fields should be specified (ie, can a `key_property` have sub properties?).

pkg/knowledge_base2/reader/properties.go
```diff
@@ -97,45 +100,46 @@ func (p *Property) Convert() (knowledgebase.Property, error) {
...
+		if dstField.Type() == srcField.Type() {
+			dstField.Set(srcField)
+		} else {
+			if conversion, found := fieldConversion[fieldName]; found {
+				err := conversion(srcField, p, propertyType)
+				if err != nil {
+					return nil, err
 				}
+			} else {
+				return nil, fmt.Errorf("invalid property type %s", fieldName)
```
gordon-klotho, line 142: Maybe add more info about the types that don't match as well.

pkg/knowledge_base2/properties/bool_property_test.go
```diff
@@ -297,7 +297,8 @@ func Test_BoolPropertyValidate(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			assert := assert.New(t)
-			err := tt.property.Validate(tt.value, construct.Properties{})
+			resource := &construct.Resource{}
+			err := tt.property.Validate(resource, tt.value)
```
gordon-klotho, line 301: nit: since resource isn't used for bool validate, can just simply `tt.property.Validate(nil, tt.value)`

pkg/knowledge_base2/resource_template.go
```diff
@@ -81,7 +81,7 @@ type (
 		// GetDefaultValue returns the default value for the property, pertaining to the specific data being passed in for execution
 		GetDefaultValue(ctx DynamicContext, data DynamicValueData) (any, error)
 		// Validate ensures the value is valid for the property and returns an error if it is not
-		Validate(value any, properties construct.Properties) error
+		Validate(resource *construct.Resource, value any) error
```
gordon-klotho, line 84: Why does it need `value`, can't it get it from the resource from its path? What does it need the resource (was the Properties) for? As far as I could tell from this diff, it's unused.

jhsinger-klotho: if you want to validate a value before its being set you would need to pass in the value, so we can either make this only check the current property, but then we lose some prechecking ability

pkg/engine2/cli.go
```diff
@@ -150,6 +152,14 @@ func addSubProperties(properties map[string]any, subProperties map[string]knowle
 			"configurationDisabled": details.ConfigurationDisabled,
 			"required":              details.Required,
 		}
+		validationFields := []string{"MinLength", "MaxLength", "MinValue", "MaxValue", "AllowedValues"}
```
gordon-klotho, line 155: nit: make this global so it doesn't allocate a new array each call

pkg/knowledge_base2/reader/embed.go
```diff
@@ -94,6 +94,9 @@ func ModelsFromFS(dir fs.FS) (map[string]*Model, error) {
 func TemplatesFromFs(dir fs.FS, models map[string]*Model) (map[construct.ResourceId]*knowledgebase.ResourceTemplate, error) {
 	templates := map[construct.ResourceId]*knowledgebase.ResourceTemplate{}
 	err := fs.WalkDir(dir, ".", func(path string, d fs.DirEntry, nerr error) error {
+		if d.Name() == "lambda_function.yaml" {
+			zap.S().Debug("Skipping lambda function: ", path)
```
gordon-klotho, line 98: debug

pkg/knowledge_base2/resource_template.go
jhsinger-klotho: the properties (and now resource) is going to be needed for validity checks that are more complex (if field a is blah, this must be TCP, etc)

PR #805, opened 2023-12-06
validate fields for property refs and produce config errors repoty

jhsinger-klotho:
- Ensures each type validates that if set with a property ref, that ref is of the correct type and that refs value passes validation (if deploy time then it automatically passes if the correct type)

- produces a config_errors.json for failed config
- Going to add failure if the failed config matches something that was an input constraint

### Standard checks

### DIFF
--- a/pkg/engine2/solution_context/decisions.go
+++ b/pkg/engine2/solution_context/decisions.go
@@ -1,7 +1,7 @@
 package solution_context
 
 import (
-	"fmt"
+	"encoding/json"
 
 	construct "github.com/klothoplatform/klotho/pkg/construct2"
 	knowledgebase "github.com/klothoplatform/klotho/pkg/knowledge_base2"
@@ -70,18 +70,16 @@ func (d PropertyValidationDecision) internal() {}
 
 func (d PropertyValidationDecision) MarshalJSON() ([]byte, error) {
 	if d.Value != nil {
-		stringVal := `{
-			"resource": "%s",
-			"property": "%s",
-			"value": "%s",
-			"error": "%s"
-		}`
-		return []byte(fmt.Sprintf(stringVal, d.Resource, d.Property.Details().Path, d.Value, d.Error)), nil
+		return json.Marshal(map[string]any{
+			"resource": d.Resource,
+			"property": d.Property.Details().Path,
+			"value":    d.Value,
+			"error":    d.Error,
+		})
 	}
-	stringVal := `{
-		"resource": "%s",
-		"property": "%s",
-		"error": "%s"
-	}`
-	return []byte(fmt.Sprintf(stringVal, d.Resource, d.Property.Details().Path, d.Error)), nil
+	return json.Marshal(map[string]any{
+		"resource": d.Resource,
+		"property": d.Property.Details().Path,
+		"error":    d.Error,
+	})
 }
--- a/pkg/engine2/solution_context/decisions.go
+++ b/pkg/engine2/solution_context/decisions.go
@@ -1,6 +1,8 @@
 package solution_context
 
 import (
+	"fmt"
+
 	construct "github.com/klothoplatform/klotho/pkg/construct2"
 	knowledgebase "github.com/klothoplatform/klotho/pkg/knowledge_base2"
 )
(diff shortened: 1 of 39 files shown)


pkg/engine2/solution_context/decisions.go
```diff
@@ -50,14 +53,35 @@ type (
...
+
+func (d PropertyValidationDecision) MarshalJSON() ([]byte, error) {
+	if d.Value != nil {
+		stringVal := `{
+			"resource": "%s",
+			"property": "%s",
+			"value": "%s",
+			"error": "%s"
+		}`
+		return []byte(fmt.Sprintf(stringVal, d.Resource, d.Property.Details().Path, d.Value, d.Error)), nil
```
gordon-klotho, line 79: This doesn't do any escaping or anything. Might be safer to just `return json.Marshal(map[string]any{ ... })`

pkg/knowledge_base2/properties/any_property.go
```diff
@@ -107,7 +109,10 @@ func (a *AnyProperty) Type() string {
 	return "any"
 }
 
-func (a *AnyProperty) Validate(resource *construct.Resource, value any) error {
+func (a *AnyProperty) Validate(resource *construct.Resource, value any, ctx knowledgebase.DynamicContext) error {
+	if a.Required && value == nil {
```
gordon-klotho, line 113: Considering the infracopilot UI flow, it might not be possible to do required checks until we figure that out (unless that's something you're already doing). In other words, the resource has to be present to configure it in the UI, so if the validation fails on adding the resource, the user can't configure it to resolve that validation.

pkg/knowledge_base2/properties/bool_property_test.go
```diff
@@ -298,7 +302,19 @@ func Test_BoolPropertyValidate(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			assert := assert.New(t)
 			resource := &construct.Resource{}
-			err := tt.property.Validate(resource, tt.value)
+			graph := construct.NewGraph()
+			for _, r := range tt.testResources {
+				graph.AddVertex(r)
+			}
```
gordon-klotho, line 308: nit: Use `graphtest` and `[]any` to make it easier (no tests use this atm though, so it doesn't matter). Same in a couple of the other tests

pkg/knowledge_base2/properties/float_property.go
```diff
@@ -91,7 +91,13 @@ func (f *FloatProperty) Type() string {
...
+func (f *FloatProperty) Validate(resource *construct.Resource, value any, ctx knowledgebase.DynamicContext) error {
+	if value == nil {
+		if f.Required {
+			return fmt.Errorf(knowledgebase.ErrRequiredProperty, f.Path, resource.ID)
+		}
+		return nil
+	}
 	floatVal, ok := value.(float64)
 	if !ok {
 		return fmt.Errorf("invalid int value %v", value)
```
gordon-klotho, line 103:
```suggestion
		return fmt.Errorf("invalid float value %v", value)
```

gordon-klotho (commented):
Main blocking Q is the UI flow

pkg/knowledge_base2/properties/any_property.go
jhsinger-klotho: yeah validation can fail, but the engine wont fail unless its correlated with an incoming resource constraint. This prevents the user from dragging in a lambda and the engine failing, but still gives us the validation info for the images BaseImage so we can block clicking export IAC until all fields pass

pkg/engine2/testdata/ecs_rds.dataflow-viz.yaml
```diff
@@ -12,6 +12,6 @@ resources:
     parent: vpc/vpc-0
 
   ecs_service/ecs_service_0 -> rds_instance/rds-instance-2:
-    path: aws:ecs_task_definition:ecs_service_0,aws:iam_role:ecs_service_0-rds-instance-2
+    path: aws:ecs_task_definition:ecs_service_0,aws:iam_role:ecs_service_0-execution-role
```
gordon-klotho, line 15: 👍 fixed the bug w/ path expand vertex dependencies so this gets set by the op rule instead of expansion

gordon-klotho (approved):

PR #808, opened 2023-12-07
Adds GetValidEdgeTargets command to the engine

DavidSeptimus-Klotho:


pkg/engine2/edge_targets.go
```diff
@@ -0,0 +1,297 @@
...
+	targetClassifications := targetTemplate.PathSatisfaction.AsTarget
+	if len(targetClassifications) == 0 {
+		return false, nil
+	}
+	for _, satisfaction := range targetClassifications {
+		if satisfaction.Classification == "" {
+			continue
+		}
+		classifications.Add(satisfaction.Classification)
+	}
```
jhsinger-klotho, line 79: could use function

```
func (kb *KnowledgeBase) GetPathSatisfactionsFromEdge(source, target construct.ResourceId) ([]EdgePathSatisfaction, error) {
```

pkg/engine2/edge_targets.go
```diff
@@ -0,0 +1,297 @@
...
+			isTarget = false
+		}
+
+		if len(context.ResourceTypes.Sources) > 0 && !slices.ContainsFunc(context.ResourceTypes.Sources, qualifiedTypeMatcher(id)) {
+			isSource = false
+		}
+
+		if len(context.ResourceTypes.Targets) > 0 && !slices.ContainsFunc(context.ResourceTypes.Targets, qualifiedTypeMatcher(id)) {
+			isTarget = false
+		}
```
jhsinger-klotho, line 176: why have this check when you check for if the sources or targets contain the id before hand?

pkg/engine2/edge_targets.go
```diff
@@ -0,0 +1,297 @@
...
+			}
+
+			if source.Namespace == target.Name || target.Namespace == source.Name {
+				continue
+			}
+
+			path, err := graph.ShortestPath(topologyGraph, source, target)
+			if len(path) > 0 && err == nil {
+				continue
+			}
```
jhsinger-klotho, line 216: sucks to have to run this just to determine if theres a path that already exists because theres definitely more efficient ways.

jhsinger-klotho (commented):
nothing blocking, but maybe adding a few unit tests just to ensure we dont break happy path functionality (if the unit tests are too large then might be ok)

jhsinger-klotho (approved):

PR #812, opened 2023-12-07
In path expand, make sure IDs are settled before referencing them

gordon-klotho:
This fixes regression in ID propagation


jhsinger-klotho (approved):
id love to get some unit tests into op eval and path expansion, since theyre bare, but i see we have a functional test which will do for now

gordon-klotho:
This PR adds some non-determinism somehow. Looking into it

jhsinger-klotho (approved):

PR #813, opened 2023-12-07
make sub expansions only run for the classification that generated them

jhsinger-klotho:
makes sure tjat sub expansions only run for the classification that generated them. It also subs out the newly created names for the phantoms that were in their place in the temp graph so we can reuse the temp graph

changed the weighting of the service api to fix the lambda being placed into the vpc issue (we may not see vpc endpoint get created because of this, but thats at least hidden for now and we dont support many)

efs iac template fix

policy adds permissions classification to help weighting in path selection

### Standard checks

### DIFF
--- a/pkg/engine2/operational_eval/vertex_path_expand.go
+++ b/pkg/engine2/operational_eval/vertex_path_expand.go
@@ -158,12 +158,33 @@ func (v *pathExpandVertex) runExpansion(eval *Evaluator, expansion path_selectio
 	if err := eval.AddResources(resources...); err != nil {
 		return err
 	}
-	if err := eval.AddEdges(result.Edges...); err != nil {
-		return err
-	}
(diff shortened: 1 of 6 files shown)


gordon-klotho (approved):
Only thing is I'm not sure whether we want to do sub-expansions as separate vertices - only because other vertices which depended on the upper expansion would be unblocked, but I'm not confident those shouldn't also wait until sub-expansions are resolved as well.

jhsinger-klotho:
> Only thing is I'm not sure whether we want to do sub-expansions as separate vertices - only because other vertices which depended on the upper expansion would be unblocked, but I'm not confident those shouldn't also wait until sub-expansions are resolved as well.

yeah im not sure either, but we could definitely make that change. This is previously the way it was operating so im thinking lets keep it and evaluate if we want to do the expansion holistically later on

PR #815, opened 2023-12-08
updating reconciler to understand how to delete resources in groups based on iac dependencies

jhsinger-klotho:
makes the reconciliation check iac dependencies of namespaced resources to see if we need to clean up more than just the namespaced resources

### Standard checks

### DIFF
--- a/pkg/engine2/reconciler/reconciler.go
+++ b/pkg/engine2/reconciler/reconciler.go
@@ -145,7 +145,7 @@ func addAllDeploymentDependencies(
 		shouldDelete := false
 
 		// check if the dep exists as a property on the resource
-		rt.LoopProperties(res, func(p knowledgebase.Property) error {
+		err = rt.LoopProperties(res, func(p knowledgebase.Property) error {
 			propVal, err := res.GetProperty(p.Details().Path)
 			if err != nil {
 				return err
@@ -179,6 +179,9 @@ func addAllDeploymentDependencies(
 			}
 			return nil
 		})
+		if err != nil {
+			return nil, err
+		}
 
 		if shouldDelete {
 			queue = appendToQueue(deleteRequest{resource: dep, explicit: explicit}, queue)
--- a/pkg/engine2/reconciler/reconciler.go
+++ b/pkg/engine2/reconciler/reconciler.go
@@ -23,11 +23,10 @@ type (
 func RemoveResource(c solution_context.SolutionContext, resource construct.ResourceId, explicit bool) error {
 	zap.S().Debugf("reconciling removal of resource %s ", resource)
 
-	queue := make([]deleteRequest, 0)
-	queue = append(queue, deleteRequest{
+	queue := []deleteRequest{{
 		resource: resource,
 		explicit: explicit,
-	})
+	}}
 
 	var errs error
 
@@ -56,6 +55,8 @@ func RemoveResource(c solution_context.SolutionContext, resource construct.Resou
 		if !canDelete {
 			continue
 		}
+		// find all namespaced resources before removing edges and the initial resource, otherwise certain resources may
+		// be moved out of their original namespace
 		namespacedResources, err := findAllResourcesInNamespace(c, resource)
 		if err != nil {
 			errs = errors.Join(errs, err)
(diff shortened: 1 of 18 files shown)


pkg/engine2/reconciler/reconciler.go
```diff
@@ -3,65 +3,195 @@ package reconciler
...
+			continue
+		}
+		if !canDelete {
+			continue
+		}
+		namespacedResources, err := findAllResourcesInNamespace(c, resource)
+		if err != nil {
+			errs = errors.Join(errs, err)
+			continue
+		}
```
gordon-klotho, line 63: nit: move this closer to where it's used below

pkg/engine2/reconciler/reconciler.go
```diff
@@ -3,65 +3,195 @@ package reconciler
...
+			errs = errors.Join(errs, op.RemoveEdge(resource, res))
+		}
+
+		err = construct.RemoveResource(op, resource)
+		if err != nil {
+			errs = errors.Join(errs, err)
+			continue
+		}
+
+		for _, res := range namespacedResources.ToSlice() {
```
gordon-klotho, line 81:
```suggestion
		for res := range namespacedResources {
```
nit: don't need to copy the set to iterate it

pkg/engine2/reconciler/reconciler.go
```diff
@@ -3,65 +3,195 @@ package reconciler
...
+
+			// Since we are explicitly deleting the namespace resource, we will explicitly delete the resources namespaced to it
```
gordon-klotho, line 83: nit: update comment, we're forwarding the `explicit` here, the comment suggests that it's passing in `true`

pkg/engine2/reconciler/reconciler.go
```diff
@@ -3,65 +3,195 @@ package reconciler
...
+			case construct.ResourceId:
+				if val == resource {
+					found = true
+				}
+			case construct.PropertyRef:
+				if val.Resource == resource {
+					found = true
+				}
+			default:
+				if reflect.ValueOf(val).Kind() == reflect.Slice || reflect.ValueOf(val).Kind() == reflect.Array {
```
gordon-klotho, line 164: This doesn't look deeper into maps, or lists of lists but that can be a `TODO`. Only case it might be relevant is for policies that have a PropertyRef.

pkg/engine2/reconciler/reconciler.go
```diff
@@ -3,65 +3,195 @@ package reconciler
...
-	}
+func RemoveResource(c solution_context.SolutionContext, resource construct.ResourceId, explicit bool) error {
+	zap.S().Debugf("reconciling removal of resource %s ", resource)
 
-	op := c.OperationalView()
+	queue := make([]deleteRequest, 0)
+	queue = append(queue, deleteRequest{
+		resource: resource,
+		explicit: explicit,
+	})
```
gordon-klotho, line 30:
```suggestion
	queue := []deleteRequest{{
		resource: resource,
		explicit: explicit,
	}}
```
nit: more concise

gordon-klotho (commented):
Can you add an engine test for the delete? Maybe covering both use-cases (delete namespace resources & delete deploy dependencies)

pkg/engine2/reconciler/reconciler.go
jhsinger-klotho: we cant, ill add a comment, but it hast to be done before we remove the resource otherwise those resources wont be namespaced anymore

pkg/engine2/reconciler/reconciler.go
jhsinger-klotho: well that should be ok because doesnt loop properties get all indices of lists and keys of maps?

pkg/engine2/reconciler/reconciler.go
gordon-klotho: The edges shouldn't impact it, since it looks like it just walks the graph. But for removing the resource, makes sense. In that case, can the for loop be moved up?

pkg/engine2/reconciler/reconciler.go
gordon-klotho: You're right, but then shouldn't this not need to be added since it'd go into each index?

gordon-klotho (approved):

PR #818, opened 2023-12-11
Descriptions added to config

atorres-klo:
Descriptions for config

### Standard checks


pkg/templates/aws/resources/api_integration.yaml
```diff
@@ -24,16 +25,20 @@ properties:
...
     type: resource(aws:api_method)
     operational_rule:
       step:
         direction: upstream
         resources:
           - selector: aws:api_method
-            properties: 
+            properties:
               RestApi: '{{ fieldValue "RestApi" .Self }}'
-        unique: true # Each Integration must have its own method
```
jhsinger-klotho, line 36: this got removed unintenionally i assume, all the other removals look like just line changes

jhsinger-klotho (approved):

PR #821, opened 2023-12-12
fix bugs around property loading, naming, and temp graph replacement

jhsinger-klotho:
fix bugs around property loading, naming, and temp graph replacement

### Standard checks

### DIFF
--- a/pkg/engine2/operational_eval/evaluator.go
+++ b/pkg/engine2/operational_eval/evaluator.go
@@ -214,6 +214,7 @@ func (eval *Evaluator) addEdge(source, target Key) error {
 			ep.Attributes[attribAddedBy] = eval.currentKey.String()
 		}
 	})
+
 	if err != nil {
 		if errors.Is(err, graph.ErrEdgeCreatesCycle) {
 			path, _ := graph.ShortestPath(eval.graph, target, source)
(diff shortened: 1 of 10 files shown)


pkg/engine2/operational_eval/evaluator.go
```diff
@@ -452,12 +453,17 @@ func (eval *Evaluator) UpdateId(oldId, newId construct.ResourceId) error {
...
-						errs = errors.Join(errs, err)
-						continue
-					}
+			}
+			if vertex.TempGraph != nil {
+				_, err := vertex.TempGraph.Vertex(oldId)
+				switch {
+				case errors.Is(err, graph.ErrVertexNotFound):
+					// do nothing
+					break
```
gordon-klotho, line 462: `break`? that'd break out of the `for _, key := range topo` loop, is that what you intended?

pkg/engine2/testdata/k8s_api.dataflow-viz.yaml
```diff
@@ -4,11 +4,11 @@ resources:
     parent: eks_cluster/eks_cluster-0
 
 
-  load_balancer/rest-api-4-integ6897f0b9:
+  load_balancer/rest-api-4-integbcc77100:
```
gordon-klotho, line 7: Is this non-deterministic?

pkg/engine2/operational_eval/evaluator.go
jhsinger-klotho: nope definitely didnt mean for that to be there, removing

pkg/engine2/testdata/k8s_api.dataflow-viz.yaml
jhsinger-klotho: i dont think so, the ordering did change, which i assumed was from the way we load the graph, but i ran the same architecture 10 times and always saw the same result

gordon-klotho: Just wondering if the UUIDs weren't deterministic, do we set the seed somewhere?

gordon-klotho (approved):

pkg/engine2/testdata/k8s_api.dataflow-viz.yaml
jhsinger-klotho: dont think we do, we could probably look into that

PR #822, opened 2023-12-12
dont check validity if theres already an edge

jhsinger-klotho:
updates matcher to not have a default since it cant have a default for tcp

only run edge validity checks if theres no existing edge

### Standard checks

### DIFF
--- a/pkg/engine2/path_selection/path_expansion.go
+++ b/pkg/engine2/path_selection/path_expansion.go
@@ -437,11 +437,18 @@ func expandPath(
 			return
 		}
 
-		valid, err := checkUniquenessValidity(ctx, source.id, target.id)
-		if err != nil {
+		// if the edge doesnt exist in the actual graph we need to check uniqueness validity
+		_, err := ctx.RawView().Edge(source.id, target.id)
+		if errors.Is(err, graph.ErrEdgeNotFound) {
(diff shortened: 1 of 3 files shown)


pkg/engine2/path_selection/path_expansion.go
```diff
@@ -437,11 +437,18 @@ func expandPath(
...
 		}
 
-		valid, err := checkUniquenessValidity(ctx, source.id, target.id)
-		if err != nil {
+		// if the edge doesnt exist in the actual graph we need to check uniqueness validity
+		_, err := ctx.RawView().Edge(source.id, target.id)
+		if errors.Is(err, graph.ErrEdgeNotFound) {
+			valid, err := checkUniquenessValidity(ctx, source.id, target.id)
+			if err != nil {
+				errs = errors.Join(errs, err)
```
gordon-klotho, line 445: Does this need to return?

jhsinger-klotho: valid should be false, but yeah not a bad idea to have it return anyways

gordon-klotho: I guess not because if it returned an error, it should also return `!valid`

gordon-klotho (approved):

PR #823, opened 2023-12-13
eks fixes and sanitize pulumi name

jhsinger-klotho:
sanitize pulumi app name

a couple eks fixes

### Standard checks

### DIFF
--- a/pkg/infra/iac3/plugin.go
+++ b/pkg/infra/iac3/plugin.go
@@ -8,10 +8,10 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
+	"regexp"
 	"strings"
 	"text/template"
 
-	"github.com/klothoplatform/klotho/pkg/config"
 	construct "github.com/klothoplatform/klotho/pkg/construct2"
 	"github.com/klothoplatform/klotho/pkg/engine2/solution_context"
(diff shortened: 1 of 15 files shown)


pkg/infra/iac3/plugin.go
```diff
@@ -20,10 +20,16 @@ import (
 	"github.com/klothoplatform/klotho/pkg/templateutils"
 )
 
-type Plugin struct {
-	Config *config.Application
-	KB     *knowledgebase.KnowledgeBase
-}
+type (
+	PulumiConfig struct {
```
gordon-klotho, line 24: 👍

pkg/templates/kubernetes/resources/target_group_binding.yaml
```diff
@@ -14,8 +14,8 @@ sanitize_name:
 properties:
   Cluster:
     type: resource
-    namespace: true
```
gordon-klotho, line 17: Is the binding not namespaced to the cluster?

gordon-klotho (approved):

pkg/templates/kubernetes/resources/target_group_binding.yaml
jhsinger-klotho: it is, not sure why that got removed

PR #826, opened 2023-12-14
add iac dependencies based on properties

### LINKED ISSUE #772: Ensure deployment dependencies across resources when deploy time property ref is used
If we use a property ref within another property, we need to ensure that there is a dependency in the deployment graph to ensure correct iac

jhsinger-klotho:
closes #772 
ensures that there is always an iac dependency if a property contains another resource or property ref

modified the way api stages are created so that the edge of api -> integration creates it instead of the api itself

### Standard checks

### DIFF
--- a/pkg/engine2/solution_context/resource_configuration.go
+++ b/pkg/engine2/solution_context/resource_configuration.go
@@ -6,6 +6,7 @@ import (
 	"reflect"
 	"strings"
 
+	"github.com/dominikbraun/graph"
 	construct "github.com/klothoplatform/klotho/pkg/construct2"
 	"github.com/klothoplatform/klotho/pkg/engine2/constraints"
 	knowledgebase "github.com/klothoplatform/klotho/pkg/knowledge_base2"
@@ -50,10 +51,18 @@ func ConfigureResource(
 		if err != nil {
 			return fmt.Errorf("failed to set property %s on resource %s: %w", field, resource.ID, err)
 		}
+		err = AddDeploymentDependenciesFromVal(ctx, resource, val)
+		if err != nil {
+			return fmt.Errorf("failed to add deployment dependencies from property %s on resource %s: %w", field, resource.ID, err)
+		}
 	case "add":
 		err = property.AppendProperty(resource, val)
 		if err != nil {
 			return fmt.Errorf("failed to add property %s on resource %s: %w", field, resource.ID, err)
+		}
+		err = AddDeploymentDependenciesFromVal(ctx, resource, val)
+		if err != nil {
+			return fmt.Errorf("failed to add deployment dependencies from property %s on resource %s: %w", field, resource.ID, err)
 		}
 	case "remove":
 		err = property.RemoveProperty(resource, val)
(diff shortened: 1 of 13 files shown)


pkg/engine2/solution_context/resource_configuration.go
```diff
@@ -71,6 +80,25 @@ func ConfigureResource(
...
 
+func AddDeploymentDependenciesFromVal(
+	ctx SolutionContext,
+	resource *construct.Resource,
+	val any,
+) error {
+	var errs error
+	ids := getResourcesFromValue(val)
+	for _, id := range ids {
+		if resource.ID.Matches(id) {
```
gordon-klotho, line 91: Can this happen? How would this look like in IaC?

pkg/engine2/solution_context/resource_configuration.go
```diff
@@ -84,6 +112,33 @@ func ConstraintOperatorToAction(op constraints.ConstraintOperator) (string, erro
...
+	default:
+		rval := reflect.ValueOf(val)
+		switch rval.Kind() {
+		case reflect.Slice, reflect.Array:
+			for i := 0; i < reflect.ValueOf(val).Len(); i++ {
+				idVal := rval.Index(i).Interface()
+				ids = append(ids, getResourcesFromValue(idVal)...)
+			}
+		case reflect.Map:
+			for _, key := range reflect.ValueOf(val).MapKeys() {
```
gordon-klotho, line 133: What about keys, could there ever be something like a `map(resource, X)`?

pkg/templates/aws/edges/rest_api-api_integration.yaml
```diff
@@ -2,3 +2,9 @@ source: aws:rest_api
 target: aws:api_integration
 direct_edge_only: true
 deployment_order_reversed: true
+operational_rules:
+  - steps:
```
gordon-klotho, line 6: Probably worth a comment that this is kind of a workaround, maybe link to a ticket.

gordon-klotho (approved):

pkg/engine2/solution_context/resource_configuration.go
jhsinger-klotho: yeah for the uri on the lb we use it as a ref to itself so that it can get the integrations route

pkg/engine2/solution_context/resource_configuration.go
jhsinger-klotho: all of our properties are string keys right?

pkg/templates/aws/edges/rest_api-api_integration.yaml
jhsinger-klotho: 👍

pkg/engine2/solution_context/resource_configuration.go
gordon-klotho: 👍 right, for the deployment variables, those are substituted

pkg/engine2/solution_context/resource_configuration.go
gordon-klotho: If that's the case, then the type should be `map(valueType)` and not let it be configurable to make sure.

PR #827, opened 2023-12-14
Sanitize environment variable names when added via consumer logic

gordon-klotho:
Note: we should probably also sanitize for other automatically generated things (ie, that don't come from user constraints), but that's left for later.

### DIFF
--- a/pkg/templates/aws/resources/ecs_task_definition.yaml
+++ b/pkg/templates/aws/resources/ecs_task_definition.yaml
@@ -28,6 +28,13 @@ properties:
   EnvironmentVariables:
     type: map(string,string)
     description: A map of the environment variables to pass to the container
+    key_property:
+      sanitize: |
+        {{ .
+          | replace `[^[:alnum:]_]+` "_"
+          | replace `^[^a-zA-Z]+` ""
+          | upper
+        }}
   Cpu:
     type: string
     default_value: '256'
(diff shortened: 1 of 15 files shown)


pkg/templates/aws/resources/ecs_task_definition.yaml
```diff
@@ -28,6 +28,13 @@ properties:
   EnvironmentVariables:
     type: map(string,string)
     description: A map of the environment variables to pass to the container
+    key_property:
+      sanitize: |
+        {{ .
+          | replace `[^[:alnum:]_]+` "_"
+          | replace `^[^a-zA-Z]+` ""
+          | upper
+        }}
```
jhsinger-klotho, line 37: i dont know that this actually follows the same stuff as lambda since ecs tests were passing, but not blocking

pkg/templates/kubernetes/models/env_var.yaml
```diff
@@ -2,5 +2,11 @@ name: kubernetes:EnvVar
 properties:
   name:
     type: string
+    sanitize: |
+      {{ .
+        | replace `[^[:alnum:]_]+` "_"
+        | replace `^[^a-zA-Z]+` ""
+        | upper
+      }}
```
jhsinger-klotho, line 10: same, this isnt necessary in kubernetes i dont think

jhsinger-klotho (approved):

PR #831, opened 2023-12-18
Add dependencies to all parts of the property ref path in path expand

gordon-klotho:
Since `lambda#Subnets` had 1 value from another path expansion, the other expansion wasn't properly depending on the final resolution of the subnets, so when `Expand(aws:lambda_function:lambda_test_app -> aws:efs_mount_target:test-efs-fs:lambda_test_app-test-efs-fs, <network>, source#Subnet#AvailabilityZone)` ran before, the lambda only had 1 subnet and wasn't properly setting up the path to mount point for the other subnet.

Pending e2e test to confirm, but resources.yaml now has 2 mount_points which it did not previously.

### DIFF
--- a/pkg/engine2/operational_eval/vertex_path_expand.go
+++ b/pkg/engine2/operational_eval/vertex_path_expand.go
@@ -13,6 +13,7 @@ import (
 	"github.com/klothoplatform/klotho/pkg/engine2/solution_context"
 	knowledgebase "github.com/klothoplatform/klotho/pkg/knowledge_base2"
 	"github.com/klothoplatform/klotho/pkg/set"
+	"go.uber.org/zap"
 )
 
 type (
@@ -32,6 +33,13 @@ func (v *pathExpandVertex) Evaluate(eval *Evaluator) error {
 	expansions, err := v.getExpansionsToRun(eval)
 	if err != nil {
 		errs = errors.Join(errs, fmt.Errorf("could not get expansions to run: %w", err))
+	}
+	log := eval.Log()
+	if len(expansions) > 1 && log.Desugar().Core().Enabled(zap.DebugLevel) {
+		log.Debugf("Expansion %s subexpansions:", v.Edge)
+		for _, expansion := range expansions {
+			log.Debugf(" %s -> %s", expansion.Dep.Source.ID, expansion.Dep.Target.ID)
+		}
 	}
 
 	for _, expansion := range expansions {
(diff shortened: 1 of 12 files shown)


pkg/engine2/operational_eval/vertex_path_expand.go
```diff
@@ -471,15 +472,8 @@ func (v *pathExpandVertex) Dependencies(eval *Evaluator) (graphChanges, error) {
...
-	var errs error
-	if propertyReferenceInfluencesEdge(v.Satisfication.Source) {
-		keys := getDepsForPropertyRef(eval.Solution, v.Edge.Source, v.Satisfication.Source.PropertyReference)
-		changes.addEdges(srcKey, keys)
-	}
-	if propertyReferenceInfluencesEdge(v.Satisfication.Target) {
-		keys := getDepsForPropertyRef(eval.Solution, v.Edge.Target, v.Satisfication.Target.PropertyReference)
-		changes.addEdges(srcKey, keys)
-	}
+	changes.addEdges(srcKey, getDepsForPropertyRef(eval.Solution, v.Edge.Source, v.Satisfication.Source.PropertyReference))
```
gordon-klotho, line 475: This was 1 part of the fix - the dependencies were getting skipped due to it having Validity.

pkg/engine2/operational_eval/vertex_path_expand.go
```diff
@@ -584,16 +579,20 @@ func getDepsForPropertyRef(
...
+		return nil
+	}
 	keys := make(set.Set[Key])
 	cfgCtx := solution_context.DynamicCtx(sol)
 	currResources := []construct.ResourceId{res}
 	parts := strings.Split(propertyRef, "#")
 	for _, part := range parts {
 		var nextResources []construct.ResourceId
 		for _, currResource := range currResources {
+			keys.Add(Key{Ref: construct.PropertyRef{Resource: currResource, Property: part}})
```
gordon-klotho, line 592: This is the other part of the fix, all elements of the property path need to be depended on.

pkg/engine2/path_selection/path_expansion.go
```diff
@@ -128,23 +135,45 @@ func renameAndReplaceInTempGraph(
...
+				continue
+			}
+
+		case err != nil:
+			errs = errors.Join(errs, err)
+			continue
+		}
+		result[i] = r
+		if i > 0 {
+			err = g.AddEdge(result[i-1].ID, r.ID)
```
gordon-klotho, line 165: Not sure why my change broke the k8s functional test, but this change (`id` -> `r.ID`) fixed it.

gordon-klotho:
e2e test passed
```
INFO:SuiteRunner:Test result for aws: name: aws
        result: Result.SUCCESS
```

jhsinger-klotho (approved):

PR #835, opened 2023-12-19
rename resources

jhsinger-klotho:
fixes bug in renaming resources

### Standard checks

### DIFF
--- a/pkg/engine2/testdata/rename.expect.yaml
+++ b/pkg/engine2/testdata/rename.expect.yaml
@@ -1,0 +1,65 @@
+resources:
+    aws:lambda_function:new_lambda_name:
+        EnvironmentVariables:
+            DYNAMODB_TABLE_1_TABLE_NAME: aws:dynamodb_table:dynamodb_table_1#Name
+        ExecutionRole: aws:iam_role:lambda_function_0-ExecutionRole
+        Image: aws:ecr_image:lambda_function_0-image
+        LogGroup: aws:log_group:new_lambda_name-log-group
+        MemorySize: 512
+        Timeout: 180
+    aws:SERVICE_API:lambda_function_0-lambda_function_0-log-group:
+    aws:ecr_image:lambda_function_0-image:
+        Context: .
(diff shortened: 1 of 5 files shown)


pkg/engine2/testdata/rename.expect.yaml
```diff
@@ -0,0 +1,65 @@
...
+edges:
+    aws:lambda_function:new_lambda_name -> aws:SERVICE_API:lambda_function_0-lambda_function_0-log-group:
+    aws:lambda_function:new_lambda_name -> aws:ecr_image:lambda_function_0-image:
+    aws:lambda_function:new_lambda_name -> aws:iam_role:lambda_function_0-ExecutionRole:
+    aws:SERVICE_API:lambda_function_0-lambda_function_0-log-group -> aws:dynamodb_table:dynamodb_table_1:
+    aws:SERVICE_API:lambda_function_0-lambda_function_0-log-group -> aws:log_group:lambda_function_0-log-group:
+    aws:SERVICE_API:lambda_function_0-lambda_function_0-log-group -> aws:log_group:new_lambda_name-log-group:
+    aws:ecr_image:lambda_function_0-image -> aws:ecr_repo:ecr_repo-0:
+    aws:iam_role:lambda_function_0-ExecutionRole -> aws:dynamodb_table:dynamodb_table_1:
+    aws:iam_role:lambda_function_0-ExecutionRole -> aws:log_group:lambda_function_0-log-group:
```
gordon-klotho, line 64: TODO: reconciler should clean this up

gordon-klotho (approved):

pkg/engine2/testdata/rename.expect.yaml
jhsinger-klotho: should it or do we just want it to reconnect them? These things are still used when we just renamed and if we clean up we may lose paths to other resources

gordon-klotho: I think in this case it should (or reuse, just 1 log group is the point). Now, the generalizable logic to accomplish this, I'm not sure which is the TODO part.

PR #837, opened 2023-12-20
Expand visualizer children to include more resources

gordon-klotho:
In draft because this will require backend changes to support the new structure (yaml lists instead of comma-separated strings)


pkg/visualizer2/connection_graph.go
```diff
@@ -0,0 +1,200 @@
...
+
+// ConnectionGraph facilitates whether resources should be connected for visualization. Resources are only considered
+// connected if there are paths that satisfy both "network" and "permissions" classifications.
+type ConnectionGraph struct {
+	Network, Permissions ClassGraph
+	Undirected           construct.Graph
+	KB                   knowledgebase.TemplateKB
+}
+
+func NewConnectionGraph(g construct.Graph, kb knowledgebase.TemplateKB) (ConnectionGraph, error) {
```
jhsinger-klotho, line 21: im confused on what we need this for and where this is managed/ why are network and permissions always required?

jhsinger-klotho: right but not all connections must satisfy both network and permissions, plus we had the reusable logic for does it have a path so are we not using that anymore? i guess im confused when do we add things to this graph

gordon-klotho: Hm, yeah... This was much different when I started (was the AllPathsBetween version) and when I rebased I didn't notice it had changed. The `path_selection` change was done when I had already moved onto the UI stuff. I'll fix the rebase and probably remove the ConnectionGraph stuff.

pkg/visualizer2/vis_graph.go
```diff
@@ -0,0 +1,53 @@
...
+	VisResource struct {
+		ID construct.ResourceId
+
+		Tag      string
+		Parent   construct.ResourceId
+		Children set.Set[construct.ResourceId]
+	}
+
+	VisGraph graph.Graph[construct.ResourceId, *VisResource]
+)
```
jhsinger-klotho, line 22: what are these used for?

jhsinger-klotho: nevermind i think i see were using a graph instead of the maps that were being used in the output function

gordon-klotho: `VisGraph` is basically what `Topology` used to be, just moved into visualizer package and using the `graph` library instead of our own Nodes & Edges struct.

jhsinger-klotho (approved):

PR #838, opened 2023-12-20
fix eks bugs

jhsinger-klotho:
fixes kubeconfig circular iac dependency and emitter bug to where we try to sanitize to the wrong type since its done before convert

### Standard checks

### DIFF
--- a/pkg/knowledge_base2/emitter.go
+++ b/pkg/knowledge_base2/emitter.go
@@ -90,24 +90,24 @@ func ConsumeFromResource(consumer, emitter *construct.Resource, ctx DynamicConte
 					addErr(consume, emit, err)
 					continue
 				}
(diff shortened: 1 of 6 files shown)


pkg/knowledge_base2/emitter.go
```diff
@@ -90,32 +90,31 @@ func ConsumeFromResource(consumer, emitter *construct.Resource, ctx DynamicConte
 					addErr(consume, emit, err)
 					continue
 				}
-				val, err = sanitizeForConsumption(ctx, resource, consumeTmpl.GetProperty(consume.PropertyPath), val)
+				pval, err := resource.GetProperty(consume.PropertyPath)
```
gordon-klotho, line 93: nit: why did this move up? `pval` isn't used until after converting & sanitizing so it can be done later

gordon-klotho (approved):

PR #840, opened 2024-01-02
allow ability to add imported resources via constraints

jhsinger-klotho:
adds the import constraint which currently just adds the imported resource and flags it as being imported.

This pr does not affect any of the graph solving or protect the imported resource from mutations

### Standard checks

### DIFF
--- a/pkg/construct2/resource.go
+++ b/pkg/construct2/resource.go
@@ -3,6 +3,7 @@ package construct2
 type Resource struct {
 	ID         ResourceId
 	Properties Properties
+	Imported   bool
 }
 
 // Id is a temporary bridge to the old Resource interface.
--- a/pkg/engine2/path_selection/candidate_validity.go
+++ b/pkg/engine2/path_selection/candidate_validity.go
@@ -246,7 +246,7 @@ func (d downstreamChecker) makeValid(resource, operationResource *construct.Reso
 		var errs error
 		rt, err := d.ctx.KnowledgeBase().GetResourceTemplate(r)
 		if err != nil || rt == nil {
-			return false, fmt.Errorf("error getting resource template for resource %s: %w", resource, err)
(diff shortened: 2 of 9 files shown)


pkg/engine2/path_selection/candidate_validity.go
```diff
@@ -246,7 +246,7 @@ func (d downstreamChecker) makeValid(resource, operationResource *construct.Reso
 		var errs error
 		rt, err := d.ctx.KnowledgeBase().GetResourceTemplate(r)
 		if err != nil || rt == nil {
-			return false, fmt.Errorf("error getting resource template for resource %s: %w", resource, err)
+			return false, fmt.Errorf("error getting resource template for resource %s: %w", resource.ID, err)
```
gordon-klotho, line 249: side comment - we should add a `String()` to resource that just prints its ID. This is a common enough occurrence and if printing properties is desired that can be implemented via `%+v` or something.

pkg/construct2/resource.go
```diff
@@ -3,6 +3,7 @@ package construct2
 type Resource struct {
 	ID         ResourceId
 	Properties Properties
+	Imported   bool
```
gordon-klotho, line 6: Not blocking, but I don't think the construct package cares on the difference between an imported or otherwise immutable resource. Just naming though, can change later if we so desire.

gordon-klotho (approved):

pkg/engine2/path_selection/candidate_validity.go
jhsinger-klotho: yeah not a bad idea

pkg/construct2/resource.go
jhsinger-klotho: i didnt want to group them together without thinking through it a little bit more but its probably true that they will be handled the same way so i figure we can always change the naming

PR #841, opened 2024-01-03
Imports engine solve

jhsinger-klotho:
Ability to solve graphs with imported resources

- checks to ensure we will fail in configuring a resource if its not an explicit command from the user
- functional tests to test importing cases with vpcs

### Standard checks


pkg/construct2/graph_io.go
```diff
@@ -46,7 +46,12 @@ func (g YamlGraph) MarshalYAML() (interface{}, error) {
 			errs = errors.Join(errs, err)
 			continue
 		}
-
+		if r.Imported {
+			if r.Properties == nil {
+				r.Properties = make(Properties)
+			}
```
gordon-klotho, line 52: Properties should never be `nil`, while it's not documented, it's part of the contract.

pkg/construct2/graph_io.go
```diff
@@ -119,10 +124,20 @@ func (g *YamlGraph) UnmarshalYAML(n *yaml.Node) error {
 	}
 
 	var errs error
-	for rid, props := range y.Resouces {
+	for rid, props := range y.Resources {
+		var imported bool
+		if imp, ok := props["imported"]; ok {
+			val, ok := imp.(bool)
+			if !ok {
+				errs = errors.Join(errs, fmt.Errorf("unable to parse imported value as boolean for resource %s", rid))
```
gordon-klotho, line 132:
```suggestion
				errs = errors.Join(errs, fmt.Errorf("unable to parse imported value as boolean for resource %s", rid))
				// Don't continue here so that the vertex is still added, otherwise it could erroneously cause failures in the edge copying
```

pkg/construct2/graph_io.go
```diff
@@ -119,10 +124,20 @@ func (g *YamlGraph) UnmarshalYAML(n *yaml.Node) error {
...
+			}
+			imported = val
+			delete(props, "imported")
+		}
 		err := g.Graph.AddVertex(&Resource{
 			ID:         rid,
 			Properties: props,
+			Imported:   bool(imported),
```
gordon-klotho, line 140: nit: why the cast, isn't imported already a bool?

pkg/engine2/operational_eval/vertex_property.go
```diff
@@ -119,12 +119,20 @@ func (v *propertyVertex) Evaluate(eval *Evaluator) error {
 		return err
 	}
 
+	// we know we cannot change properties of imported resources only users through constraints
+	// we still want to be able to update ids in case they are setting the property of a namespaced resource
+	// so we just conditionally run the edge operational rules
+	//
+	// we still need to run the resource operational rules though,
+	// to make sure dependencies exist where properties have operational rules set
```
gordon-klotho, line 127: I still think this isn't the right approach to importing, but I won't block the PR on it.

pkg/engine2/operational_rule/operational_step.go
```diff
@@ -306,6 +316,11 @@ func (ctx OperationalRuleContext) SetField(resource, fieldResource *construct.Re
 		return fmt.Errorf("error appending field %s#%s with %s: %w", resource.ID, path, fieldResource.ID, err)
 	}
 	zap.S().Infof("appended field %s#%s with %s", resource.ID, path, fieldResource.ID)
+	ctx.namespace(resource, fieldResource, oldId)
+	return nil
+}
+
+func (ctx OperationalRuleContext) namespace(resource, fieldResource *construct.Resource, oldId construct.ResourceId) {
```
gordon-klotho, line 323:
```suggestion
func (ctx *OperationalRuleContext) namespace(resource, fieldResource *construct.Resource, oldId construct.ResourceId) {
```
nit: this works as-is because `.Data` is a pointer so it gets copied over, but for methods that mutate I generally prefer them clearly marked such by using a pointer receiver.

pkg/knowledge_base2/properties/float_property_test.go
```diff
@@ -238,7 +237,7 @@ func Test_FloatProperty_Validate(t *testing.T) {
 			for _, call := range test.mockKBCalls {
 				mockKB.On(call.Method, call.Arguments...).Return(call.ReturnArguments...)
 			}
-			ctx := knowledgebase.DynamicValueContext{
+			ctx := knowledgebase2.DynamicValueContext{
```
gordon-klotho, line 240: nit: For consistency, this should be imported as `knowledgebase` - that's what we do everywhere else in the code.

pkg/knowledge_base2/properties/resource_property.go
```diff
@@ -108,6 +108,11 @@ func (r *ResourceProperty) ZeroValue() any {
 }
 
 func (r *ResourceProperty) Contains(value any, contains any) bool {
+	if val, ok := value.(construct.ResourceId); ok {
+		if cont, ok := contains.(construct.ResourceId); ok {
+			return val == cont
```
gordon-klotho, line 113: Should `Contains` be a match if `value` is a selector?
Eg. `Subnet.VPC.Contains("aws:vpc") == true`

pkg/templates/aws/resources/subnet.yaml
```diff
@@ -54,6 +54,11 @@ properties:
   MapPublicIpOnLaunch:
     type: bool
     default_value: false
+  Id:
+    type: string
+    configuration_disabled: true
+    deploy_time: true
+    required: true
```
gordon-klotho, line 61: What does `deploy_time` + `required` mean? Seems paradoxical except for imports. I'm leaning to a different flag than `required` to indicate which property is the unique identifier - this will be needed to distinguish which property to use when doing the IaC when there are multiple required properties.

gordon-klotho (approved):
Some things I'd like cleaned up, but nothing blocking

pkg/construct2/graph_io.go
jhsinger-klotho: i ran into a nil pointer exception during my testing which is why i added this. I didnt look into why the properties were nil too deeply though

pkg/construct2/graph_io.go
jhsinger-klotho: must have been an autocomplete that i missed, will remove

pkg/engine2/operational_eval/vertex_property.go
jhsinger-klotho: vpc is a simpler case so ill try to add another imported resource and see if this falls apart, but i do think it makes sense to have our logic account for this rather than a special case in the beginning because we do need to use rules to determine edges, etc

pkg/knowledge_base2/properties/resource_property.go
jhsinger-klotho: yeah it maybe should, the weird part is in the way its used today were more so checking for equals, but in the list,set,map case only care that a single index is equal which is why contains is called. Im wondering if we need a different method to cover this then

pkg/templates/aws/resources/subnet.yaml
jhsinger-klotho: right now it means required for importing since we know when we create a resource, we wont have the deploy_time properties.

Not entirely sure what you mean about distinguishing for iac

pkg/construct2/graph_io.go
gordon-klotho: Sounds like a bug somewhere. I'd rather fix the cause of breaking the contract than start sprinkling guards against breaking the contract. (we should also document that `Resource.Properties` is never nil on the godoc)

PR #842, opened 2024-01-03
support imports in iac

jhsinger-klotho:
changes to the iac plugin to render imported resources using a new templated function

### Standard checks


pkg/infra/iac3/resource_template.go
```diff
@@ -242,6 +254,33 @@ func exportsNodeToTemplate(tc *TemplatesCompiler, tmpl *ResourceTemplate, node *
...
 
+func importFuncNodeToTemplate(node *sitter.Node, name string) (*template.Template, string, error) {
+	importFunc := doQuery(node, findImportFunc)
+	imp, found := importFunc()
+	if !found {
+		return nil, "", nil
+	}
+	outputType := imp["return_type"].Content()
+	var expressionBody string
+	if outputType == "void" {
```
gordon-klotho, line 265: Does import returning `void` make any sense? What's the point of importing if it's not used. If it's used, it must not be `void`.

pkg/infra/iac3/resource_template_test.go
```diff
@@ -2,6 +2,7 @@ package iac3
 
 import (
 	"context"
+	_ "embed"
```
gordon-klotho, line 5: Unused import?

pkg/infra/iac3/resource.go
```diff
@@ -41,9 +41,16 @@ func (tc *TemplatesCompiler) RenderResource(out io.Writer, rid construct.Resourc
 			return err
 		}
 	}
-	err = resTmpl.Template.Execute(out, inputs)
-	if err != nil {
-		return fmt.Errorf("could not render resource %s: %w", rid, err)
+	if r.Imported {
+		err = resTmpl.ImportResource.Execute(out, inputs)
```
gordon-klotho, line 45: So this combined with not having a distinguishing "identifier" flag on a property (and just relying on `required`) means that we can't have any default import templates. Fine for now if we only want a handful of resources to be importable, but I think we should consider how to scale this up to be able to import any resources we support.

gordon-klotho (approved):

pkg/infra/iac3/resource.go
jhsinger-klotho: i guess it depends on if one deploy time property is always all it takes to import things. If so we can use the one required deploy time property for a base template, which shouldnt be too hard, but also writing the importFunc in the template should be real quick and simple as well

gordon-klotho: > the one required deploy time property

Is this guaranteed to be singular?

jhsinger-klotho: it doesnt have to be, i dont know what resource would require multiple, but i could definitely see that happening

PR #848, opened 2024-01-08
unit tests

jhsinger-klotho:
using mocks to add unit tests

### Standard checks


Makefile
```diff
@@ -30,3 +30,6 @@ engine_test:
 .PHONY: regen_tests
 regen_tests:
 	find pkg/engine2/testdata -type f -name '*.input.yaml' -exec ./create_test.sh {} \;
+
+mocks:
+	mockgen -source=./pkg/engine2/operational_rule/operational_rule.go --destination=./pkg/engine2/operational_eval/operational_rule_mock_test.go --package=operational_eval
```
gordon-klotho, line 35: This should probably be in a `go generate`

pkg/engine2/operational_eval/vertex_property_test.go
```diff
@@ -0,0 +1,106 @@
...
+				res: &construct.Resource{ID: construct.ResourceId{Name: "test"}},
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+			opctx := NewMockOpRuleHandler(ctrl)
+			opctx.EXPECT().HandlePropertyRule(*rule).Return(nil).Times(1)
+			if err := tt.args.v.evaluateResourceOperational(tt.args.res, opctx); (err != nil) != tt.wantErr {
```
gordon-klotho, line 49: nit: use `assert` for standardization

gordon-klotho (approved):
Overall, 👍 nice

PR #857, opened 2024-01-10
Adds ALB Support to the Engine

DavidSeptimus-Klotho:
This change's primary goal was adding support for specifying ALB Listener rules.

It includes a few other changes:
- Improved support for ACM Certificates
- Improved valid edge detection
- Added support for more template property validations

### DIFF
--- a/pkg/templates/aws/edges/load_balancer-load_balancer_listener.yaml
+++ b/pkg/templates/aws/edges/load_balancer-load_balancer_listener.yaml
@@ -2,12 +2,6 @@ source: aws:load_balancer
 target: aws:load_balancer_listener
 deployment_order_reversed: true
 operational_rules:
-  - if: '{{eq (fieldValue "Type" .Source) "application"}}'
-    configuration_rules:
-      - resource: '{{ .Target }}'
-        configuration:
-          field: Protocol
-          value: HTTP
   - if: '{{ eq (fieldValue "Type" .Source) "network" }}'
     configuration_rules:
       - resource: '{{ .Target }}'
--- a/pkg/templates/aws/resources/load_balancer.yaml
+++ b/pkg/templates/aws/resources/load_balancer.yaml
@@ -19,6 +19,9 @@ properties:
   IpAddressType:
     type: string
     description: Specifies the type of IP addresses used by the load balancer's subnets
+    allowed_values:
+      - ipv4
+      - dualstack
   LoadBalancerAttributes:
     type: map(string,string)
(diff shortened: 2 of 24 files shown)


pkg/templates/aws/edges/load_balancer-load_balancer_listener.yaml
```diff
@@ -2,12 +2,6 @@ source: aws:load_balancer
 target: aws:load_balancer_listener
 deployment_order_reversed: true
 operational_rules:
-  - if: '{{eq (fieldValue "Type" .Source) "application"}}'
-    configuration_rules:
-      - resource: '{{ .Target }}'
-        configuration:
-          field: Protocol
-          value: HTTP
```
jhsinger-klotho, line 10: do we have a rule somewhere to have the listener track the protocol of the LB?

pkg/templates/aws/resources/load_balancer.yaml
```diff
@@ -91,4 +99,4 @@ classification:
 delete_context:
   requires_no_downstream: true
 views:
-  dataflow: big
```
jhsinger-klotho, line 94: i know we were discussing this, are you just going with the approach on the ui to flip this for nlb?

pkg/templates/aws/resources/load_balancer_listener.yaml
```diff
@@ -2,23 +2,43 @@ qualified_type_name: aws:load_balancer_listener
...
   Port:
     type: int
     default_value: 80
+    min_value: 1
+    max_value: 65535
+    required: true
   Protocol:
     type: string
+    allowed_values:
+        - GENEVE
```
jhsinger-klotho, line 22: still curious on what this is

jhsinger-klotho (approved):

pkg/templates/aws/resources/load_balancer_listener.yaml
DavidSeptimus-Klotho: Looks like it's for gateway load balancers: https://docs.aws.amazon.com/elasticloadbalancing/latest/gateway/gateway-load-balancers.html.

pkg/templates/aws/edges/load_balancer-load_balancer_listener.yaml
DavidSeptimus-Klotho: A downstream target group will match the protocol of the listener, which I believe is existing behavior. That said, I'm not sure I agree with it since it's kind of up to the user to decide where they want to terminate SSL.

PR #858, opened 2024-01-10
dont have default since it might not exist

### LINKED ISSUE #851: Can't delete EKS cluster

jhsinger-klotho:
closes #851 

we cant rely on a default here, for eks case the cluster security group doesnt exist in our graph, but is what we create a rule against so it just needs to be set. Making the field required should ensure it always has a value

### Standard checks

### DIFF
--- a/pkg/templates/aws/resources/security_group_rule.yaml
+++ b/pkg/templates/aws/resources/security_group_rule.yaml
@@ -22,7 +22,7 @@ properties:
       security group rule
   SecurityGroupId:
     type: string # Not resource type because we need deploy time properties to be able to exist here
-    default_value: '{{ upstream "aws:security_group" .Self }}#Id'
+    required: true
     description: The ID of the security group to which this rule should be attached
   Type:
     type: string


pkg/templates/aws/resources/security_group_rule.yaml
```diff
@@ -22,7 +22,7 @@ properties:
       security group rule
   SecurityGroupId:
     type: string # Not resource type because we need deploy time properties to be able to exist here
-    default_value: '{{ upstream "aws:security_group" .Self }}#Id'
+    required: true
```
gordon-klotho, line 25: Should this also be namespaced to the SG?

gordon-klotho (approved):

pkg/templates/aws/resources/security_group_rule.yaml
jhsinger-klotho: this has to be a string value so i dont think namespace fits too too well since we expect it to be a resource. And in some cases the SG wont exist in our graph either

PR #862, opened 2024-01-11
fix templates for deployment order

jhsinger-klotho:
this never mattered since we bundle them into helm charts, but now that we add the dependencies based on fields on the property it was causing a cycle

also fixes lambda -> elasticache by exposing the cache node address for redis nodes

### Standard checks

### DIFF
--- a/pkg/templates/aws/resources/elasticache_cluster.yaml
+++ b/pkg/templates/aws/resources/elasticache_cluster.yaml
@@ -47,11 +47,16 @@ properties:
     configuration_disabled: true
     deploy_time: true
     description: The port number on which each of the cache nodes accepts connections.
+  CacheNodeAddress:
+    type: string
+    configuration_disabled: true
+    deploy_time: true
+    description: The endpoint address of a single node in the ElastiCache cluster.
   ClusterAddress:
     type: string
     configuration_disabled: true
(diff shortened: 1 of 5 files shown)


pkg/templates/aws/resources/elasticache_cluster.yaml
```diff
@@ -47,11 +47,16 @@ properties:
     configuration_disabled: true
     deploy_time: true
     description: The port number on which each of the cache nodes accepts connections.
+  CacheNodeAddress:
+    type: string
+    configuration_disabled: true
+    deploy_time: true
+    description: The endpoint address of a single node in the ElastiCache cluster.
```
gordon-klotho, line 54: Why a single node and not all the nodes' addresses?

gordon-klotho (approved):
The deployment order stuff is fine

pkg/templates/aws/resources/elasticache_cluster.yaml
jhsinger-klotho: yeah we definitely could return an array of all the addresses, i did single node just because its plug in the ref and your code will work, you dont have to split an array or anything and i figured the no touch approach is good for both our e2e suite and tutorials, etc

PR #872, opened 2024-01-12
Adds conditional rendering for LB listener rule conditions

DavidSeptimus-Klotho:
If an LB listener rule's conditions are nil, render an empty array in IaC output.


pkg/infra/iac3/templates/aws/load_balancer_listener_rule/factory.ts
```diff
@@ -16,7 +16,11 @@ function create(args: Args): aws.lb.ListenerRule {
     return new aws.lb.ListenerRule(args.Name, {
         listenerArn: args.Listener.arn,
         priority: args.Priority,
+        //TMPL {{- if .Conditions }}
         conditions: args.Conditions,
+        //TMPL {{- else }}
+        conditions: [],
```
gordon-klotho, line 22: nit: does this cause a TS compilation error viewing the file? If it does, just prefixing one of them with `//TMPL` as well will fix

gordon-klotho (approved):

PR #884, opened 2024-01-17
add resource rule vertices

jhsinger-klotho:
adds the ability to add additional rules into resource templates.

creates resource rule vertices that can operate as expected in op eval.

slight refactoring to the dependencies call so it can be unit tested

### Standard checks


pkg/engine2/operational_eval/vertex_edge.go
```diff
@@ -116,61 +104,20 @@ func (ev *edgeVertex) Evaluate(eval *Evaluator) error {
...
-				v.EdgeRules[ev.Edge] = append(v.EdgeRules[ev.Edge], knowledgebase.OperationalRule{
-					If:                 rule.If,
-					ConfigurationRules: []knowledgebase.ConfigurationRule{config},
-				})
-			} else {
-				errs = errors.Join(errs, fmt.Errorf("existing vertex for %s is not a property vertex", ref))
-				continue
-			}
+		// the configurations that are returned can be executed out of band from the property vertex
+		// since the property vertex has already been evaluated. This is a hack to get around improper dep ordering
```
gordon-klotho, line 108: 👍 good comment

pkg/engine2/operational_eval/vertex_property.go
```diff
@@ -18,21 +18,17 @@ type (
 	propertyVertex struct {
 		Ref construct.PropertyRef
 
-		Template  knowledgebase.Property
-		EdgeRules map[construct.SimpleEdge][]knowledgebase.OperationalRule
+		Template      knowledgebase.Property
+		EdgeRules     map[construct.SimpleEdge][]knowledgebase.OperationalRule
+		ResourceRules map[string][]knowledgebase.OperationalRule
```
gordon-klotho, line 23: What are these for? Resource rules that set a property should be in the property itself, no?

pkg/engine2/operational_eval/vertex_property_test.go
```diff
@@ -113,3 +113,103 @@ func Test_propertyVertex_evaluateEdgeOperational(t *testing.T) {
 		})
 	}
 }
+
+func Test_propertyVertex_Dependencies(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	dcap := NewMockdependencyCapturer(ctrl)
```
gordon-klotho, line 119: Should probably not share mocks across tests otherwise they could have implicit dependencies (eg, case 2 only works when run serially after case 1 runs its `mocks()`). IMO, change the test to `mocks func(*MockdependencyCapturer)` and create the mock per-test in the `Run`.

pkg/engine2/operational_eval/vertex_property_test.go
```diff
@@ -113,3 +113,103 @@ func Test_propertyVertex_evaluateEdgeOperational(t *testing.T) {
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			prop := &propertyVertex{
+				Ref:           tt.fields.Ref,
+				Template:      tt.fields.Template,
+				EdgeRules:     tt.fields.EdgeRules,
+				ResourceRules: tt.fields.ResourceRules,
+			}
```
gordon-klotho, line 203: This is a little awkward, why not just have `tt` have a `prop propertyVertex` field? Or some other nicer ways to slim down the test cases (like using strings instead of IDs / edges / etc).

pkg/engine2/operational_eval/vertex_resource_rule.go
```diff
@@ -0,0 +1,86 @@
...
+
+type (
+	resourceRuleVertex struct {
+		Resource construct.ResourceId
+		Rule     knowledgebase.AdditionalRule
+	}
+)
+
+func (v resourceRuleVertex) Key() Key {
+	return Key{Ref: construct.PropertyRef{Resource: v.Resource}, RuleHash: v.Rule.Hash()}
```
gordon-klotho, line 20: A little awkward to re-hash every call to `Key()` since we call it often.

pkg/engine2/operational_eval/vertex_resource_rule_test.go
```diff
@@ -0,0 +1,220 @@
...
+				},
+				RuleHash: "9a510ea0226e156085c51625d22ac3679eb7530fd36ba925fcbe211da9c1b373",
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert := assert.New(t)
+			got := tt.v.Key()
+			assert.Equal(tt.want, got)
```
gordon-klotho, line 39: I'm not sure this test is useful for what `Key()` is supposed to do. A more useful test would be that two keys with different rules have different keys regardless of what those keys actually resolve to.

pkg/engine2/operational_eval/vertex_resource_rule_test.go
```diff
@@ -0,0 +1,220 @@
...
+		})
+	}
+}
+
+func Test_resourceRuleVertex_Dependencies(t *testing.T) {
```
gordon-klotho, line 44: Side comment, nonblocking:

This test kinda demonstrates why I don't really like structural testing versus behavioural. Reading this test doesn't give any insight into how a resourceRule operates. There's no realistic example rules with expected output. The test is just for testing's sake, and doesn't assist in understanding (which I use in concert with documentation and reading the actual code).

pkg/engine2/operational_eval/vertex_resource_rule_test.go
```diff
@@ -0,0 +1,220 @@
...
+				If: "test",
+			},
+			mocks: func() {
+				dcap.EXPECT().ExecuteOpRule(knowledgebase.DynamicValueData{
+					Resource: construct.ResourceId{Name: "test"},
+				}, knowledgebase.OperationalRule{
+					If: "test",
+				}).Return(nil)
+			},
+		},
```
gordon-klotho, line 68: Missing a test case with other fields in the `AdditionalRule` to make sure those get translated into `OperationalRule` properly.

pkg/engine2/operational_eval/vertex_resource_rule_test.go
```diff
@@ -0,0 +1,220 @@
...
+			initial: &resourceRuleVertex{},
+			other:   &propertyVertex{},
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert := assert.New(t)
+			defer func() {
+				if r := recover(); r != nil {
```
gordon-klotho, line 154: Testing panics is just testing for coverage and doesn't actually give us any more confidence in the engine since a panic should never happen except by programming error. We don't test for index out of bounds or nil pointers either.

pkg/knowledge_base2/operational_rule.go
```diff
@@ -90,6 +99,24 @@ const (
 	ClosestSelectionOperator SelectionOperator = ""
 )
 
+func (rule AdditionalRule) Hash() string {
+	// Convert the struct to a byte slice.
+	// Note that the struct must be able to be converted to JSON,
+	// so all fields must be exported (i.e., start with a capital letter).
+	byteSlice, err := json.Marshal(rule)
+	if err != nil {
+		log.Fatal(err)
```
gordon-klotho, line 108: This is a disguised panic, not sure how I feel about that especially because we use `zap` for logging.

pkg/engine2/operational_eval/vertex_property_test.go
jhsinger-klotho: ctrl.Finish() is supposed to rest mocks, but i can move this

pkg/engine2/operational_eval/vertex_property.go
jhsinger-klotho: i think this was before i removed configuration from additional rules, so will remove this, its just leftover

pkg/engine2/operational_eval/vertex_resource_rule.go
jhsinger-klotho: i can create the hash as a part of reading them in, so will change that

pkg/engine2/operational_eval/vertex_resource_rule_test.go
jhsinger-klotho: i disagree, it assists in the sense of, we offload the determination of the dependencies by feeding the operational rule into the dependency capturer. If you want to know how the dependency capturer works, then thats a different story than this method

gordon-klotho (approved):

PR #893, opened 2024-01-18
Ecs fixes fg

jhsinger-klotho:
ecs fixes which got sample multi container deployments to work

### Standard checks


pkg/infra/iac3/templates/aws/ecs_task_definition/container_definitions.ts.tmpl
```diff
@@ -0,0 +1,103 @@
...
+                name: "{{ $value.Name }}",
+                value: {{ modelCase $value.Value }},
+            },
+            {{- end }}
+        ],
+        {{- end }}
+        {{- if $cd.DependsOn }}
+        dependsOn: [
+            {{- range $cd.DependsOn }}
+            {{- if $i }}
```
gordon-klotho, line 27: Is this `$i` referring to the right thing? Or do you need to assign a range variable

pkg/engine2/testdata/ecs_rds.expect.yaml
```diff
@@ -29,26 +29,33 @@ resources:
...
+                    awslogs-region: aws:region:region-0#Name
                     awslogs-stream-prefix: ecs_service_0-ecs_service_0
-              Memory: "512"
+              Memory: 512
               Name: ecs_service_0
               PortMappings:
                 - ContainerPort: 80
                   HostPort: 80
                   Protocol: TCP
+        Cpu: "256"
```
gordon-klotho, line 56: Are these (with Memory) supposed to be strings?

pkg/templates/aws/resources/region.yaml
```diff
@@ -2,7 +2,7 @@ qualified_type_name: aws:region
 display_name: Region
 
 properties:
-  RegionName:
+  Name:
```
gordon-klotho, line 5: The original property name was to disambiguate from the resource ID's Name field. We can't technically do `fieldValue` or `fieldRef` on those, but that may not be readily apparent and/or could be changed in the future.

pkg/infra/iac3/templates/aws/ecs_task_definition/container_definitions.ts.tmpl
```diff
@@ -0,0 +1,103 @@
...
+            ]
+            {{- end }}
+        },
+        {{- end }}
+        {{- if $cd.Memory }}
+        memory: {{ $cd.Memory }},
+        {{- end }}
+        {{- if $cd.MountPoints }}
+        mountPoints: [
+            {{- range $key, $value := $cd.MountPoints.M }}
```
gordon-klotho, line 72: Things like this make `HashedSet` seem weird. How would someone else writing iac template know to do this? Nothing actionable right now, just something to think about for later.

gordon-klotho (commented):
Only blocker is the `dependsOn` template bit.

pkg/engine2/testdata/ecs_rds.expect.yaml
jhsinger-klotho: yeah theyre strings on the base of the task definition

pkg/templates/aws/resources/region.yaml
jhsinger-klotho: i dont think we ever need fieldValue or fieldRef on the id name because you can access it via .Name and we sub the name in properties if it ever changes.

pkg/infra/iac3/templates/aws/ecs_task_definition/container_definitions.ts.tmpl
jhsinger-klotho: yeah i get that, we might need a better approach

pkg/infra/iac3/templates/aws/ecs_task_definition/container_definitions.ts.tmpl
jhsinger-klotho: yeah i cleaned up this template, also tested ecs -> efs because there was a fix needed for that as well

pkg/templates/aws/resources/region.yaml
gordon-klotho: Only case I could see for that might be for description-type fields like in a subnet rule or environment variable - ie, freeform things. And the only benefit over typing it out is that it'd get updated if the resource renamed. A niche case, so definitely not blocking, just digging up what I recall for why this was named that way.

pkg/templates/aws/resources/ecs_task_definition.yaml
```diff
@@ -43,14 +56,20 @@ properties:
...
+          are part of the task are stopped
       Image:
         type: resource(aws:ecr_image)
         required: true
         operational_rule:
           step:
             direction: downstream
             resources:
-              - aws:ecr_image:{{ .Self.Name }}-{{ fieldValue (printf "%s.Name" (pathAncestor .Path 1)) .Self}}
+              - aws:ecr_image:{{ .Self.Name }}-{{ if pathAncestorExists .Path 1}}{{ if hasField (printf "%s.Name" (pathAncestor .Path 1)) .Self}}{{ fieldValue (printf "%s.Name" (pathAncestor .Path 1)) .Self}}{{else}}image{{end}}{{else}}image{{end}}
```
gordon-klotho, line 72: I worry about using `fieldValue` for things that aren't ResourceIds since it won't work if it is imported. Not a concern for now, but just something we need to be aware of.

pkg/infra/iac3/templates/aws/rds_instance/factory.ts
```diff
@@ -49,7 +49,9 @@ function properties(object: aws.rds.Instance, args: Args) {
         }),
         RdsConnectionArn: pulumi.interpolate`arn:aws:rds-db:${region.name}:${accountId.accountId}:dbuser:${object.resourceId}/${object.username}`,
         Endpoint: object.endpoint,
-        ConnectionString: pulumi.interpolate`${args.Engine}://${object.username}:${object.password}@${object.endpoint.address}:${object.endpoint.port}/${args.DatabaseName}`,
+        ConnectionString: pulumi.interpolate`${args.Engine}://${object.username}:${object.password}@${object.endpoint}/${args.DatabaseName}`,
+        Host: object.address.apply((address) => address.split(':')[0]),
+        Port: object.address.apply((address) => address.split(':')[1]),
```
gordon-klotho, line 54: Pretty sure `address` is only the dns name, I think you want `endpoint` here which has the port.

gordon-klotho (commented):
Address vs Endpoint is the only blocking

gordon-klotho (approved):

PR #898, opened 2024-01-23
All deps

### LINKED ISSUE #885: Add secret names as environment variables on connection
Also, need to allow `<resource>#Name` as a property ref (currently not possible due to Name not being an official property but rather part of the ID).

### LINKED ISSUE #895: Downstream call returns duplicates
looks like our WalkDown function was changed in this commit https://github.com/klothoplatform/klotho/commit/3229c71be4cefa4ddb6e6d43adaf4c8a1b3aa1bd#diff-1048046ff9f8ffa2ae3375583203c3f69b1a13411c095bcb6f9e781869f07ac4

as a result im seeing many duplicates being returned when temmplates call allDownstream.

Image

reproducer
```
constraints:
- operator: must_exist
  scope: edge
  target:
    source: aws:cloudfront_distribution:cloudfront_distribution_2
    target: aws:rest_api:rest_api_0
resources:
    aws:api_stage:rest_api_0:api_stage-0:
        Deployment: aws:api_deployment:rest_api_0:api_deployment-0
        RestApi: aws:rest_api:rest_api_0
        StageName: stage
    aws:cloudfront_distribution:cloudfront_distribution_2:
        DefaultCacheBehavior:
            AllowedMethods:
                - DELETE
                - GET
                - HEAD
                - OPTIONS
                - PATCH
                - POST
                - PUT
            CachedMethods:
                - HEAD
                - GET
            DefaultTtl: 3600
            ForwardedValues:
                Cookies:
                    Forward: none
                QueryString: true
            MaxTtl: 86400
            MinTtl: 0
            ViewerProtocolPolicy: allow-all
        Enabled: true
        Restrictions:
            GeoRestriction:
                RestrictionType: none
        ViewerCertificate:
            CertificateSource: cloudfront
            MinimumProtocolVersion: TLSv1
            SSLSupportMethod: sni-only
    aws:api_deployment:rest_api_0:api_deployment-0:
        RestApi: aws:rest_api:rest_api_0
        Triggers:
            rest_api_0_integration_0: rest_api_0_integration_0
            rest_api_0_integration_0_method: rest_api_0_integration_0_method
    aws:rest_api:rest_api_0:
        BinaryMediaTypes:
            - application/octet-stream
            - image/*
    aws:api_resource:rest_api_0:api_resource-0:
        FullPath: /{proxy+}
        PathPart: '{proxy+}'
        RestApi: aws:rest_api:rest_api_0
    aws:api_method:rest_api_0:rest_api_0_integration_0_method:
        Authorization: NONE
        HttpMethod: ANY
        RequestParameters:
            method.request.path.proxy: true
        Resource: aws:api_resource:rest_api_0:api_resource-0
        RestApi: aws:rest_api:rest_api_0
    aws:api_integration:rest_api_0:rest_api_0_integration_0:
        IntegrationHttpMethod: ANY
        Method: aws:api_method:rest_api_0:rest_api_0_integration_0_method
        RequestParameters:
            integration.request.path.proxy: method.request.path.proxy
        Resource: aws:api_resource:rest_api_0:api_resource-0
        RestApi: aws:rest_api:rest_api_0
        Route: /{proxy+}
edges:
    aws:api_stage:rest_api_0:api_stage-0 -> aws:api_deployment:rest_api_0:api_deployment-0:
    aws:api_stage:rest_api_0:api_stage-0 -> aws:rest_api:rest_api_0:
    aws:api_deployment:rest_api_0:api_deployment-0 -> aws:api_integration:rest_api_0:rest_api_0_integration_0:
    aws:api_deployment:rest_api_0:api_deployment-0 -> aws:api_method:rest_api_0:rest_api_0_integration_0_method:
    aws:api_deployment:rest_api_0:api_deployment-0 -> aws:rest_api:rest_api_0:
    aws:rest_api:rest_api_0 -> aws:api_integration:rest_api_0:rest_api_0_integration_0:
    aws:rest_api:rest_api_0 -> aws:api_method:rest_api_0:rest_api_0_integration_0_method:
    aws:rest_api:rest_api_0 -> aws:api_resource:rest_api_0:api_resource-0:
    aws:api_resource:rest_api_0:api_resource-0 -> aws:api_integration:rest_api_0:rest_api_0_integration_0:
    aws:api_resource:rest_api_0:api_resource-0 -> aws:api_method:rest_api_0:rest_api_0_integration_0_method:
    aws:api_method:rest_api_0:rest_api_0_integration_0_method -> aws:api_integration:rest_api_0:rest_api_0_integration_0:
```

jhsinger-klotho:
adds secret id as auto env var (closes #885 )
makes sure all downstream doesnt return duplicates (closes #895 )

### Standard checks


pkg/knowledge_base2/graph.go
```diff
@@ -111,11 +112,13 @@ func firstFunctional(
 func allDeps(
 	ids *[]construct.ResourceId,
 ) graph_addons.WalkGraphFunc[construct.ResourceId] {
+	resourceSet := set.Set[construct.ResourceId]{}
```
gordon-klotho, line 115: Aren't the other functions/layers supposed to have this same behaviour? If so, might as well just change the `*[]construct.ResourceId` to just a `set.Set[construct.ResourceId]` which would simplify the code here a bit.

pkg/knowledge_base2/graph_test.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"testing"
+
+	construct "github.com/klothoplatform/klotho/pkg/construct2"
+	"github.com/klothoplatform/klotho/pkg/construct2/graphtest"
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/mock/gomock"
+)
+
+func TestDownstream(t *testing.T) {
```
gordon-klotho, line 12: There is already a `TestDownstream` in `kbtesting/graph_test.go` (the different package was to prevent circular dependencies)

pkg/knowledge_base2/graph.go
jhsinger-klotho: sure i can update to this

pkg/knowledge_base2/graph_test.go
jhsinger-klotho: we should try to keep testing local to the package otherwise its hard to know where it lives, but also there was clearly stuff missed from this tests so going to keep what i wrote

gordon-klotho: I agree we should try, but since `properties` depends on `knowledgebase2` the test can't use those unless it's in a different package. I don't think we should keep the unit tests for a function split in two places. It shouldn't be hard to add a test for duplication to the one in `kbtesting`.

pkg/knowledge_base2/graph.go
```diff
@@ -40,7 +41,7 @@ func resourceLocal(
 	dag construct.Graph,
 	kb TemplateKB,
 	rid construct.ResourceId,
-	ids *[]construct.ResourceId,
+	ids *set.Set[construct.ResourceId],
```
gordon-klotho, line 44: nit: `Set` is already a pointer, and this function doesn't initialize it (unlike the `*[]`)

pkg/knowledge_base2/graph.go
```diff
@@ -190,7 +191,7 @@ func Downstream(dag construct.Graph, kb TemplateKB, rid construct.ResourceId, la
 		return nil, fmt.Errorf("unknown layer %s", layer)
 	}
 	err := graph_addons.WalkDown(dag, rid, f)
-	return result, err
+	return result.ToSlice(), err
```
gordon-klotho, line 194: note: this will be non-deterministic, not sure if that's important or not

pkg/knowledge_base2/graph.go
```diff
@@ -235,16 +236,19 @@ func Upstream(dag construct.Graph, kb TemplateKB, rid construct.ResourceId, laye
...
 
 func layerWalkFunc(
 	dag construct.Graph,
 	kb TemplateKB,
 	rid construct.ResourceId,
 	layer DependencyLayer,
-	result []construct.ResourceId,
+	result set.Set[construct.ResourceId],
 ) (graph_addons.WalkGraphFunc[construct.ResourceId], error) {
+	if result == nil {
```
gordon-klotho, line 249: If `result == nil` then this function doesn't really do anything since it doesn't return the set. Actually, is this function used? I'm not sure the `[]ResourceId` argument would work either.

pkg/knowledge_base2/graph.go
jhsinger-klotho: i think thats fine, im not sure why we would need it to be deterministic

pkg/knowledge_base2/graph.go
jhsinger-klotho: its used in here so that arguement wasnt actually needed for the reason it was being used. I think thats the only place its used but dont want to refactor that as a part of this pr if we think that needs to change
```
// LayeredUpstream returns the first resource that matches `selector` which is upstream of `resource` for the specified layer
func (ctx DynamicValueContext) LayeredDownstream(
	selector any,
	resource construct.ResourceId,
	layer string,
) (construct.ResourceId, error) {
	selId, err := TemplateArgToRID(selector)
	if err != nil {
		return construct.ResourceId{}, err
	}

	dependencyLayer := DependencyLayer(layer)
	f, err := layerWalkFunc(ctx.Graph, ctx.KnowledgeBase, resource, dependencyLayer, nil)
	if err != nil {
		return construct.ResourceId{}, err
	}
	result := construct.ResourceId{}
	wrapper := func(path graph_addons.Path[construct.ResourceId], nerr error) error {
		id := path[len(path)-1]
		if selId.Matches(id) {
			result = id
			return graph_addons.StopWalk
		}
		return f(path, nerr)
	}
	err = graph_addons.WalkDown(ctx.Graph, resource, wrapper)
	if err != nil {
		return construct.ResourceId{}, err
	}
	return result, nil
}

// Downstream returns the fi
```

pkg/knowledge_base2/graph_test.go
jhsinger-klotho: we can mock properties if we need to, so i still think we should look to move it local to the package but for now i moved my tests into the kbtesting file

gordon-klotho: Sure we can mock properties, but seems overkill compared to just using the ones in `properties`. Just seems like extra maintenance for the only benefit of which package/folder the test lives in.

pkg/knowledge_base2/graph.go
gordon-klotho: That's fine then. When we add tests for `LayeredDownstream` it should uncover any issues if there are any.

pkg/knowledge_base2/template_kb_mock_test.go
```diff
@@ -0,0 +1,257 @@
...
+import (
+	reflect "reflect"
+
+	graph "github.com/dominikbraun/graph"
+	construct2 "github.com/klothoplatform/klotho/pkg/construct2"
+	gomock "go.uber.org/mock/gomock"
+)
+
+// MockTemplateKB is a mock of TemplateKB interface.
+type MockTemplateKB struct {
```
gordon-klotho, line 21: Does this replace the `MockKB` in `kbtesting`?

gordon-klotho (approved):

pkg/knowledge_base2/template_kb_mock_test.go
gordon-klotho: The reason the mock was put in that package (and without the `_test.go` suffix) was so that it could be used by tests outside of the package.

PR #899, opened 2024-01-23
Cdn behaviors

### LINKED ISSUE #887: Cloudfront -> API doesn't setup or allow configuration of behaviours

jhsinger-klotho:
adds more configuration to cloudfront distributions (closes #887 )

added flag to edge template to know when an edge will always need to be processed (this is for situations where graph state is determining on when configuration needs to happen) (api triggers and cdn behaviors)

makes map add default values in parsing transformation if they dont exist in the keys
mnakes int property parsing convert floats
added unit tests for the above situation

Screen Shot 2024-01-23 at 9 00 45 AM
Screen Shot 2024-01-23 at 9 01 01 AM

example iac
```
import * as aws from '@pulumi/aws'
import * as awsInputs from '@pulumi/aws/types/input'
import * as command from '@pulumi/command'
import * as docker from '@pulumi/docker'
import * as pulumi from '@pulumi/pulumi'


const kloConfig = new pulumi.Config('klo')
const protect = kloConfig.getBoolean('protect') ?? false
const awsConfig = new pulumi.Config('aws')
const awsProfile = awsConfig.get('profile')
const accountId = pulumi.output(aws.getCallerIdentity({}))
const region = pulumi.output(aws.getRegion({}))

const ecr_repo_0 = new aws.ecr.Repository("ecr_repo-0", {
        imageScanningConfiguration: {
            scanOnPush: true,
        },
        imageTagMutability: 'MUTABLE',
        forceDelete: true,
        encryptionConfigurations: [{ encryptionType: 'KMS' }],
        tags: {
            env: 'production',
            AppName: "ecr_repo-0",
        },
    })
const lambda_function_5_log_group = new aws.cloudwatch.LogGroup("lambda_function_5-log-group", {
        name: "/aws/lambda/lambda_function_5",
        retentionInDays: 5,
    })
const rest_api_0 = new aws.apigateway.RestApi("rest_api_0", {
        binaryMediaTypes: ["application/octet-stream", "image/*"],
    })
const lambda_function_5_image = (() => {
        const base = new docker.Image(
            `${"lambda_function_5-image"}-base`,
            {
                build: {
                    context: ".",
                    dockerfile: "lambda_function_5-image.Dockerfile",
                    platform: 'linux/amd64',
                },
                skipPush: true,
                imageName: pulumi.interpolate`${ecr_repo_0.repositoryUrl}:base`,
            },
        )

        const sha256 = new command.local.Command(
            `${"lambda_function_5-image"}-base-get-sha256-${Date.now()}`,
            { create: pulumi.interpolate`docker image inspect -f {{.ID}} ${base.imageName}` },
            { parent: base }
        ).stdout.apply((id) => id.substring(7))

        return new docker.Image(
            "lambda_function_5-image",
            {
                build: {
                    context: ".",
                    dockerfile: "lambda_function_5-image.Dockerfile",
                    platform: 'linux/amd64',
                },
                registry: aws.ecr
                    .getAuthorizationTokenOutput(
                        { registryId: ecr_repo_0.registryId },
                        { async: true }
                    )
                    .apply((registryToken) => {
                        return {
                            server: ecr_repo_0.repositoryUrl,
                            username: registryToken.userName,
                            password: registryToken.password,
                        }
                    }),
                imageName: pulumi.interpolate`${ecr_repo_0.repositoryUrl}:${sha256}`,
            },
            { parent: base }
        )
    })()
const lambda_function_5_executionrole = new aws.iam.Role("lambda_function_5-ExecutionRole", {
        assumeRolePolicy: pulumi.jsonStringify({Statement: [{Action: ["sts:AssumeRole"], Effect: "Allow", Principal: {Service: ["lambda.amazonaws.com"]}}], Version: "2012-10-17"}),
        managedPolicyArns: [
            ...["arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole"],
        ],
    })
const api_resource_2 = new aws.apigateway.Resource(
        "api_resource-2",
        {
            restApi: rest_api_0.id,
            parentId: rest_api_0.rootResourceId,
            pathPart: "something",
        },
        { parent: rest_api_0 }
    )
const lambda_function_5 = new aws.lambda.Function(
        "lambda_function_5",
        {
            packageType: 'Image',
            imageUri: lambda_function_5_image.imageName,
            memorySize: 512,
            timeout: 180,
            role: lambda_function_5_executionrole.arn,
            name: "lambda_function_5",
            tags: {
                env: 'production',
                service: "lambda_function_5",
            },
        },
        {
            dependsOn: [lambda_function_5_executionrole, lambda_function_5_image, lambda_function_5_log_group],
        }
    )
const api_resource_1 = new aws.apigateway.Resource(
        "api_resource-1",
        {
            restApi: rest_api_0.id,
            parentId: rest_api_0.rootResourceId,
            pathPart: "items",
        },
        { parent: rest_api_0 }
    )
const rest_api_0_integration_0_method = new aws.apigateway.Method(
        "rest_api_0_integration_0_method",
        {
            restApi: rest_api_0.id,
            resourceId: api_resource_2.id,
            httpMethod: "POST",
            authorization: "NONE",
            requestParameters: {"method.request.path.proxy": true},
        },
        {
            parent: api_resource_2,
        }
    )
const rest_api_0_integration_0_lambda_function_5 = new aws.lambda.Permission("rest_api_0_integration_0-lambda_function_5", {
        action: "lambda:InvokeFunction",
        function: lambda_function_5.name,
        principal: "apigateway.amazonaws.com",
        sourceArn: pulumi.interpolate`${rest_api_0.executionArn}/*`,
    })
const rest_api_0_integration_0_lambda_function_5_0 = new aws.lambda.Permission("rest_api_0_integration_0-lambda_function_5-0", {
        action: "lambda:InvokeFunction",
        function: lambda_function_5.name,
        principal: "apigateway.amazonaws.com",
        sourceArn: pulumi.interpolate`${rest_api_0.executionArn}/*`,
    })
const rest_api_0_integration_0_lambda_function_5_1 = new aws.lambda.Permission("rest_api_0_integration_0-lambda_function_5-1", {
        action: "lambda:InvokeFunction",
        function: lambda_function_5.name,
        principal: "apigateway.amazonaws.com",
        sourceArn: pulumi.interpolate`${rest_api_0.executionArn}/*`,
    })
const rest_api_0_integration_1_method = new aws.apigateway.Method(
        "rest_api_0_integration_1_method",
        {
            restApi: rest_api_0.id,
            resourceId: api_resource_1.id,
            httpMethod: "GET",
            authorization: "NONE",
        },
        {
            parent: api_resource_1,
        }
    )
const rest_api_0_integration_1_lambda_function_5 = new aws.lambda.Permission("rest_api_0_integration_1-lambda_function_5", {
        action: "lambda:InvokeFunction",
        function: lambda_function_5.name,
        principal: "apigateway.amazonaws.com",
        sourceArn: pulumi.interpolate`${rest_api_0.executionArn}/*`,
    })
const rest_api_0_integration_1_lambda_function_5_0 = new aws.lambda.Permission("rest_api_0_integration_1-lambda_function_5-0", {
        action: "lambda:InvokeFunction",
        function: lambda_function_5.name,
        principal: "apigateway.amazonaws.com",
        sourceArn: pulumi.interpolate`${rest_api_0.executionArn}/*`,
    })
const rest_api_0_integration_1_lambda_function_5_1 = new aws.lambda.Permission("rest_api_0_integration_1-lambda_function_5-1", {
        action: "lambda:InvokeFunction",
        function: lambda_function_5.name,
        principal: "apigateway.amazonaws.com",
        sourceArn: pulumi.interpolate`${rest_api_0.executionArn}/*`,
    })
const rest_api_0_integration_2_method = new aws.apigateway.Method(
        "rest_api_0_integration_2_method",
        {
            restApi: rest_api_0.id,
            resourceId: api_resource_2.id,
            httpMethod: "GET",
            authorization: "NONE",
        },
        {
            parent: api_resource_2,
        }
    )
const rest_api_0_integration_2_lambda_function_5 = new aws.lambda.Permission("rest_api_0_integration_2-lambda_function_5", {
        action: "lambda:InvokeFunction",
        function: lambda_function_5.name,
        principal: "apigateway.amazonaws.com",
        sourceArn: pulumi.interpolate`${rest_api_0.executionArn}/*`,
    })
const rest_api_0_integration_0 = new aws.apigateway.Integration(
        "rest_api_0_integration_0",
        {
            restApi: rest_api_0.id,
            resourceId: api_resource_2.id,
            httpMethod: rest_api_0_integration_0_method.httpMethod,
            integrationHttpMethod: "POST",
            type: "AWS_PROXY",
            uri: lambda_function_5.invokeArn,
        },
        { parent: rest_api_0_integration_0_method }
    )
const rest_api_0_integration_1 = new aws.apigateway.Integration(
        "rest_api_0_integration_1",
        {
            restApi: rest_api_0.id,
            resourceId: api_resource_1.id,
            httpMethod: rest_api_0_integration_1_method.httpMethod,
            integrationHttpMethod: "POST",
            type: "AWS_PROXY",
            uri: lambda_function_5.invokeArn,
        },
        { parent: rest_api_0_integration_1_method }
    )
const rest_api_0_integration_2 = new aws.apigateway.Integration(
        "rest_api_0_integration_2",
        {
            restApi: rest_api_0.id,
            resourceId: api_resource_2.id,
            httpMethod: rest_api_0_integration_2_method.httpMethod,
            integrationHttpMethod: "POST",
            type: "AWS_PROXY",
            uri: lambda_function_5.invokeArn,
        },
        { parent: rest_api_0_integration_2_method }
    )
const api_deployment_0 = new aws.apigateway.Deployment(
        "api_deployment-0",
        {
            restApi: rest_api_0.id,
            triggers: {restApi0Integration0: "rest_api_0_integration_0", restApi0Integration0Method: "rest_api_0_integration_0_method", restApi0Integration1: "rest_api_0_integration_1", restApi0Integration1Method: "rest_api_0_integration_1_method", restApi0Integration2: "rest_api_0_integration_2", restApi0Integration2Method: "rest_api_0_integration_2_method"},
        },
        {
            dependsOn: [rest_api_0, rest_api_0_integration_0, rest_api_0_integration_0_method, rest_api_0_integration_1, rest_api_0_integration_1_method, rest_api_0_integration_2, rest_api_0_integration_2_method],
        }
    )
const api_stage_0 = new aws.apigateway.Stage("api_stage-0", {
        deployment: api_deployment_0.id,
        restApi: rest_api_0.id,
        stageName: "stage",
    })
export const api_stage_0_Url = api_stage_0.invokeUrl
const cloudfront_distribution_2 = new aws.cloudfront.Distribution("cloudfront_distribution_2", {
        origins: [{customOriginConfig: {httpPort: 80, httpsPort: 443, originProtocolPolicy: "https-only", originSslProtocols: ["TLSv1.2", "TLSv1", "SSLv3", "TLSv1.1"]}, domainName: api_stage_0.invokeUrl.apply((d) => d.split('//')[1].split('/')[0]), originId: "api_stage-0", originPath: "/stage"}, {customOriginConfig: {httpPort: 80, httpsPort: 443, originProtocolPolicy: "https-only", originSslProtocols: ["TLSv1.2", "TLSv1", "SSLv3", "TLSv1.1"]}, domainName: api_stage_0.invokeUrl.apply((d) => d.split('//')[1].split('/')[0]), originId: "api_stage-0", originPath: "/stage"}],
        enabled: true,
        viewerCertificate: {cloudfrontDefaultCertificate: true},
        orderedCacheBehaviors: [{allowedMethods: ["HEAD", "GET", "OPTIONS"], cachedMethods: ["HEAD", "GET"], defaultTtl: 0, forwardedValues: {cookies: {forward: "none"}, queryString: true}, maxTtl: 0, minTtl: 0, pathPattern: "/items", smoothStreaming: false, targetOriginId: "api_stage-0", viewerProtocolPolicy: "redirect-to-https"}, {allowedMethods: ["HEAD", "OPTIONS", "PATCH", "POST", "PUT", "DELETE", "GET"], cachedMethods: ["GET", "HEAD"], defaultTtl: 0, forwardedValues: {cookies: {forward: "none"}, queryString: true}, maxTtl: 0, minTtl: 0, pathPattern: "/something", smoothStreaming: false, targetOriginId: "api_stage-0", viewerProtocolPolicy: "redirect-to-https"}],
        defaultCacheBehavior: {
            ...{allowedMethods: ["DELETE", "GET", "HEAD", "OPTIONS", "PATCH", "POST", "PUT"], cachedMethods: ["HEAD", "GET"], defaultTtl: 3600, forwardedValues: {cookies: {forward: "none"}, queryString: true}, maxTtl: 86400, minTtl: 0, viewerProtocolPolicy: "allow-all"},
            targetOriginId: "api_stage-0",
        },
        restrictions: {geoRestriction: {restrictionType: "none"}},
    })
export const cloudfront_distribution_2_Domain = cloudfront_distribution_2.domainName

```
### Standard checks

### DIFF
--- a/pkg/knowledge_base2/edge_template.go
+++ b/pkg/knowledge_base2/edge_template.go
@@ -12,6 +12,8 @@ type (
 		Source construct.ResourceId `yaml:"source"`
 		Target construct.ResourceId `yaml:"target"`
 
+		AlwaysProcess bool `yaml:"always_process"`
+
 		// DirectEdgeOnly signals that the edge cannot be used within constructing other paths
 		// and can only be used as a direct edge
 		DirectEdgeOnly bool `yaml:"direct_edge_only"`
--- a/pkg/knowledge_base2/edge_template.go
+++ b/pkg/knowledge_base2/edge_template.go
@@ -12,6 +12,8 @@ type (
 		Source construct.ResourceId `yaml:"source"`
 		Target construct.ResourceId `yaml:"target"`
 
+		// AlwaysProcess signals that the edge should always be processed even if the source and target exist in the input graph
+		// currently we dont check edges for operational rules if they previously existed and this flag is set to false
 		AlwaysProcess bool `yaml:"always_process"`
 
 		// DirectEdgeOnly signals that the edge cannot be used within constructing other paths
--- a/pkg/knowledge_base2/properties/int_property.go
+++ b/pkg/knowledge_base2/properties/int_property.go
@@ -2,6 +2,7 @@ package properties
 
 import (
 	"fmt"
+	"math"
 
 	construct "github.com/klothoplatform/klotho/pkg/construct2"
 	knowledgebase "github.com/klothoplatform/klotho/pkg/knowledge_base2"
@@ -66,9 +67,15 @@ func (i *IntProperty) Parse(value any, ctx knowledgebase.DynamicContext, data kn
 		return val, nil
 	}
 	if val, ok := value.(float32); ok {
-		return int(val), nil
+		if float64(val) == math.Trunc(float64(val)) {
+			return int(val), nil
+		}
+		return nil, fmt.Errorf("invalid int value %v, of type %T", value, value)
 	} else if val, ok := value.(float64); ok {
-		return int(val), nil
+		if val == math.Trunc(val) {
+			return int(val), nil
+		}
+		return nil, fmt.Errorf("invalid int value %v, of type %T", value, value)
 	}
 	val, err := ParsePropertyRef(value, ctx, data)
 	if err == nil {
--- a/pkg/knowledge_base2/properties/int_property.go
+++ b/pkg/knowledge_base2/properties/int_property.go
@@ -65,6 +65,11 @@ func (i *IntProperty) Parse(value any, ctx knowledgebase.DynamicContext, data kn
 	if val, ok := value.(int); ok {
 		return val, nil
 	}
+	if val, ok := value.(float32); ok {
+		return int(val), nil
+	} else if val, ok := value.(float64); ok {
+		return int(val), nil
+	}
 	val, err := ParsePropertyRef(value, ctx, data)
 	if err == nil {
 		return val, nil
--- a/pkg/knowledge_base2/properties/int_property.go
+++ b/pkg/knowledge_base2/properties/int_property.go
@@ -58,6 +58,7 @@ func (i *IntProperty) GetDefaultValue(ctx knowledgebase.DynamicContext, data kno
 }
 
 func (i *IntProperty) Parse(value any, ctx knowledgebase.DynamicContext, data knowledgebase.DynamicValueData) (any, error) {
+
 	if val, ok := value.(string); ok {
 		var result int
 		err := ctx.ExecuteDecode(val, data, &result)
@@ -66,16 +67,20 @@ func (i *IntProperty) Parse(value any, ctx knowledgebase.DynamicContext, data kn
 	if val, ok := value.(int); ok {
 		return val, nil
 	}
+	EPSILON := 0.0000001
 	if val, ok := value.(float32); ok {
-		if float64(val) == math.Trunc(float64(val)) {
-			return int(val), nil
+		ival := int(val)
+		if math.Abs(float64(val)-float64(ival)) > EPSILON {
+			return 0, fmt.Errorf("cannot convert non-integral float to int: %f", val)
 		}
-		return nil, fmt.Errorf("invalid int value %v, of type %T", value, value)
+		return int(val), nil
+
 	} else if val, ok := value.(float64); ok {
-		if val == math.Trunc(val) {
-			return int(val), nil
+		ival := int(val)
+		if math.Abs(val-float64(ival)) > EPSILON {
+			return 0, fmt.Errorf("cannot convert non-integral float to int: %f", val)
 		}
-		return nil, fmt.Errorf("invalid int value %v, of type %T", value, value)
+		return int(val), nil
 	}
 	val, err := ParsePropertyRef(value, ctx, data)
 	if err == nil {
--- a/pkg/knowledge_base2/properties/map_property.go
+++ b/pkg/knowledge_base2/properties/map_property.go
@@ -112,6 +112,16 @@ func (m *MapProperty) Parse(value any, ctx knowledgebase.DynamicContext, data kn
 					continue
 				}
 				result[key] = val
+			} else {
+				val, err := prop.GetDefaultValue(ctx, data)
+				if err != nil {
+					errs = errors.Join(errs, fmt.Errorf("unable to get default value for sub property %s: %w", key, err))
+					continue
+				}
+				if val == nil {
+					continue
+				}
+				result[key] = val
 			}
 		}
 	}
--- a/pkg/engine2/solution_context.go
+++ b/pkg/engine2/solution_context.go
@@ -2,6 +2,7 @@ package engine2
 
 import (
 	"errors"
+	"fmt"
 
 	construct "github.com/klothoplatform/klotho/pkg/construct2"
 	"github.com/klothoplatform/klotho/pkg/engine2/constraints"
@@ -90,8 +91,25 @@ func (ctx solutionContext) LoadGraph(graph construct.Graph) error {
 	if err := op.AddVerticesFrom(graph); err != nil {
 		return err
 	}
-	if err := raw.AddEdgesFrom(graph); err != nil {
+
+	edges, err := graph.Edges()
+	if err != nil {
 		return err
+	}
+	for _, edge := range edges {
+		edgeTemplate := ctx.KB.GetEdgeTemplate(edge.Source, edge.Target)
+		if edgeTemplate == nil {
+			return fmt.Errorf("edge template %s -> %s not found", edge.Source, edge.Target)
+		}
+		if edgeTemplate.AlwaysProcess {
+			if err := op.AddEdge(edge.Source, edge.Target); err != nil {
+				return err
+			}
+		} else {
+			if err := raw.AddEdge(edge.Source, edge.Target); err != nil {
+				return err
+			}
+		}
 	}
 
 	// ensure any deployment dependencies due to properties are in place
--- a/pkg/engine2/testdata/cf_distribution.expect.yaml
+++ b/pkg/engine2/testdata/cf_distribution.expect.yaml
@@ -1,6 +1,8 @@
 resources:
     aws:cloudfront_distribution:cloudfront_distribution_2:
-        CloudfrontDefaultCertificate: true
+        ViewerCertificate:
+            CloudfrontDefaultCertificate: true
+        CacheBehaviors: []
         DefaultCacheBehavior:
             AllowedMethods:
                 - DELETE
--- a/pkg/engine2/testdata/same_path_multiple_methods.expect.yaml
+++ b/pkg/engine2/testdata/same_path_multiple_methods.expect.yaml
@@ -8,6 +8,10 @@ resources:
         Triggers:
             rest_api_0_integration_0: rest_api_0_integration_0
             rest_api_0_integration_0_method: rest_api_0_integration_0_method
+            rest_api_0_integration_1: rest_api_0_integration_1
+            rest_api_0_integration_1_method: rest_api_0_integration_1_method
+            rest_api_0_integration_2: rest_api_0_integration_2
+            rest_api_0_integration_2_method: rest_api_0_integration_2_method
     aws:rest_api:rest_api_0:
         BinaryMediaTypes:
             - application/octet-stream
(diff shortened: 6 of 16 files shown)


pkg/knowledge_base2/edge_template.go
```diff
@@ -12,6 +12,8 @@ type (
 		Source construct.ResourceId `yaml:"source"`
 		Target construct.ResourceId `yaml:"target"`
 
+		AlwaysProcess bool `yaml:"always_process"`
```
gordon-klotho, line 15: Can you godoc this? What does it mean?

I understand its purpose, but couldn't we determine whether it needs to be recomputed based on dependencies that changed? Is this just a temporary solution for that?

pkg/knowledge_base2/properties/int_property.go
```diff
@@ -65,6 +65,11 @@ func (i *IntProperty) Parse(value any, ctx knowledgebase.DynamicContext, data kn
 	if val, ok := value.(int); ok {
 		return val, nil
 	}
+	if val, ok := value.(float32); ok {
+		return int(val), nil
+	} else if val, ok := value.(float64); ok {
+		return int(val), nil
+	}
```
gordon-klotho, line 72: Not sure I'm a fan of `3.2` being a valid integer and truncating versus throwing a validation error. I don't mind `3.0` getting converted though.

pkg/knowledge_base2/edge_template.go
jhsinger-klotho: we never add edges to the operational view so they never get recomputed if they already exist. This is an existing bug for deployment triggers i realized and also happened when doing cdn behaviors. Ill add a comment, but seems like the amount of work to approach this in a programmatic manner is really high

pkg/knowledge_base2/properties/int_property.go
jhsinger-klotho: so you just want to validate that the float value ends in .0?

gordon-klotho: Well, it's not quite that simple in floating point. Probably just check:
```
ival := int(val)
if abs(float64(ival) - val) > EPSILON { // EPSILON is some low number, like 0.00001
    return 0, fmt.Errorf("cannot convert non-integral float to int: %f", val)
}
```

pkg/knowledge_base2/edge_template.go
gordon-klotho: Because this would mean that if a user added their own custom behaviour, that it would get overwritten the next time the engine ran, right?

pkg/knowledge_base2/properties/int_property.go
```diff
@@ -65,6 +66,17 @@ func (i *IntProperty) Parse(value any, ctx knowledgebase.DynamicContext, data kn
 	if val, ok := value.(int); ok {
 		return val, nil
 	}
+	if val, ok := value.(float32); ok {
+		if float64(val) == math.Trunc(float64(val)) {
```
gordon-klotho, line 70: Like I mentioned in the general comment, `==` is not a safe operation on floating-point numbers.

pkg/knowledge_base2/edge_template.go
jhsinger-klotho: no since our configuration rules only ever do append it should still be there

pkg/knowledge_base2/properties/int_property.go
jhsinger-klotho: out of curiousity, why isnt this a safe operation?

gordon-klotho: Because floating point numbers are weird and in some circumstances have multiple ways of representing the same number. They can have rounding errors that could be incurred during parsing from string or any operations such as via template functions.

pkg/knowledge_base2/edge_template.go
gordon-klotho: Would this make it impossible to only expose a part of a Rest API via the cloudfront distribution?

pkg/knowledge_base2/properties/int_property.go
jhsinger-klotho: but wouldnmt the approach of checking if  abs(float64(ival) - val) > EPSILON  have some similar errors in terms, ill implement the other route though

pkg/knowledge_base2/edge_template.go
jhsinger-klotho: currently yes, if the api is not the default origin. if we want cdn to have a edge to the integration thats another route, but i dont think thats as intuitive.

pkg/knowledge_base2/properties/int_property.go
gordon-klotho: It's to cover the case that, for example, `0.4+0.4+0.2 == 0.9999999`. By using an EPSILON instead of `==`, it would allow for small amounts of accumulated errors.

pkg/knowledge_base2/edge_template.go
gordon-klotho: I think if it's CDN -> Rest API, then it should just have a `ANY /*` behaviour to indicate that it's pass-through. If it's per-route, then it should be CDN -> Integration corresponding with that specific route. This would be the most intuitive to me, and the most flexible as it would be capable of both forms.

jhsinger-klotho: i disagree otherwise you cant have multiple origins

jhsinger-klotho: by doing the integrations its at least allowing multiple origins as long as their paths dont conflict. This is the ux aaron and i had discussed so if we want to change it we can bring it up in the channel

pkg/knowledge_base2/properties/map_property.go
```diff
@@ -112,6 +112,16 @@ func (m *MapProperty) Parse(value any, ctx knowledgebase.DynamicContext, data kn
 					continue
 				}
 				result[key] = val
+			} else {
+				val, err := prop.GetDefaultValue(ctx, data)
```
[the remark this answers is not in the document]
jhsinger-klotho: i guess it depends if we think that our templates are setup in a manner to support that

in the template below, you wouldnt be able to have the default_value on cookies, it would have to live on the ForwardedValues, otherwise you may never evaluate Cookies
```
      ForwardedValues:
        type: map
        properties:
          QueryString:
            type: bool
            default_value: true
          Cookies:
            type: map
            properties:
              Forward:
                type: string
                default_value: none
```

so the template would have to look like.
```
      ForwardedValues:
        type: map
        default_value:
          Cookies: none
        properties:
          QueryString:
            type: bool
            default_value: true
          Cookies:
            type: map
            properties:
              Forward:
                type: string
                default_value: none
```

Theres the potential for this to cause issues if people use the append command instead of the set command because then they may append certain values that are incompatible (which validation would catch) but there would be no possible way for them to ever initialize ForwardedValues with QueryString= false if that conflicts with Cookies=none.

doesnt our property deps finder run the TransformPropertyValue for the default Value? In which case it should pick up on nested dependencies if theyre templated

PR #902, opened 2024-01-24
Op eval tests

### LINKED ISSUE #846: Improve Unit Test for operational eval to 20%

jhsinger-klotho:
closes #846 

adding unit tests to op eval, causing some refactor for interfaces to make code unit testable

### Standard checks


pkg/engine2/operational_eval/vertex_property_test.go
```diff
@@ -230,3 +231,136 @@ func Test_propertyVertex_Dependencies(t *testing.T) {
...
+		mocks   func(mockProperty *MockProperty, mockRc *MockResourceConfigurer, data testData)
+		wantErr bool
+	}
+	tests := []testData{
+		{
+			name: "existing value and no constraints",
+			ref: construct.PropertyRef{
+				Resource: id,
+				Property: "test",
+			},
```
gordon-klotho, line 251: nit: could use `graphtest.ParseRef` to let these be strings to be more compact without losing readability

pkg/engine2/testdata/vpc_import_to_lambda.dataflow-viz.yaml
```diff
@@ -4,6 +4,7 @@ resources:
     children:
         - aws:security_group:vpc:lambda_function-security_group
         - aws:subnet:vpc:subnet1
+        - aws:subnet:vpc:subnet2
```
gordon-klotho, line 7: I assume this changed due to the addition of the namespace (shown in the `.expect.yaml`)

pkg/engine2/testdata/vpc_import_to_lambda.expect.yaml
```diff
@@ -31,7 +31,7 @@ resources:
             - aws:security_group:vpc:lambda_function-security_group
         Subnets:
             - aws:subnet:vpc:subnet1
-            - aws:subnet:subnet2
+            - aws:subnet:vpc:subnet2
```
gordon-klotho, line 34: Was this fixed by the `&& setConstraint.Operator == ""` to the vertex property?

pkg/knowledge_base2/kb.go
```diff
@@ -12,6 +12,7 @@ import (
 )
 
 //go:generate mockgen --source=./kb.go -destination=./template_kb_mock_test.go -package=knowledgebase2
+//go:generate mockgen --source=./kb.go -destination=../engine2/operational_eval/template_kb_mock_test.go -package=operational_eval
```
gordon-klotho, line 15: This is really awkward - should just generate one mock. That's why there's the `graphtest`, `enginetest`, `kbtesting` etc packages. It's also a little awkward, but it's standard for Go for reusable test utilities (see for eg. [httptest](https://pkg.go.dev/net/http/httptest) in the stdlib).

gordon-klotho (approved):
Nothing blocking

pkg/knowledge_base2/kb.go
jhsinger-klotho: well get into the circular dependencies again. Theres no way to test the actual solution context package for example. but we could try a different method, i just think this gets rid of that worry

pkg/engine2/testdata/vpc_import_to_lambda.expect.yaml
jhsinger-klotho: yup

PR #904, opened 2024-01-29
Increase path_selection coverage to > 10%

gordon-klotho:
```
❯ go test -cover ./pkg/engine2/path_selection
ok      github.com/klothoplatform/klotho/pkg/engine2/path_selection     0.009s  coverage: 13.9% of statements
```

### Standard checks

- **Unit tests**: ✅ 
- **Docs**: n/a

### DIFF
--- a/pkg/engine2/path_selection/candidate_weight_test.go
+++ b/pkg/engine2/path_selection/candidate_weight_test.go
@@ -1,0 +1,159 @@
+package path_selection
+
+import (
+	"testing"
+
+	construct "github.com/klothoplatform/klotho/pkg/construct2"
+	"github.com/klothoplatform/klotho/pkg/construct2/graphtest"
+	"github.com/klothoplatform/klotho/pkg/engine2/enginetesting"
+	knowledgebase "github.com/klothoplatform/klotho/pkg/knowledge_base2"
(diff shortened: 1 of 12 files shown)


pkg/engine2/path_selection/candidate_weight_test.go
```diff
@@ -0,0 +1,159 @@
...
+			wantWeight: 21,
+		},
+		{
+			name:       "downstream indirect",
+			graph:      []any{"p:compute:a -> p:glue:b -> p:glue:c -> p:glue:d -> p:compute:e"},
+			src:        "p:compute:a",
+			target:     "p:compute:e",
+			id:         "p:glue:c",
+			wantWeight: 15,
+		},
```
jhsinger-klotho, line 60: should also add cases for in graph, but unrelated to downstream/upstream (checking that it uses the undirected graph to determine weight)

gordon-klotho: 👍 was planning on doing that (among other changes) for the +7% this week

jhsinger-klotho (approved):
not blocking, but i dont think the unit tests really cover that much of the functionality of those methods because they only have 2 simple cases, so we should add more to those. the 13% seems like it might be because a lot of the code path is touched although we need to add in tests for more variation of how we would expect it to behave given certain inputs

PR #919, opened 2024-02-01
Lambda to lambda

jhsinger-klotho:
Changes ECR Image to want a unique repo so we can deploy multiple images in the same architecture

adds service discovery back to ecs to be used in applications to understand what ip to hit when trying to make cross compute calls

tests to ecs have issues at the moment due to operational eval and incorrect dependencies

lambda -> lambda
```
(engine-e2e-tests) (base) jordansinger@Jordans-MacBook-Pro engine-e2e-tests % ENGINE_PATH=/Users/jordansinger/workspace/klotho/engine IAC_CLI_PATH=/Users/jordansinger/workspace/klotho/iac e2e run --region eu-west-1 --include aws/lambda/lambda

============================= test session starts ==============================
platform darwin -- Python 3.10.13, pytest-7.4.3, pluggy-1.3.0
rootdir: /Users/jordansinger/workspace/engine-e2e-tests/output/aws-rbkjwv
asyncio: mode=strict
created: 1/1 worker
1 worker [1 item]

.                                                                        [100%]
- generated xml file: /var/folders/n_/bbppnfk179d6l4vz4qmz0chh0000gn/T/tmpifycrags -
============================== 1 passed in 3.77s ===============================
```

ecs -> lambda
```
(engine-e2e-tests) (base) jordansinger@Jordans-MacBook-Pro engine-e2e-tests % ENGINE_PATH=/Users/jordansinger/workspace/klotho/engine IAC_CLI_PATH=/Users/jordansinger/workspace/klotho/iac e2e run --region eu-west-1 --include aws/ecs_fargate/lambda

============================= test session starts ==============================
platform darwin -- Python 3.10.13, pytest-7.4.3, pluggy-1.3.0
rootdir: /Users/jordansinger/workspace/engine-e2e-tests/output/aws-3rwbox
asyncio: mode=strict
created: 1/1 worker
1 worker [1 item]

.                                                                        [100%]
- generated xml file: /var/folders/n_/bbppnfk179d6l4vz4qmz0chh0000gn/T/tmpps1tlhs6 -
============================== 1 passed in 2.03s ===============================
```

### Standard checks

### DIFF
--- a/pkg/knowledge_base2/properties/property.go
+++ b/pkg/knowledge_base2/properties/property.go
@@ -73,6 +73,18 @@ func ValidatePropertyRef(value construct.PropertyRef, propertyType string, ctx k
 	if err != nil {
 		return nil, fmt.Errorf("error getting property %s on resource %s, while validating property ref: %w", value.Property, value.Resource, err)
 	}
+
+	// recurse down in case of a nested property ref
+	for _, ok := propVal.(construct.PropertyRef); ok; _, ok = propVal.(construct.PropertyRef) {
+		propVal, err = ValidatePropertyRef(propVal.(construct.PropertyRef), propertyType, ctx)
+		if err != nil {
+			return nil, err
+		}
+		if propVal == nil {
+			return nil, nil
+		}
+	}
+
 	return propVal, nil
 }
 
--- a/pkg/templates/aws/edges/ecr_image-ecr_repo.yaml
+++ b/pkg/templates/aws/edges/ecr_image-ecr_repo.yaml
@@ -1,2 +1,3 @@
 source: aws:ecr_image
 target: aws:ecr_repo
+unique: one-to-one
--- a/pkg/templates/aws/resources/ecr_image.yaml
+++ b/pkg/templates/aws/resources/ecr_image.yaml
@@ -16,6 +16,7 @@ properties:
         direction: downstream
         resources:
           - aws:ecr_repo
+        unique: true
     description: A reference to the AWS ECR repository resource where the image will
       be stored
   Context:
--- a/pkg/engine2/testdata/2_routes.dataflow-viz.yaml
+++ b/pkg/engine2/testdata/2_routes.dataflow-viz.yaml
@@ -3,14 +3,14 @@ resources:
   lambda_function/lambda_function_1:
     children:
         - aws:ecr_image:lambda_function_1-image
-        - aws:ecr_repo:ecr_repo-0
+        - aws:ecr_repo:lambda_function_1-image-ecr_repo
         - aws:iam_role:lambda_function_1-ExecutionRole
     tag: big
 
   lambda_function/lambda_function_0:
     children:
         - aws:ecr_image:lambda_function_0-image
-        - aws:ecr_repo:ecr_repo-0
+        - aws:ecr_repo:lambda_function_0-image-ecr_repo
         - aws:iam_role:lambda_function_0-ExecutionRole
     tag: big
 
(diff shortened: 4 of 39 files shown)


pkg/knowledge_base2/properties/property.go
```diff
@@ -73,6 +73,18 @@ func ValidatePropertyRef(value construct.PropertyRef, propertyType string, ctx k
 	if err != nil {
 		return nil, fmt.Errorf("error getting property %s on resource %s, while validating property ref: %w", value.Property, value.Resource, err)
 	}
+
+	// recurse down in case of a nested property ref
+	for _, ok := propVal.(construct.PropertyRef); ok; _, ok = propVal.(construct.PropertyRef) {
+		propVal, err = ValidatePropertyRef(propVal.(construct.PropertyRef), propertyType, ctx)
```
gordon-klotho, line 79:
```suggestion
	for propValRef, ok := propVal.(construct.PropertyRef); ok; propValRef, ok = propVal.(construct.PropertyRef) {
		propVal, err = ValidatePropertyRef(propValRef, propertyType, ctx)
```

pkg/templates/aws/edges/ecr_image-ecr_repo.yaml
```diff
@@ -1,2 +1,3 @@
 source: aws:ecr_image
 target: aws:ecr_repo
+unique: one-to-one
```
gordon-klotho, line 3: Wouldn't this make multiple repos, one per image instead of re-using the repo for multiple images? Should it be `many-to-one`?

gordon-klotho (approved):
Mostly fine, just the unique question (and its effect on the repo's name).

pkg/templates/aws/edges/ecr_image-ecr_repo.yaml
jhsinger-klotho: so currently we dont support more than 1 image per repo based on how our iac is, but we could change this and just make the repo generation unique everywhere for images via op rules instead

gordon-klotho: What about our iac prevents hosting more than 1 image to a repo? It's fine if we're working around it, just put that (not just 'what', but also 'why') in comments here in the meantime in that case.

pkg/templates/aws/resources/ecr_image.yaml
```diff
@@ -16,6 +16,7 @@ properties:
         direction: downstream
         resources:
           - aws:ecr_repo
+        unique: true
```
gordon-klotho, line 19: If the edge is `many-to-one`, is this unique flag still correct?

pkg/templates/aws/edges/ecr_image-ecr_repo.yaml
jhsinger-klotho: its because we push it with the same name so we overwrite the image if you have multiple images we are trying to push, we would have to start tracking tags for the images we are pulling in because in compute we feed the image.imageName which is not distinguishable between tags i dont believe. We can dig into it further, but in e2e tests i deployed 2 lambdas and they both ended up using the same image

pkg/templates/aws/resources/ecr_image.yaml
jhsinger-klotho: yeah we would have to do some refactoring to make this work otherwise. This is just saying we prefer 1:1 but we wont prevent many:1

pkg/templates/aws/edges/ecr_image-ecr_repo.yaml
gordon-klotho: Ah, we don't really set the name, which is odd. `${args.Repo.repositoryUrl}:` just pushes to the repo root... I'm surprised that even works. Couldn't we just throw the `Name` in there to support it? Ie, something like:
```js
pulumi.interpolate`${args.Repo.repositoryUrl}/{{.Name}}:{{ if .Tag }}${args.Tag}-{{ end }}${sha256}`
```

PR #920, opened 2024-02-01
Print errors to stdout

gordon-klotho:
Note: the added testdata cases showing the validation errors are demonstrating some gaps in the validation logic, but were unchanged by this PR.

### Standard checks


pkg/engine2/testdata/ecs_rds.err.json
```diff
@@ -0,0 +1,68 @@
+[
+  {
+    "error": {
+      "chain": [
+        "required property Id is not set on resource aws:subnet:vpc-0:subnet-0"
```
gordon-klotho, line 5: `Id` is one of the "required + deploy time" variables. Those should probably not cause validation errors.

pkg/engine2/testdata/k8s_api.err.json
```diff
@@ -0,0 +1,117 @@
...
+    "property": "BaseImage",
+    "resource": "aws:ecr_image:pod2-ecr_image",
+    "value": null
+  },
+  {
+    "error": {
+      "children": [
+        {
+          "chain": [
+            "invalid int value 80"
```
gordon-klotho, line 73: I think this is because they're `"80" (string)` and not being converted to an `int`

gordon-klotho:
Don't know why the unit tests fail, they've been consistently successful for me. Looks like maybe some non-determinism on op eval?

pkg/engine2/testdata/ecs_rds.err.json
jhsinger-klotho: yeah we have a card for this, so we might want to pick it up as a part of this workstream

pkg/engine2/testdata/k8s_api.err.json
jhsinger-klotho: that should be able to be converted but if not we need to either cut a card or update the code

pkg/engine2/cli.go
```diff
@@ -256,22 +264,107 @@ func setupProfiling() func() {
...
+	zap.S().Info("Running engine")
+	err := em.Engine.Run(context)
+	if err != nil {
+		returnCode = 1
+		if ee, ok := err.(engine_errs.EngineError); ok {
+			engErrs = append(engErrs, ee)
+		} else {
+			engErrs = append(engErrs, engine_errs.InternalError{Err: engine_errs.ErrorsToTree(err)})
+		}
+	}
```
jhsinger-klotho, line 280: could you just add a comment saying that specific errors wont be lumped into errors on the run just so we remember

pkg/engine2/cli.go
```diff
@@ -256,22 +264,107 @@ func setupProfiling() func() {
...
+
+	if len(context.Solutions) > 0 {
+		writeDebugGraphs(context.Solutions[0])
+		for _, d := range context.Solutions[0].GetDecisions().GetRecords() {
+			d, ok := d.(solution_context.MaybeErroDecision)
```
jhsinger-klotho, line 285: what is MaybeErroDecision

pkg/engine2/cli.go
```diff
@@ -256,22 +264,107 @@ func setupProfiling() func() {
...
+	}()
+	defer func() {
+		r := recover()
+		if r == nil {
+			return
+		}
+		zap.S().Errorf("panic: %v", r)
+		switch r := r.(type) {
+		case engine_errs.EngineError:
+			engErrs = append(engErrs, r)
```
jhsinger-klotho, line 342: should an engine error ever cause a panic?

pkg/engine2/solution_context/decisions.go
```diff
@@ -29,6 +27,11 @@ type (
 		internal()
 	}
 
+	MaybeErroDecision interface {
+		// AsEngineError returns an EngineError if the decision is an error, otherwise nil.
+		AsEngineError() engine_errs.EngineError
+	}
+
```
jhsinger-klotho, line 34: typo in the name, i also dont love the name starting with Maybe, makes it kind of confusing

pkg/engine2/cli.go
jhsinger-klotho: i see what it is before, just would rename since its not very clear

jhsinger-klotho:
"showing the validation errors are demonstrating some gaps in the validation logic, but were unchanged by this PR." are the gaps just the 2 things you commented or are there more?

pkg/engine2/cli.go
gordon-klotho: It's mostly to interrogate the decision to see if it can be converted to an error. Originally I had like `ToError` but since it could return `nil` (for when a validation wasn't an error), I felt like it didn't match well. I'd take a suggestion, not sure how else to convey that "this decision _might_ be an error, call this method to find out"

pkg/engine2/cli.go
gordon-klotho: It _shouldn't_, but in case it does I wouldn't want to wrap it in an internal error. I figured the check was worth the extra 2 line, even if it doesn't happen right now.

jhsinger-klotho (approved):

jhsinger-klotho (approved):

PR #921, opened 2024-02-01
increasing op eval unit test coverage

jhsinger-klotho:
mainly happy path tests to increase coverage on expected inputs/outputs. Now that the test base is in place we can ramp up the test coverage on those functions

```
ok  	github.com/klothoplatform/klotho/pkg/engine2/operational_eval	0.474s	coverage: 27.7% of statements
```

### Standard checks

### DIFF
--- a/pkg/engine2/operational_eval/evaluator.go
+++ b/pkg/engine2/operational_eval/evaluator.go
@@ -197,8 +197,10 @@ func (eval *Evaluator) isEvaluated(k Key) (bool, error) {
 	_, err = eval.unevaluated.Vertex(k)
 	if errors.Is(err, graph.ErrVertexNotFound) {
 		return true, nil
+	} else if err != nil {
+		return false, err
 	}
-	return false, err
+	return false, nil
 }
 
 func (eval *Evaluator) addEdge(source, target Key) error {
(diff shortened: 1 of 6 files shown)


pkg/engine2/operational_eval/evaluator.go
```diff
@@ -197,8 +197,10 @@ func (eval *Evaluator) isEvaluated(k Key) (bool, error) {
 	_, err = eval.unevaluated.Vertex(k)
 	if errors.Is(err, graph.ErrVertexNotFound) {
 		return true, nil
+	} else if err != nil {
+		return false, err
 	}
-	return false, err
+	return false, nil
```
gordon-klotho, line 203: What's the difference here?

pkg/engine2/operational_eval/evaluator_test.go
```diff
@@ -0,0 +1,306 @@
...
+			}
+			err := eval.enqueue(tt.changes)
+			if tt.wantErr {
+				assert.Error(err)
+				return
+			}
+			assert.NoError(err)
+			adj, err := eval.graph.AdjacencyMap()
+			assert.NoError(err)
+			for k, v := range tt.want {
```
gordon-klotho, line 224: Need to also check that there's nothing in the `AdjacencyMAp` that isn't in `tt.want` (Order check should be sufficient).

gordon-klotho (approved):
Nothing blocking

PR #922, opened 2024-02-01
init

AlaShibanAtKlo:
### Standard checks

### DIFF
--- a/pkg/templates/aws/resources/ecr_image.yaml
+++ b/pkg/templates/aws/resources/ecr_image.yaml
@@ -6,9 +6,22 @@ properties:
     type: string
     required: true
     description: The base image to use for the Docker image
+    sanitize: |
+      # https://docs.docker.com/engine/reference/commandline/tag/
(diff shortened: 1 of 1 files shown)


jhsinger-klotho (approved):

pkg/templates/aws/resources/ecr_image.yaml
```diff
@@ -19,19 +32,45 @@ properties:
...
+      }}
   Context:
     type: string
     default_value: '.'
     description: The build context path for the Docker build, specifying where to
       find the Dockerfile and other supporting files
+    sanitize: |
+      # No specific AWS documentation, general file path rules applied
+      {{ . 
+        | replace `[^[:alnum:]/.-]+` ""
```
gordon-klotho, line 51: You can definitely have a valid file path that doesn't match these rules, for example the `:` is valid.

gordon-klotho (commented):
Rest is 👍

PR #925, opened 2024-02-02
initial improvement for more resources

AlaShibanAtKlo:
### Standard checks

### DIFF
--- a/pkg/templates/aws/resources/iam_role.yaml
+++ b/pkg/templates/aws/resources/iam_role.yaml
@@ -13,6 +13,9 @@ properties:
         properties:
           Effect:
             type: string
+            allowed_values:
+              - Allow
+              - Deny
           Action:
             type: list(string)
           Resource:
@@ -33,7 +36,7 @@ properties:
                 type: map(string,string)
               StringLike:
                 type: map(string,string)
-              Null:
+              null:
                 type: map(string,string)
   ManagedPolicies:
     type: set(string)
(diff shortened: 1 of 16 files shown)


pkg/templates/aws/resources/s3_bucket_policy.yaml
```diff
@@ -37,7 +39,7 @@ properties:
                 type: map(string,string)
               StringLike:
                 type: map(string,string)
-              Null:
+              null:
```
jhsinger-klotho, line 42: we definitely dont want the actual property names changed, was this intended?

pkg/templates/aws/resources/iam_role.yaml
```diff
@@ -73,7 +79,7 @@ properties:
                     type: map(string,string)
                   StringLike:
                     type: map(string,string)
-                  Null:
```
jhsinger-klotho, line 76: same here

pkg/templates/aws/resources/iam_role.yaml
```diff
@@ -33,7 +36,7 @@ properties:
                 type: map(string,string)
               StringLike:
                 type: map(string,string)
-              Null:
+              null:
```
jhsinger-klotho, line 39: and here

gordon-klotho:
Is this PR still relevant?

gordon-klotho:
Closing per discussion. Keeping the branch in case we bring it back

PR #943, opened 2024-02-27
Global tags

jhsinger-klotho:
ability to put a global tag in place on all valid resources

added a test to make sure tags are added to resource tempaltes (have to manually override the list if it cant have tags) and updated test data

example 
```
/Users/jordansinger/workspace/klotho/engine Run -c add.yaml -o compiled -t MYGLOBALTAG
/Users/jordansinger/workspace/klotho/iac Generate -a test -i ./compiled/resources.yaml -o compiled

const subnet_1_route_table_nat_gateway_elastic_ip = new aws.ec2.Eip("subnet-1-route_table-nat_gateway-elastic_ip", {
        tags: {GLOBAL_KLOTHO_TAG: "MYGLOBALTAG", RESOURCE_NAME: "subnet-1-route_table-nat_gateway-elastic_ip"},
    })
const subnet_3 = new aws.ec2.Subnet("subnet-3", {
        vpcId: vpc_0.id,
        cidrBlock: "10.0.64.0/18",
        availabilityZone: availability_zone_1,
        mapPublicIpOnLaunch: false,
        tags: {GLOBAL_KLOTHO_TAG: "MYGLOBALTAG", RESOURCE_NAME: "subnet-3"},
    })
const subnet_0_route_table_nat_gateway = new aws.ec2.NatGateway("subnet-0-route_table-nat_gateway", {
        allocationId: subnet_0_route_table_nat_gateway_elastic_ip.id,
        subnetId: subnet_2.id,
        tags: {GLOBAL_KLOTHO_TAG: "MYGLOBALTAG", RESOURCE_NAME: "subnet-0-route_table-nat_gateway"},
    })
const subnet_1_route_table_nat_gateway = new aws.ec2.NatGateway("subnet-1-route_table-nat_gateway", {
        allocationId: subnet_1_route_table_nat_gateway_elastic_ip.id,
        subnetId: subnet_3.id,
        tags: {GLOBAL_KLOTHO_TAG: "MYGLOBALTAG", RESOURCE_NAME: "subnet-1-route_table-nat_gateway"},
    })
```
### Standard checks

### DIFF
--- a/pkg/engine/cli.go
+++ b/pkg/engine/cli.go
@@ -52,6 +52,7 @@ var architectureEngineCfg struct {
 	inputGraph  string
 	constraints string
 	outputDir   string
+	globalTag   string
 }
 
 var getValidEdgeTargetsCfg struct {
@@ -122,6 +123,7 @@ func (em *EngineMain) AddEngineCli(root *cobra.Command) {
 	flags.StringVarP(&architectureEngineCfg.inputGraph, "input-graph", "i", "", "Input graph file")
 	flags.StringVarP(&architectureEngineCfg.constraints, "constraints", "c", "", "Constraints file")
 	flags.StringVarP(&architectureEngineCfg.outputDir, "output-dir", "o", "", "Output directory")
+	flags.StringVarP(&architectureEngineCfg.globalTag, "global-tag", "t", "", "Global tag")
 	flags.StringVar(&engineCfg.profileTo, "profiling", "", "Profile to file")
 
 	getPossibleEdgesCmd := &cobra.Command{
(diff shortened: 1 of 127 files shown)


pkg/engine/cli.go
```diff
@@ -366,7 +368,9 @@ func (em *EngineMain) RunEngine(cmd *cobra.Command, args []string) (exitCode int
 		return
 	}
 
-	context := &EngineContext{}
+	context := &EngineContext{
+		GlobalTag: architectureEngineCfg.globalTag,
```
gordon-klotho, line 372: This is kind of leaking AWS specifics into general contexts. I'm okay with this for now, but maybe throw a TODO or FIXME and/or ticket to genericize it.

gordon-klotho (approved):

pkg/engine/cli.go
jhsinger-klotho: i mean we could say the concept of tag applies to other things as well, like we can take kubernetes objects via labels and i would assume gcp azure have a similar concept. we could name it something else if we want

PR #948, opened 2024-02-28
make cloudfront connect to load balanacers

jhsinger-klotho:
ability to connect Cloudfront -> LB

### Standard checks

### DIFF
--- a/pkg/engine/testdata/unsupported_edge.err.json
+++ b/pkg/engine/testdata/unsupported_edge.err.json
@@ -6,21 +6,5 @@
       "Source": "aws:cloudfront_distribution:cloudfront_distribution_2",
       "Target": "aws:rds_instance:mydb"
     }
-  },
-  {
-    "classification": "network",
-    "error_code": "edge_unsupported",
-    "satisfaction_edge": {
-      "Source": "aws:cloudfront_distribution:cloudfront_distribution_2",
-      "Target": "aws:rds_instance:mydb"
-    }
-  },
-  {
-    "classification": "cloudfront_origin",
-    "error_code": "edge_unsupported",
-    "satisfaction_edge": {
-      "Source": "aws:cloudfront_distribution:cloudfront_distribution_2",
-      "Target": "aws:rds_instance:mydb"
-    }
   }
 ]
--- a/pkg/templates/aws/edges/cloudfront_distribution-load_balancer.yaml
+++ b/pkg/templates/aws/edges/cloudfront_distribution-load_balancer.yaml
@@ -1,0 +1,88 @@
+source: aws:cloudfront_distribution
+target: aws:load_balancer
+always_process: true
+operational_rules:
+  - configuration_rules:
+      - resource: '{{ .Source }}'
+        configuration:
+          field: Origins
+          value:
+            - CustomOriginConfig:
+                HttpPort: 80
+                HttpsPort: 443
+                OriginProtocolPolicy: http-only
+                OriginSslProtocols:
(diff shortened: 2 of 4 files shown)


pkg/engine/testdata/unsupported_edge.err.json
```diff
@@ -6,21 +6,5 @@
...
     }
-  },
-  {
-    "classification": "network",
-    "error_code": "edge_unsupported",
-    "satisfaction_edge": {
-      "Source": "aws:cloudfront_distribution:cloudfront_distribution_2",
-      "Target": "aws:rds_instance:mydb"
-    }
-  },
```
gordon-klotho, line 17: This change seems unintentional - this should still fail, right? Even if it should no longer fail, this test is supposed to cover the unsupported edge, so could you update the input if that's the case?

pkg/templates/aws/edges/cloudfront_distribution-load_balancer.yaml
```diff
@@ -0,0 +1,88 @@
...
+            {{ $rules := allDownstream "aws:load_balancer_listener_rule" .Target }}
+            {{ range $i, $rule := $rules }}
+              {{ range $j, $condition := (fieldValue "Conditions" $rule) }}
+                {{range $k, $pathPattern := $condition.PathPattern.Values }}
+                  {{- if not (sliceContains $routes $pathPattern) }}
+                    {{ $routes = appendSlice $routes $pathPattern }}
+                  {{- end }}
+                {{- end }}
+              {{ end }}
+            {{ end }}
```
gordon-klotho, line 37: This kind of stuff is exactly why I wish we could use an actual scripting language instead of making our own. No action item, just commentary.

pkg/templates/aws/edges/cloudfront_distribution-load_balancer.yaml
```diff
@@ -0,0 +1,88 @@
...
+                  {{- if (sliceContains $condition.PathPattern.Values $route) }}
+                    {{ range $j, $method := $condition.HttpRequestMethod.Values }}
+                      {{ $methods = appendSlice $methods $method }}
+                    {{- end }}
+                  {{- end }}
+                {{- end }}
+              {{ end }}
+              {
+                "AllowedMethods": [
+                    {{- if or (sliceContains $methods "PATCH") (sliceContains $methods "POST") (sliceContains $methods "PUT") (sliceContains $methods "DELETE") }}
```
gordon-klotho, line 51: Is there a reason this wouldn't be just the `$methods`?

gordon-klotho (approved):

pkg/templates/aws/edges/cloudfront_distribution-load_balancer.yaml
jhsinger-klotho: cloudfront has to have them be set in certain groups, not really a free formed list

pkg/templates/aws/edges/cloudfront_distribution-load_balancer.yaml
jhsinger-klotho: yeah i think we can find ways to simplify this

pkg/engine/testdata/unsupported_edge.err.json
jhsinger-klotho: nope its intentionaly because we see that theres a path now that we can go cloudfront -> LB - > RDS somehow i guess. It still does fail on one of the other paths of classifications though

PR #949, opened 2024-02-28
memorydb support

jhsinger-klotho:
adds support for memory db clusters, users, acls, and subnet groups.

corresponding e2e test result from lambda https://github.com/klothoplatform/engine-e2e-tests/pull/11
```
======================================================================================================= test session starts ========================================================================================================
platform darwin -- Python 3.10.13, pytest-7.4.3, pluggy-1.3.0
rootdir: /Users/jordansinger/workspace/engine-e2e-tests/output/aws-43ubkr
asyncio: mode=strict
collected 1 item

test_suites/aws/test_modules/lambda/test_modules/memorydb/test_memorydb.py .                                                                                                                                                 [100%]

======================================================================================================== 1 passed in 3.48s =========================================================================================================
```

### Standard checks


pkg/templates/aws/resources/memorydb_acl.yaml
```diff
@@ -0,0 +1,36 @@
+qualified_type_name: aws:memorydb_acl
+display_name: MemoryDB ACL
+
+properties:
```
gordon-klotho, line 4: Should this have the cluster as the Namespace?

pkg/templates/aws/resources/memorydb_cluster.yaml
```diff
@@ -0,0 +1,135 @@
...
+    type: string
+    description: The endpoints of the cluster as a string.
+    configuration_disabled: true
+    deploy_time: true
+
+consumption:
+  emitted:
+    - model: EnvironmentVariables
+      value:
+        '{{ .Self.Name }}_ARN': '{{ fieldRef "Arn" .Self }}'
```
gordon-klotho, line 115: Careful with this one, if any other resources want to emit their ARN and they have the same name it would conflict.

pkg/templates/aws/resources/memorydb_subnetgroup.yaml
```diff
@@ -0,0 +1,45 @@
...
+              Type: private
+          - aws:subnet
+    description: A list of subnets for the RDS subnet group, with at least 2 needed,
+      and all subnets need to be of 'private' type
+  Description:
+    type: string
+    description: The description of the subnet group
+
+    
+  Arn:
```
gordon-klotho, line 26:
```suggestion
  Arn:
```
nit

gordon-klotho (approved):

pkg/templates/aws/resources/memorydb_acl.yaml
jhsinger-klotho: it can be reused across clusters

pkg/templates/aws/resources/memorydb_cluster.yaml
jhsinger-klotho: yeah thats true, i think we do this in a couple places so will have to have a good solution

PR #950, opened 2024-02-29
remove extra tags key from target group

jhsinger-klotho:
### Standard checks


gordon-klotho (approved):
Assuming this is already changed in templates

PR #952, opened 2024-03-04
Don't evaluate edge rules if property is set by constraints

gordon-klotho:
## Edge Rules
This is required for `gitea` StackPack to change HealthCheck protocol on an NLB to be `HTTP` (and others, I don't remember).

## `kb`
Also adds `kb` tool to help inspect the knowledgebase.

### Example
```sh
kb --classification network aws:ecs_service aws:efs_file_system
```
image

```sh
kb --distance 3 aws:ecs_service
```
image

## Support EnableExecuteCommand on ECS Service
For debugging ECS services

### Standard checks


jhsinger-klotho:
we need a way to then signal if an edge rule is critical if we want to guarantee correctness, otherwise we cant make that guarantee

pkg/engine/operational_eval/vertex_property.go
```diff
@@ -274,6 +277,25 @@ func (v *propertyVertex) evaluateResourceOperational(
...
+func (v *propertyVertex) shouldEvalEdges(cs []constraints.ResourceConstraint) bool {
+	if knowledgebase.IsCollectionProperty(v.Template) {
+		return true
+	}
+	for _, c := range cs {
+		if c.Target != v.Ref.Resource || c.Property != v.Ref.Property {
+			continue
+		}
+		// NOTE(gg): does operator even matter here? If it's not a collection,
+		// what does an 'add' mean? Should it allow edges to overwrite?
```
jhsinger-klotho, line 291: if its not a collection yeah im not sure that it would matter

pkg/engine/path_selection/path_selection.go
```diff
@@ -83,6 +84,24 @@ func BuildPathSelectionGraph(
...
+			if edgeTemplate == nil || edgeTemplate.DirectEdgeOnly {
+				validPath = false
+				break
+			}
+		}
+		if !validPath {
+			continue
+		}
+
+		satisfied_paths++
```
jhsinger-klotho, line 104: dont we already do this when building the path selection graph? Isnt this redundant

pkg/graph_addons/topology.go
```diff
@@ -0,0 +1,137 @@
...
+)
+
+// TopologicalSort provides a stable topological ordering of resource IDs.
+// This is a modified implementation of graph.StableTopologicalSort with the primary difference
+// being any uses of the internal function `enqueueArbitrary`.
+func TopologicalSort[K comparable, T any](g graph.Graph[K, T], less func(K, K) bool) ([]K, error) {
+	return toplogicalSort(g, less, false)
+}
+
+func toplogicalSort[K comparable, T any](g graph.Graph[K, T], less func(K, K) bool, invertLess bool) ([]K, error) {
```
jhsinger-klotho, line 17: do we want some unit tests for these?

jhsinger-klotho (commented):
nothing blocking really just a few questions

pkg/engine/path_selection/path_selection.go
gordon-klotho: Ah, yeah, I meant to move it instead of duplicate it. The main impact of this is that it removed resources used in paths that never connected to the target. Per the comment, it's mostly to support the `kb` tool to prevent extra irrelevant resources.

jhsinger-klotho (approved):

PR #954, opened 2024-03-05
Template Fixes for ECR and Cloudfront

DavidSeptimus-Klotho:
- Added Platform property to `aws:ecr_image`
- Removed requirement that path and method are set on the same alb listener rule condition for detecting methods on `aws:cloudfront_distribution -> aws:load_balancer` edges. A given rule's paths and methods are always on different conditions. Each condition only handles one condition type.

### DIFF
--- a/pkg/engine/testdata/2_routes.expect.yaml
+++ b/pkg/engine/testdata/2_routes.expect.yaml
@@ -101,6 +101,7 @@ resources:
     aws:ecr_image:lambda_function_0-image:
         Context: .
         Dockerfile: lambda_function_0-image.Dockerfile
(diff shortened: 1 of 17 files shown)


jhsinger-klotho:
why did all the tests change? Also good catch on the cdn, i assume that fixes the not picking up POST and what not. Also i have a task assigned to me for the ecr image which you should take

DavidSeptimus-Klotho:
The expectations all needed that platform property. The other test file just have changes to resource ordering, which doesn't seem likely to be associated with this specific set of changes.

gordon-klotho (approved):

PR #959, opened 2024-03-06
Use TopologicalSort for viz yaml same as resources.yaml

gordon-klotho:
### Standard checks

### DIFF
--- a/pkg/graph_addons/topology.go
+++ b/pkg/graph_addons/topology.go
@@ -39,6 +39,9 @@ func topologicalSort[K comparable](deps map[K]map[K]graph.Edge[K], less func(K, 
 			enqueue(vertex)
 		}
 	}
+	sort.Slice(queue, func(i, j int) bool {
+		return less(queue[i], queue[j])
+	})
 
 	// enqueueArbitrary enqueues an arbitray but deterministic id from the remaining unvisited ids.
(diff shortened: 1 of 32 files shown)


pkg/visualizer/file.go
```diff
@@ -37,7 +37,7 @@ func (f *File) WriteTo(w io.Writer) (n int64, err error) {
 	wh.Writef("provider: %s\n", f.Provider)
 	wh.Write("resources:\n")
 
-	resourceIds, err := construct.ReverseTopologicalSort(f.Graph)
+	resourceIds, err := construct.TopologicalSort(f.Graph)
```
gordon-klotho, line 40: This is the change that causes all the tests to change.

pkg/graph_addons/topology.go
```diff
@@ -39,6 +39,9 @@ func topologicalSort[K comparable](deps map[K]map[K]graph.Edge[K], less func(K,
 			enqueue(vertex)
 		}
 	}
+	sort.Slice(queue, func(i, j int) bool {
```
gordon-klotho, line 42: This has no effect but improves readability by putting the sort next to where all the queuing happens.

jhsinger-klotho (approved):

PR #960, opened 2024-03-07
support imported edges

jhsinger-klotho:
### Standard checks

### DIFF
--- a/pkg/engine/solution_context/view_raw.go
+++ b/pkg/engine/solution_context/view_raw.go
@@ -101,6 +101,19 @@ func (view RawAccessView) RemoveVertex(hash construct.ResourceId) error {
 func (view RawAccessView) AddEdge(source, target construct.ResourceId, options ...func(*graph.EdgeProperties)) error {
 	dfErr := view.inner.DataflowGraph().AddEdge(source, target, options...)
 
+	// check to see if both resources are imported and if so allow the edge
+	src, err := view.inner.DataflowGraph().Vertex(source)
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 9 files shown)


pkg/engine/solution_context/view_raw.go
```diff
@@ -101,6 +101,19 @@ func (view RawAccessView) RemoveVertex(hash construct.ResourceId) error {
...
+	src, err := view.inner.DataflowGraph().Vertex(source)
+	if err != nil {
+		return err
+	}
+	dst, err := view.inner.DataflowGraph().Vertex(target)
+	if err != nil {
+		return err
+	}
+	if src.Imported && dst.Imported {
+		return nil
```
gordon-klotho, line 114:
```suggestion
		// don't need to add it to the deployment graph - if both resources are imported, then
		// they don't need each other since neither are technically being deployed.
		return nil
```

pkg/infra/iac/templates/aws/cloudfront_distribution/factory.ts
```diff
@@ -47,7 +47,11 @@ function create(args: Args): aws.cloudfront.Distribution {
     })
 }
 
-function properties(object: ReturnType<typeof create>, args: Args) {}
+function properties(object: ReturnType<typeof create>, args: Args) {
+    return {
+        DomainName: pulumi.interpolate`https://${object.domainName}`,
```
gordon-klotho, line 52: This isn't the domain name, but a URL base.

gordon-klotho (approved):

PR #966, opened 2024-03-14
Add CachePolicyId and OriginRequestPolicyId

gordon-klotho:
and change default from ForwardedValues (legacy) to managed policies

Both Infracopilot and StackSnap need this (and currently set it manually in the IaC).

### Standard checks

### DIFF
--- a/pkg/templates/aws/resources/cloudfront_distribution.yaml
+++ b/pkg/templates/aws/resources/cloudfront_distribution.yaml
@@ -98,6 +98,12 @@ properties:
       SmoothStreaming:
         type: bool
         default_value: false
+      CachePolicyId:
+        type: string
+        default_value: '4135ea2d-6df8-44a3-9df3-4b5a84be39ad' # Managed-CachingDisabled
+      OriginRequestPolicyId:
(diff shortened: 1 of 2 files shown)


pkg/templates/aws/resources/cloudfront_distribution.yaml
```diff
@@ -69,16 +69,6 @@ properties:
...
-        properties:
-          QueryString:
-            type: bool
-            default_value: true
-          Cookies:
-            type: map
-            properties:
-              Forward:
-                type: string
-                default_value: none
```
jhsinger-klotho, line 81: why are these getting removed?

jhsinger-klotho: cant we just not set them rather than remove the ability to utilize them

gordon-klotho: Ah, yeah, they shouldn't - was a bit overzealous. Just need to remove the default values

jhsinger-klotho (approved):

PR #980, opened 2024-04-17
make ecs images be able to be strings

jhsinger-klotho:
### Standard checks

### DIFF
--- a/pkg/templates/aws/resources/ecs_task_definition.yaml
+++ b/pkg/templates/aws/resources/ecs_task_definition.yaml
@@ -95,7 +95,7 @@ properties:
             min_value: 2
             max_value: 60
       Image:
-        type: resource(aws:ecr_image)
+        type: string
         required: true
         operational_rule:
           step:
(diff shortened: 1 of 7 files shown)


pkg/templates/aws/resources/ecs_task_definition.yaml
```diff
@@ -95,14 +95,15 @@ properties:
...
         required: true
         operational_rule:
           step:
             direction: downstream
             resources:
               - aws:ecr_image:{{ .Self.Name }}-{{ if pathAncestorExists .Path 1}}{{ if hasField (printf "%s.Name" (pathAncestor .Path 1)) .Self}}{{ fieldValue (printf "%s.Name" (pathAncestor .Path 1)) .Self}}{{else}}image{{end}}{{else}}image{{end}}
             unique: true
+            use_property_ref: ImageName
         description: Reference to an Amazon Elastic Container Registry (ECR) image that
           will be used for the container within the task
```
gordon-klotho, line 108: Should probably update this to reflect the type change (and probably hint at being a property ref).

gordon-klotho (approved):

gordon-klotho (approved):
Tests 💯

PR #990, opened 2024-08-14
Fix Tests

atorres-klo:
fixed path for create_k2_test.sh, updated some testdata

### Standard checks

### DIFF
--- a/pkg/k2/testdata/bucket_ro/my-bucket.resources.yaml
+++ b/pkg/k2/testdata/bucket_ro/my-bucket.resources.yaml
@@ -8,9 +8,9 @@ resources:
             RESOURCE_NAME: my-bucket
 edges:
 outputs:
+    Bucket:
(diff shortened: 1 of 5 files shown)


pkg/k2/testdata/bucket_ro/my-bucket.resources.yaml
```diff
@@ -8,9 +8,9 @@ resources:
             RESOURCE_NAME: my-bucket
 edges:
 outputs:
+    Bucket:
```
gordon-klotho, line 11: These changes to outputs are suspicious - is there some nondeterminism here?

gordon-klotho (approved):

pkg/k2/testdata/bucket_ro/my-bucket.resources.yaml
atorres-klo: very possible, but didn't investigate too deeply

PR #993, opened 2024-08-15
Construct Template Enhancements

DavidSeptimus-Klotho:
This PR make substantial enhancements to construct templates

### Interpolation enhancements
- You can mix and match go template `{{}}` and existing `${}` interpolation syntax in construct/binding templates.
- Additionally, referencing properties containing a `.` is now possible using `[part1.part2]` syntax.

```yaml
prop1: '{{ toUpper .Inputs.Input1 }}'
prop2: '${inputs:Input1}-{{ .Inputs.Input1 }}-${resources:[{{ .Inputs.Name }}]}'
prop3: '{{ if .Inputs.ReadOnly }}readonly{{ else }}readwrite{{ end }}'
```

### Input rule enhancements
- You can now write `for_each`-`do` oriented rules (mixing `for_each` and `if` in a single rule will cause an error). The `do` component of the rule will be executed once for each index or entry in the item selected in the `for_each` go template using the `.Select` function.
- Resources added by input rules include a customizable prefix.
- In a `for_each`, rule, use the `.Select` function to select a value from the current interpolation context to iterate over and reference the current selection with `.Selected`.
- You can referenced the following within `.Selected` where applicable:
  - `.Key`
  - `.Index`
  - `.Value`

```yaml
input_rules:
  - for_each: '{{ .Select "inputs.Routes" }}'
    prefix: '{{ toLower (replace `[^\w]` "-"  .Selected.Value.Path) }}'
    do:
      resources:
        ...
``` 
- Input rules now contain a `rules` field that enables nested rules. Nested rules inherit the `prefix` of their parent by default. Prefix hierarchies are merged with a `.` separator (e.g., `MyRoute.Method`)

```yaml
input_rules:
  - if: {{ .Inputs.SomeBool }}
    then:
      resources:
        Resource1:
        ...
      rules:
        - for_each: {{ .Select "inputs.DependentInput" }}
          do: ...
```

### Structured Inputs
Structured inputs are now supported using the same syntax as resource template properties. Notably, `model` inputs are not available and resource refs (e.g. `resource:Id#property` are not supported)

No validation enhancements were made to this functionality -- that will be a separate effort, so most validation declared in templates is still ignored.

```yaml
Inputs:
  HeathCheck:
    type: map
    properties:
      Protocol:
        type: string
        required: true
        allowed_values:
          - TCP
          - UDP
          - HTTPS
          - HTTP
      Port:
        type: int
        default_value: 80
      Path:
        type: string
        default_value: /health
```


pkg/k2/constructs/construct_evaluator.go
```diff
@@ -1,28 +1,23 @@
 package constructs
 
 import (
-	"bytes"
 	"context"
 	"errors"
 	"fmt"
+	"github.com/klothoplatform/klotho/pkg/async"
+	template2 "github.com/klothoplatform/klotho/pkg/k2/constructs/template"
```
gordon-klotho, line 8: Since text/template isn't used, this can just be `template` and not aliased, right?

Also, nit: any way to configure your IntelliJ to sort these? I believe the Go language server uses the same as [`goimports`](https://pkg.go.dev/golang.org/x/tools/cmd/goimports) (an official tool).

pkg/k2/constructs/construct_evaluator_test.go
```diff
@@ -162,94 +187,140 @@ func TestInterpolateValue(t *testing.T) {
...
 		},
+		{
+			name:     "Go template interpolation",
+			rawValue: "Hello {{ .Inputs.stringInput }}",
+			data:     DynamicValueData{currentOwner: mockConstruct},
+			expected: "Hello Hello",
+		},
+		{
+			name:     "Go template evaluated before interpolation string",
+			rawValue: "Hello ${inputs:{{\"stringInput\"}}}",
```
gordon-klotho, line 308:
```suggestion
			rawValue: `Hello ${inputs:{{"stringInput"}}}`,
```
nit: a little nicer to use backticks to avoid needing escapes

pkg/k2/constructs/construct_marshaller.go
```diff
@@ -212,6 +217,8 @@ func (m *ConstructMarshaller) marshalRefs(o InfraOwner, rawVal any) (any, error)
 			}
 			ref.Index(i).Set(reflect.ValueOf(serializedField))
 		}
+	default:
+		// Do nothing
```
gordon-klotho, line 221: nit: add more details to the comment or don't include a `default`

pkg/k2/constructs/construct_marshaller_test.go
```diff
@@ -535,10 +521,12 @@ func TestConstructMarshaller_marshalRefs(t *testing.T) {
...
 			}
 			got, err := marshaller.marshalRefs(tt.args.o, tt.args.rawVal)
-			if (err != nil) != tt.wantErr {
-				t.Errorf("ConstructMarshaller.marshalRefs() error = %v, wantErr %v", err, tt.wantErr)
-				return
+			if tt.wantErr && assert.Error(t, err, "ConstructMarshaller.marshalRefs() error = %v, wantErr %v", err, tt.wantErr) {
+				t.FailNow()
+			} else if !tt.wantErr && !assert.NoError(t, err, "ConstructMarshaller.marshalRefs() error = %v, wantErr %v", err, tt.wantErr) {
+				t.FailNow()
 			}
```
gordon-klotho, line 528:
```suggestion
			if tt.wantErr  {
				require.Error(t, err)
			} else {
				require.NoError(t, err)
			}
```
nit: using `require` gets the FailNow behaviour built in, and the message seems redundant (assert/require already print the error if fails and the test name already has `marshalRefs` in it)

pkg/k2/constructs/constructs.go
```diff
@@ -1,70 +1,44 @@
...
 )
 
 type (
 	ResourceOwner interface {
 		GetResource(resourceId string) (resource *Resource, ok bool)
 		SetResource(resourceId string, resource *Resource)
 		GetResources() map[string]*Resource
-		GetTemplateResourcesIterator() Iterator[string, ResourceTemplate]
-		InterpolationSource
+		GetTemplateResourcesIterator() template.Iterator[string, template.ResourceTemplate]
```
gordon-klotho, line 15: TODO (for later, not this PR): https://go.dev/wiki/RangefuncExperiment ([1.23](https://tip.golang.org/doc/go1.23#language) adds it to the language)

pkg/k2/constructs/construct_evaluator.go
```diff
@@ -845,81 +415,176 @@ func (ce *ConstructEvaluator) evaluateResources(o ResourceOwner, interpolationCt
...
 
-func (ce *ConstructEvaluator) templateFunctions(ps *PropertySource) template.FuncMap {
-	funcs := template.FuncMap{}
-	funcs["inputs"] = GetPropertyFunc(ps)
-	return funcs
+func (ce *ConstructEvaluator) evaluateForEachRule(dv *DynamicValueData, rule template2.InputRuleTemplate) error {
+	parentPrefix := dv.resourceKeyPrefix
+
+	ctx := DynamicValueContext{
+		constructs: ce.Constructs,
```
gordon-klotho, line 432: Ahh, I see why you changed it to a pointer, but this could just as easily be `&ce.Constructs`

pkg/k2/constructs/dynamic_value.go
```diff
@@ -0,0 +1,325 @@
...
+		}
+	default:
+		return DynamicValueSelection{
+			Source: src,
+		}
+	}
+}
+
+// Next returns the next value in the selection and whether there are more values
+func (s *DynamicValueSelection) Next() (any, bool) {
```
gordon-klotho, line 287: This function can't be used in templates, probably worth documenting it as such.

> The name of a niladic method of the data, preceded by a period, such as .Method The result is the value of invoking the method with dot as the receiver, dot.Method(). Such a method must have one return value (of any type) or two return values, the second of which is an error.

pkg/k2/constructs/dynamic_value.go
```diff
@@ -0,0 +1,325 @@
...
+		return ""
+	}
+	switch v := v.(type) {
+	case string:
+		return v
+	case fmt.Stringer:
+		return v.String()
+	default:
+		return fmt.Sprintf("%v", v)
+	}
```
gordon-klotho, line 324: This switch is just a more complicated version of `fmt.Sprintf("%v", v)`. It's slightly faster, but I don't think it matters here.

pkg/k2/constructs/import_resources.go
```diff
@@ -0,0 +1,212 @@
...
+			}
+			return nil
+		})
+	}
+	return errors.Join(errs...)
+}
+
+func (ce *ConstructEvaluator) importResourcesFromInputs(o InfraOwner, ctx context.Context) error {
+	return o.ForEachInput(func(i property.Property) error {
+		// get the construct from the evaluator if it exists and is the correct type or return an error
```
gordon-klotho, line 158: Good comment, should probably just be the godoc for this method

pkg/k2/constructs/interpolation.go
```diff
@@ -0,0 +1,413 @@
+package constructs
```
gordon-klotho, line 1: 🙏 thanks for splitting up that huge file, much more readable

pkg/k2/constructs/template/construct_template_test.go
```diff
@@ -65,7 +67,7 @@ func TestParseConstructTemplateId(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			ctId, err := ParseConstructTemplateId(tt.input)
+			ctId, err := property.ParseConstructType(tt.input)
```
gordon-klotho, line 70: Feels like this should be moved into the `property` package since it's testing a function from there. Same with a couple other tests here

pkg/k2/constructs/template/execution/context.go
```diff
@@ -0,0 +1,131 @@
...
+type (
+	// Context is an interface that defines the methods to execute a template and decode the result into a value
+	Context interface {
+		// ExecuteUnmarshal executes the template tmpl using data as input and unmarshals the value into v
+		ExecuteUnmarshal(tmpl string, data any, v any) error
+		// Unmarshal unmarshals the template result into a value
+		Unmarshal(data *bytes.Buffer, v any) error
+	}
+
+	DefaultExecutionContext struct{}
```
gordon-klotho, line 24: Generally, interfaces are defined where they're used, not implemented. I also don't see any other implementations (maybe I'm missing it), so probably not necessary to preemptively make an interface for it.

pkg/k2/constructs/template/execution/context.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	err := json.Unmarshal([]byte(bstr), v)
+	if err == nil {
+		return nil
+	}
+
+	return err
+
+}
```
gordon-klotho, line 131: Publishing WIP review. This comment is a bookmark so I can return to it.

pkg/k2/constructs/template/inputs/properties_template.go
```diff
@@ -0,0 +1,430 @@
...
+	cloned := *p
+	cloned.Properties = make(InputTemplateMap, len(p.Properties))
+	for k, v := range p.Properties {
+		cloned.Properties[k] = v.Clone()
+	}
+	return &cloned
+}
+
+// fieldConversion is a map providing functionality on how to convert inputs into our internal types if they are not inherently the same structure
+var fieldConversion = map[string]func(val reflect.Value, p *InputTemplate, kp property.Property) error{
```
gordon-klotho, line 294: nit: for readability, would be nice to type alias the `func`

pkg/k2/constructs/template/inputs/properties_template.go
```diff
@@ -0,0 +1,430 @@
...
+	"SanitizeTmpl": func(val reflect.Value, p *InputTemplate, kp property.Property) error {
+		sanitizeTmpl, ok := val.Interface().(string)
+		if !ok {
+			return fmt.Errorf("invalid sanitize template")
+		}
+		if sanitizeTmpl == "" {
+			return nil
+		}
+		// generate random uuid as the name of the template
+		name := uuid.New().String()
```
gordon-klotho, line 304: Why use a uuid and not the property's name? Since we don't have delegate templates, the only place this is relevant is in making the error messages.

pkg/k2/constructs/template/inputs/properties_template_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"testing"
+
+	knowledgebase "github.com/klothoplatform/klotho/pkg/knowledgebase"
+	"github.com/klothoplatform/klotho/pkg/knowledgebase/properties"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_ConvertProperty(t *testing.T) {
```
gordon-klotho, line 11: non-blocking: would be nice to have more test cases

pkg/k2/constructs/template/properties.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	"github.com/klothoplatform/klotho/pkg/construct"
+	"github.com/klothoplatform/klotho/pkg/k2/constructs/template/inputs"
+	"github.com/klothoplatform/klotho/pkg/k2/constructs/template/property"
+	"gopkg.in/yaml.v3"
+)
+
+type Properties struct {
+	propertyMap property.PropertyMap
```
gordon-klotho, line 11: Is there a need for a dedicated type for this?

pkg/k2/constructs/template/property/construct_type.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+type ConstructType struct {
+	Package string `yaml:"package"`
+	Name    string `yaml:"name"`
+}
+
+func (c *ConstructType) UnmarshalYAML(value *yaml.Node) error {
+	// Split the value into parts
+	parts := strings.Split(value.Value, ".")
```
gordon-klotho, line 22: is `value.Value` safe to assume being the string? How does it behave if using a multiline string or other cases that have characters that wouldn't be present when unmarshalling to a string

nit: reduce the 3x copy of the parsing to one and reuse

pkg/k2/constructs/template/templates/aws/bucket/bucket.yaml
```diff
@@ -33,4 +32,12 @@ outputs:
...
\ No newline at end of file
+    value: ${resources:Bucket#BucketRegionalDomainName}
+
+input_rules:
+  - if: '{{ .Inputs.IndexDocument }}'
+    then:
+      resources:
+        Bucket:
+          properties:
+            IndexDocument: ${inputs:IndexDocument}
```
gordon-klotho, line 43: :+1: bonus step to static sites

pkg/construct/properties.go
```diff
@@ -25,14 +27,22 @@ func (r *Resource) SetProperty(pathStr string, value any) error {
...
 func (r *Resource) GetProperty(pathStr string) (any, error) {
 	path, err := r.PropertyPath(pathStr)
 	if err != nil {
 		return nil, err
 	}
-	return path.Get(), nil
+	if value, ok := path.Get(); ok {
+		return value, nil
+	}
+	// Backwards compatibility: if the property does not exist, return nil instead of an error.
```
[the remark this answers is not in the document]
DavidSeptimus-Klotho: IIRC, I did this, so I wouldn't have to introduce error handling in places that were relying on the previous behavior.

pkg/engine/operational_eval/dependency_capture_mock_test.go
```diff
@@ -57,15 +57,15 @@ func (mr *MockdependencyCapturerMockRecorder) DAG() *gomock.Call {
 // ExecuteDecode mocks base method.
 func (m *MockdependencyCapturer) ExecuteDecode(tmpl string, data knowledgebase.DynamicValueData, value any) error {
 	m.ctrl.T.Helper()
-	ret := m.ctrl.Call(m, "ExecuteDecode", tmpl, data, value)
+	ret := m.ctrl.Call(m, "ExecuteUnmarshal", tmpl, data, value)
```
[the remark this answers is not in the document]
DavidSeptimus-Klotho: This probably got caught up in a refactoring. Reverting.

pkg/k2/constructs/template/execution/context.go
DavidSeptimus-Klotho: It's used in the `properties` package, which contains all the property implementations. I'm open to moving it if you feel strongly about it.

pkg/k2/constructs/template/inputs/properties_template.go
DavidSeptimus-Klotho: This bit is copied verbatim from the knowledgebase properties implementation, but I can take a closer look at the points you raised in this comment and related subsequent comments.

pkg/k2/constructs/template/execution/context.go
gordon-klotho: I don't feel strongly, but the general practice of putting interfaces at the usage site makes them be way less bloated over time. It's a little weird compare to other languages like Java / TS, but it also makes it much easier to test since the surface area is smaller (the interface is what's used, not the union of all the uses).

pkg/k2/constructs/template/properties.go
DavidSeptimus-Klotho: `PropertyMap` is largely functionality extracted from [`ResourceTemplate`](https://github.com/klothoplatform/klotho/blob/main/pkg/knowledgebase/resource_template.go#L313-L313) and applied more generically to a map instead of to a template's `Properties` field. Its primary use is for working with nested properties. Interestingly, it looks like it also inherits a lack of support for `seg1[seg2.seg3]` path syntax in its `GetProperty()` method.

pkg/k2/constructs/template/property/construct_type.go
DavidSeptimus-Klotho: I don't understand what you mean by the nit. For the rest, I can decode the value to a string and validate it with a regex.

gordon-klotho: Oh, the nit is that `UnmarshalYAML`, `ParseConstrucType` and `FromString` all do the same thing. Decoding to string is the most safe, since I know it'd behave well in those edge cases.

pkg/engine/operational_rule/operational_rule.go
```diff
@@ -132,7 +132,7 @@ func (ctx OperationalRuleContext) CleanProperty(step knowledgebase.OperationalSt
 	if err != nil {
 		return err
 	}
-	prop := path.Get()
+	prop, _ := path.Get()
```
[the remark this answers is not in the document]
DavidSeptimus-Klotho: I'm not sure where you get that from?

The old behavior was to simply return `nil` from `PropertyPath.Get()` if either the path wasn't found or the value was `nil`. The new interface adds the boolean in the return to clarify the case where a path was found, but its content is `nil`. The pattern of ignoring the boolean was introduced to avoid accidentally changing behavior in areas of the engine in which `PropertyPath` is already being used.

pkg/k2/constructs/dynamic_value.go
DavidSeptimus-Klotho: You are correct. I'll update the documentation.

pkg/engine/operational_rule/operational_rule.go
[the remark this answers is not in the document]
gordon-klotho: For example, with the way that you changed 
```diff
func (i *arrayIndexPathItem) Get() (any, bool) {
+	if !i.a.IsValid() || !reflectutil.IsAnyOf(reflectutil.GetConcreteElement(i.a), reflect.Slice, reflect.Array) {
+		return nil, false
+	}
+	if i.a.Len() <= i.index {
+		return nil, false
+	}
	return i.a.Index(i.index).Interface()
```
This changes what would have previously `panic`d into one that returns false.

gordon-klotho (approved):

PR #996, opened 2024-08-16
Improves error reporting

DavidSeptimus-Klotho:
- In concise mode, errors are rendered cleanly at the end of the terminal output
- Silenced cobra error logging to avoid breaking our TUI output
- Added error detection for language host startup
- Added error messages to clarify issues with the klotho sdk installation

image

### DIFF
--- a/pkg/k2/language_host/language_host.go
+++ b/pkg/k2/language_host/language_host.go
@@ -3,11 +3,13 @@ package language_host
 import (
 	"context"
 	_ "embed"
+	"errors"
 	"fmt"
 	"io"
 	"os"
 	"os/exec"
 	"regexp"
+	"strings"
 	"syscall"
 
 	"github.com/klothoplatform/klotho/pkg/command"
@@ -19,27 +21,47 @@ import (
 //go:embed python/python_language_host.py
 var pythonLanguageHost string
 
-type ServerAddress struct {
+type ServerState struct {
 	Log     *zap.SugaredLogger
 	Address string
-	HasAddr chan struct{}
+	Error   error
+	Done    chan struct{}
+}
+
+func NewServerState(log *zap.SugaredLogger) *ServerState {
+	return &ServerState{
+		Log:  log,
+		Done: make(chan struct{}),
+	}
 }
 
 var listenOnPattern = regexp.MustCompile(`(?m)^\s*Listening on (\S+)$`)
+var exceptionPattern = regexp.MustCompile(`(?s)(?:^|\n)\s*Exception occurred: (.+)$`)
 
-func (f *ServerAddress) Write(b []byte) (int, error) {
+func (f *ServerState) Write(b []byte) (int, error) {
 	if f.Address != "" {
 		return len(b), nil
 	}
 	s := string(b)
-	matches := listenOnPattern.FindStringSubmatch(s)
+
+	// captures the gRPC server address
+	matches := exceptionPattern.FindStringSubmatch(s)
+	if len(matches) >= 2 {
+		f.Error = errors.New(strings.TrimSpace(matches[1]))
+		f.Log.Debug(s)
+	}
+
+	// captures a fatal error in the language host that occurs before the address is printed to stdout
+	matches = listenOnPattern.FindStringSubmatch(s)
 	if len(matches) >= 2 {
 		// address is the first match
 		f.Address = matches[1]
 		f.Log.Debugf("Found language host listening on %s", f.Address)
-		close(f.HasAddr)
 	}
 
+	if f.Address != "" || f.Error != nil {
+		close(f.Done)
+	}
 	return len(b), nil
 }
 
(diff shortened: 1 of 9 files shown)


pkg/k2/language_host/language_host.go
```diff
@@ -19,15 +21,26 @@ import (
 //go:embed python/python_language_host.py
 var pythonLanguageHost string
 
-type ServerAddress struct {
+type ServerState struct {
 	Log     *zap.SugaredLogger
 	Address string
+	Error   error
+	HasErr  chan error
 	HasAddr chan struct{}
```
gordon-klotho, line 29: Might be worth combining both of these channels into
```
Done chan struct{}
```
The `error` in the new HasErr doesn't seem to be used anyway.

pkg/tui/log_core.go
```diff
@@ -47,7 +47,7 @@ func (c *LogCore) With(f []zapcore.Field) zapcore.Core {
 			if c.verbosity.CombineLogs() {
 				field.AddTo(nc.enc)
 			}
-			// else (if the field is the construct and we're not combining logs) don't add it to the encoder
+			// else (if the field is the construct and we're not combining errors) don't add it to the encoder
```
gordon-klotho, line 50: Combing logs is the correct word here. This describes the difference, for eg, between `-v=2` and `-v=3`

pkg/tui/verbosity.go
```diff
@@ -27,14 +27,8 @@ func (v Verbosity) LogLevel() zapcore.Level {
 	}
 }
 
-// CombineLogs controls whether to show all logs comingled in the TUI. In other words,
-// sorted by timestamp, not grouped by construct.
+// CombineLogs controls whether to show all errors commingled in the TUI.
+// In other words, sorted by timestamp, not grouped by construct.
```
gordon-klotho, line 31: I don't think this comment is accurate - it's still just the logs comingled (uses `program.Println`) or grouped (uses `tui.model`).

DavidSeptimus-Klotho: *sigh* Intellij refactoring strikes again.

pkg/k2/language_host/language_host.go
```diff
@@ -19,15 +21,24 @@ import (
...
 	Address string
-	HasAddr chan struct{}
+	Error   error
+	Done    chan any
+}
+
+func NewServerState(log *zap.SugaredLogger) *ServerState {
+	return &ServerState{
+		Log:  log,
+		Done: make(chan any),
```
gordon-klotho, line 34: nit: `chan any` uses the size of a pointer whereas `chan struct{}` uses zero size. It also uses the types to indicate that it's just a signal, and that no meaningful information other than its completion can be derived from the channel. It's not a big deal, but just explaining why it was that previously.

pkg/k2/language_host/language_host.go
```diff
@@ -37,7 +48,12 @@ func (f *ServerAddress) Write(b []byte) (int, error) {
 		// address is the first match
 		f.Address = matches[1]
 		f.Log.Debugf("Found language host listening on %s", f.Address)
-		close(f.HasAddr)
+		close(f.Done)
+	}
+	matches = exceptionPattern.FindStringSubmatch(s)
+	if len(matches) >= 2 {
+		f.Error = errors.New(strings.TrimSpace(matches[1]))
+		close(f.Done)
```
gordon-klotho, line 56: if the address is found, then it errors (or it errors twice), this will double close the channel which IIRC will panic. Make sure to check for `Address` or `Error` before closing

gordon-klotho (approved):

pkg/k2/language_host/language_host.go
```diff
@@ -19,27 +21,47 @@ import (
...
+	// captures a fatal error in the language host that occurs before the address is printed to stdout
+	matches = listenOnPattern.FindStringSubmatch(s)
 	if len(matches) >= 2 {
 		// address is the first match
 		f.Address = matches[1]
 		f.Log.Debugf("Found language host listening on %s", f.Address)
-		close(f.HasAddr)
 	}
 
+	if f.Address != "" || f.Error != nil {
```
gordon-klotho, line 62: I think this condition is inverted - those conditions are when it is already closed.

DavidSeptimus-Klotho: If one of those is not nil/empty, we close the channel. This is the only place we close the channel. What am I missing?

gordon-klotho: Consider the 2nd time an Error or Address are encountered, this condition would be true the second time, resulting in a double close. Unless we're doing something to shut it down, but I don't think we are (and not sure we should for cases like Address -> Error).

DavidSeptimus-Klotho: Just updated this change. Now, we gate the close behind an if-else-if, so it only calls one or the other. Once Either branch evaluates to `true`, we close the channel and any subsequent calls to write will just return immediately. So you only get either an error or an address, but never both.

gordon-klotho (approved):

